PR #23, opened 2023-06-01
AskFileMessage accept parameter can now take a dict

willydouhard:
Right now `AskFileMessage`'s accept parameter takes a list. For instance `["text/plain"]` will generate `{"text/plain": []}` in the UI.
The UI uses react dropzone, here is how it handles file type restrictions: https://react-dropzone.org/#!/Accepting%20specific%20file%20types

This approach is not allowing to filter on specific file types. For instance, if I only want to accept a python file, i need the following accept in react dropzone: `{"text/plain": [".py"]}`.

This PR enables `AskFileMessage`'s accept parameter to either take a list (for simplicity) or a dict for more detailed use cases.

### DIFF
--- a/cypress/e2e/ask_file/main.py
+++ b/cypress/e2e/ask_file/main.py
@@ -3,17 +3,22 @@ import chainlit as cl
 
 @cl.on_chat_start
 def start():
-    file = None
-
-    # Wait for the user to upload a file
-    while file == None:
-        file = cl.AskFileMessage(
-            content="Please upload a text file to begin!", accept=["text/plain"]
-        ).send()
+    txt_file = cl.AskFileMessage(
+        content="Please upload a text file to begin!", accept=["text/plain"]
+    ).send()
     # Decode the file
-    text = file.content.decode("utf-8")
+    text = txt_file.content.decode("utf-8")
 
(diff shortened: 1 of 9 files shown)


cypress/e2e/ask_file/main.py
```diff
@@ -3,17 +3,22 @@
...
+        content=f"`Text file {txt_file.name}` uploaded, it contains {len(text)} characters!"
+    ).send()
 
-    # Wait for the user to upload a file
-    while file == None:
-        file = cl.AskFileMessage(
-            content="Please upload a text file to begin!", accept=["text/plain"]
-        ).send()
+    py_file = cl.AskFileMessage(
+        content="Please upload a python file to begin!", accept={"text/plain": [".py"]}
```
tpatel, line 17: Just a suggestion: you could add another test that submits a `text/plain` `.js` file to make sure it's not accepted.

tpatel (approved):
✨✨✨ AWESOME ✨✨✨

cypress/e2e/ask_file/main.py
willydouhard: Added a test with a cpp file

PR #25, opened 2023-06-01
perf: disable video on cypress tests by default

tpatel:
This speeds up the CI (no video recording + no video compression), we can still enable the video by reverting the flag locally.


tpatel:
replaced by https://github.com/Chainlit/chainlit/pull/27

PR #37, opened 2023-06-05
feat: add the pdf element

tpatel:
## Description

- This element allows displaying local or remote pdf files in the chat UI
- For now the implementation is simple: the backend sends a base64 string to the frontend which creates a data-uri for an iframe
- Reworked the `InlinedElements` component to avoid looping multiple times over the `elements` array
- Simplified the types in element.ts
- Updated the view.tsx so that it takes advantage of TypeScript safety
- Updated the filenames for the components in components/element/inlined/ to show they support a collection of items

## Screenshot

This feature uses the browser native PDF reader, hence the different pdf reader styles.

Firefox

image

Chrome

image

Safari

image


src/chainlit/element.py
```diff
@@ -144,3 +146,48 @@ def before_emit(self, text_element):
...
+        if url:
+            size = getattr(self, "size", None)
+            language = getattr(self, "language", None)
+            element = client.create_element(
+                name=self.name,
+                url=url,
+                type=self.type,
+                display=self.display,
+                size=size,
+                language=language,
```
willydouhard, line 181: pas sur que language soit important pour un PDF. c'est plus pour du code dans le TextElement

src/chainlit/frontend/src/components/chat/message/inlined.tsx
```diff
@@ -20,19 +16,40 @@ export default function InlinedElements({ elements, actions }: Props) {
...
   }
 
-  const images = elements.filter(
-    (el) => el.type === ElementType.img
-  ) as IImageElement[];
-
-  const texts = elements.filter(
-    (el) => el.type === ElementType.txt
-  ) as ITextElement[];
+  // Categorize the elements by element type
```
willydouhard, line 19: can we use a multi line comment here?

src/chainlit/frontend/src/components/chat/message/inlined.tsx
```diff
@@ -20,19 +16,40 @@ export default function InlinedElements({ elements, actions }: Props) {
...
+    }
+  );
 
   return (
     <Stack gap={1} mt={1}>
-      {images.length ? <InlinedImageList images={images} /> : null}
-      {Object.keys(texts).length ? <InlinedTextList items={texts} /> : null}
+      {elementsByType.image?.length ? (
+        <InlinedImageList
+          items={elementsByType.image.filter((x) => x.type == 'image')}
```
willydouhard, line 43: since we have elementsByType.image do we still need to filter again?

src/chainlit/element.py
tpatel: Ok, I wanted to keep it generic, I've removed it, we can add it later!

src/chainlit/frontend/src/components/chat/message/inlined.tsx
tpatel: Right, this was part of a fight I had with TypeScript, it's removed now!

src/chainlit/frontend/src/components/chat/message/inlined.tsx
tpatel: done

willydouhard (approved):

PR #40, opened 2023-06-06
Experiment with FastAPI (async architecture by default)

willydouhard:
**TLDR:** This proposal aims to transition from a synchronous Python runtime (Flask + threads) to an asynchronous one (FastAPI + event loop) for improved maintainability, compatibility with the broader Python ecosystem, and performance. Check out the [PR](https://github.com/Chainlit/cookbook/pull/2/files) on the cookbook repo to see the differences in real-life Chainlit apps.

### Motivation
Currently, Chainlit relies on a synchronous Python server to execute the developer's code and update the UI state accordingly. Since it processes requests one at a time, users must wait for their turn if another user's LangChain agent takes several seconds to complete.

To alleviate this issue, we use [threads](https://stackoverflow.com/questions/12758952/green-threads-and-thread-in-python) for each request. However, this approach has limitations:

1. It requires [patching](https://stackoverflow.com/questions/31128045/what-is-monkeypatching-with-the-python-gevent-library) and can cause compatibility issues with other packages.
2. Threads remain synchronous, which is problematic for I/O-intensive LLM apps, as they have to wait for one API call to finish before starting another.
3. It does not scale well.

Additionally, the current design prevents the use of async Python packages (OpenAI and LangChain have async implementations).

### Solution
I propose switching from a synchronous runtime to an asynchronous one (FastAPI + asyncio). If you're unfamiliar with this concept, you can learn more [here](https://superfastpython.com/asyncio-event-loop/#:~:text=The%20event%20loop%20is%20the%20core%20of%20every%20asyncio%20application,IO%20operations%2C%20and%20run%20subprocesses.&text=Event%20loops%20are%20a%20common,given%20their%20use%20in%20JavaScript).

This change implies that we would use the async and await keywords by default.

#### OpenAI example

```python
    response = openai.Completion.create(
        model=model_name, prompt=fromatted_prompt, **settings
    )
```

Becomes
```python
    response = await openai.Completion.acreate(
        model=model_name, prompt=fromatted_prompt, **settings
    )
```

#### LangChain example

```python
llm = OpenAI(temperature=0)
chain = LLMChain(llm=llm, prompt=PromptTemplate.from_template(prompt_template))
result = chain("hello world")
```

Becomes

```python
llm = OpenAI(temperature=0)
chain = LLMChain(llm=llm, prompt=PromptTemplate.from_template(prompt_template))
result = await chain.acall("hello world")
```

#### Running synchronous code

At times, an async implementation may not be available (e.g., for some LangChain tools). In such cases, we create a thread as a fallback:

```python
result = agent("2+2")
```

Becomes
```python
res = await asyncify(agent.__call__)("2+2")
```

To summarize, we leverage async implementations wherever possible and fallback to threads when only a synchronous implementation is available. This approach improves performance, compatibility, and supports both async and sync calls.

### Developer experience

Simplicity continues to be Chainlit's primary goal. While this proposal is a step in the right direction, I recognize that not everyone may be familiar with the async/await pattern. The intent of this PR is to ensure the community is comfortable with this transition regarding developer experience.

All APIs, classes, and methods remain backward compatible, except for the usage of async/await pattern.

The rule of thumb is to always use async implementations when available. When not available and for CPU bound/long running  tasks, use `asyncify` to run in a thread.

To provide context, here is the [PR](https://github.com/Chainlit/cookbook/pull/2/files) on the cookbook repo, showcasing the steps to migrate an existing Chainlit app using this proposal.

Finally, here are examples of Chainlit implementations using this proposal.

#### Synchronous @cl.langchain_run

The LangChain integration was initially hacky and required patching the LangChain library. Since then, LangChain has significantly improved, providing an opportunity to remove those hacks. If you are calling the LangChain agent yourself, pass the Chainlit callback as shown below:

```python
import chainlit as cl
from chainlit.sync import asyncify

@cl.langchain_run
async def run(agent, input_str):
    res = await asyncify(agent.__call__)(
        "2+2", callbacks=[cl.ChainlitCallbackHandler()]
    )
    await cl.Message(content=res["text"]).send()
```

#### Async @cl.langchain_run

Similarly, for the async implementation of LangChain, pass the Chainlit callback as shown below:

```python
@cl.langchain_run
async def run(agent, input_str):
    res = await agent.acall("2+2", callbacks=[cl.AsyncChainlitCallbackHandler()])
    await cl.Message(content=res["text"]).send()
```

If you want to test this right now:

1. Download the [wheel ](https://chainlit-cloud.s3.eu-west-3.amazonaws.com/dev-release/chainlit-0.3.0-py3-none-any.whl)
2. Run `pip uninstall chainlit`
3. Run `pip install PATH_TO_WHEEL`


ogabrielluiz:
It seems that in most cases there would be only one asyncify call, right? If the person creates a custom chain that uses other chains inside then there could be more than one. 

Have you tested this with VectorStores or other LangChain modules that do not have an implementation for async calls?

willydouhard:
> It seems that in most cases there would be only one asyncify call, right? If the person creates a custom chain that uses other chains inside then there could be more than one.
> 
> Have you tested this with VectorStores or other LangChain modules that do not have an implementation for async calls?

Only one asyncify call yes, every sub call from the chain will run in the same thread and hence not block the event loop.

I have tested with LangChain modules but not Vector DBs yet. I'll try that asap!

hooman-bayer:
Hi @willydouhard this is a quite nice solution. Just please note whether ‘streaming’ will work too in this scenarios. I see that you have created a callback handler for chainlit make sure it can accept stream as an input and generate tokens

willydouhard:
> Hi @willydouhard this is a quite nice solution. Just please note whether ‘streaming’ will work too in this scenarios. I see that you have created a callback handler for chainlit make sure it can accept stream as an input and generate tokens

Thanks! Streaming should work. The callback you mention is `cl.AsyncChainlitCallbackHandler()` will handle streaming full asynchronously if the LLM and LangChain agent you use support it.

If you do not use Langchain you can use the openai package directly:

```
response = await openai.Completion.acreate(
        model=model_name, prompt=fromatted_prompt, stream=True, **settings
    )
```

hooman-bayer:
@willydouhard I can also test tmrw morning once I’m next to my laptop ! But this is really amazing! Congrats again, this is going to bring chainlit to the commercial level potentially scaling to 100s users etc.

willydouhard:
> @willydouhard I can also test tmrw morning once I’m next to my laptop ! But this is really amazing! Congrats again, this is going to bring chainlit to the commercial level potentially scaling to 100s users etc.

Ty, would love to have your feedback after testing it!

derekcheungsa:
hi Willy, thanks for asking for community feedback.  In most cases I think devs will just take from sample code and adapt.  The only jarring call I can see is this one : res = await asyncify(agent.__call__)("2+2").  Seems reaching into internals.  

If that's only way, I think ok.  As the goal is a good one with better stability and performance.

willydouhard:
> hi Willy, thanks for asking for community feedback. In most cases I think devs will just take from sample code and adapt. The only jarring call I can see is this one : res = await asyncify(agent.**call**)("2+2"). Seems reaching into internals.
> 
> If that's only way, I think ok. As the goal is a good one with better stability and performance.

the __call__ could be re written like this:
```
@cl.langchain_run
async def run(agent, input_str):
    @cl.asyncify
    def run_agent_sync():
        res = agent(input_str, callbacks=[cl.ChainlitCallbackHandler()])
        return res

    res = await run_agent_sync()
    await cl.Message(content=res["text"]).send()
```

This is already working and doing the same thing, syntactic sugar!

src/ci-requirements.txt
```diff
@@ -1,2 +1,3 @@
-langchain==0.0.194
+langchain
```
tpatel, line 1: Are we fine not pinning langchain to a specific version here?

CHANGELOG.md
```diff
@@ -6,7 +6,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+### Added
+
+- Moving from sync to async runtime (breaking changes):
+  - Support async implementation (eg openai, langchain) 
+  - Performance improvements
+  - Removed patching of different libraries
+- Elements:
+  - Merged LocalImage and RemoteImage to Image
+  - New Avatar element to display avatars in messages
+- AskFileMessage now supports multi file uploads (small breaking change)
```
tpatel, line 18: I'm not sure this is a breaking change. It's a new feature but it doesn't break the current usage if I'm correct?

cypress/e2e/ask_multiple_files/main.py
```diff
@@ -0,0 +1,16 @@
+import chainlit as cl
+
+
+@cl.on_chat_start
+async def start():
+    files = await cl.AskFileMessage(
+        content="Please upload a text file to begin!",
```
tpatel, line 7:
```suggestion
        content="Please upload from one to two python files to begin!",
```

CHANGELOG.md
```diff
@@ -6,7 +6,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 
-Nothing unreleased at the moment.
+### Added
+
+- Moving from sync to async runtime (breaking changes):
+  - Support async implementation (eg openai, langchain) 
+  - Performance improvements
+  - Removed patching of different libraries
+- Elements:
+  - Merged LocalImage and RemoteImage to Image
```
tpatel, line 16:
```suggestion
  - Merged LocalImage and RemoteImage to Image (breaking change)
```

CHANGELOG.md
```diff
@@ -6,7 +6,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
-Nothing unreleased at the moment.
+### Added
+
+- Moving from sync to async runtime (breaking changes):
```
tpatel, line 11:
```suggestion
- Moving from sync to async runtime (breaking change):
```

cypress/e2e/langchain_postprocess/spec.cy.ts
```diff
@@ -7,12 +7,17 @@ describe("LangChain postprocess", () => {
     cy.wait(["@settings"]);
   });
 
-  it("should be able to postprocess a LangChain output", () => {
+  it("should be able to postprocess an LangChain output", () => {
```
tpatel, line 10:
```suggestion
  it("should be able to postprocess a LangChain output", () => {
```

src/chainlit/__init__.py
```diff
@@ -232,4 +244,8 @@ def sleep(duration: int):
     "on_stop",
     "action_callback",
     "sleep",
+    "ChainlitCallbackHandler",
```
tpatel, line 247: Should this be changed to `SyncChainlitCallbackHandler`? This would make it easier to choose between the two.

tpatel (commented):
A first pass, I'll review it again later today!

CHANGELOG.md
tpatel: Oh I was looking at the signature, not at the response, thanks 👌

src/chainlit/cli/mock.py
```diff
@@ -0,0 +1,48 @@
+def mock_openai():
```
tpatel, line 1: A proper place to mock the openai AI ❤️

src/chainlit/element.py
```diff
@@ -1,97 +1,75 @@
...
+import aiofiles
+
+from chainlit.emitter import get_emitter, BaseClient
 from chainlit.telemetry import trace_event
 from chainlit.types import ElementType, ElementDisplay, ElementSize
-from base64 import b64encode, b64decode
-from urllib import parse as urlparse
+
+type_to_mime = {
+    "image": "image/png",
```
tpatel, line 12: I would prefer to stick with a generic mime type here, to prevent any issue with an incorrect mime type.
The suggested mime type is the one we attribute to non-pdf elements, so it looks like the best fit.

```suggestion
    "image": "binary/octet-stream",
```

tpatel (approved):
✨✨✨ AWESOME ✨✨✨

src/chainlit/frontend/src/components/Hotkeys.tsx
```diff
@@ -0,0 +1,10 @@
+import { useHotkeys } from 'react-hotkeys-hook';
+import { useSetRecoilState } from 'recoil';
+import { settingsState } from 'state/settings';
+
+export default function Hotkeys() {
+  const setSettings = useSetRecoilState(settingsState);
+  useHotkeys('s', () => setSettings((old) => ({ ...old, open: !old.open })));
```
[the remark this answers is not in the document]
willydouhard: tested on safari and chrome and that was not the case

PR #41, opened 2023-06-08
avoid premature form submission when using IME

kun432:
### Description

When using an Japanese Input Method Editor (IME), the Enter key is used for finalizing character conversion. This conflicts with pressing the Enter key to submit form, and causes premature form submission and frustration for Japanese users who are in the middle of inputting text.

This PR ensures that the Enter key will only send the message if the user is not in the middle of an IME conversion, which should provide a more intuitive experience for users typing in Japanese.

NOTES: not sure but this might also happen in Chinese and Korean language which use IME.

### Solution

- track whether the user is in the middle of an IME conversion by using the `compositionstart` and `compositionend` events
- modify `handleKeyDown` to only send the message if the Enter key is pressed and the user is not in the middle of an IME conversion.

### Reference

- https://en.wikipedia.org/wiki/Japanese_input_method
- https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionstart_event
- https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionend_event

### DIFF
--- a/src/chainlit/frontend/src/components/chat/inputBox/input.tsx
+++ b/src/chainlit/frontend/src/components/chat/inputBox/input.tsx
@@ -32,6 +32,7 @@ const Input = ({ onSubmit, onReply }: Props) => {
   const askUser = useRecoilValue(askUserState);
   const session = useRecoilValue(sessionState);
   const [value, setValue] = useState('');
+  const [isComposing, setIsComposing] = useState(false);
 
   const socketOk = session?.socket && !session?.error;
   const disabled = !socketOk || loading || askUser?.spec.type === 'file';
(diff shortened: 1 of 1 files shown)


willydouhard:
I never used an IME myself before, thank you for the learning! I approved the CI run and will test locally before merging.

willydouhard:
Could you rebase on main @kun432 ? this will fix your CI build

kun432:
@willydouhard done!

willydouhard (approved):

willydouhard:
CI passed and approved! I let you merge. Thank you for your contribution 🥳

kun432:
@willydouhard 

Thank you for the approval! but seems I don't have write access to your repository and can't merge this. could you please do the merge?

kun432:
@willydouhard

Thank you for merging my PR! I'm glad tocontribute to the project!😄

willydouhard:
Happy to have you with us, can't wait to see what you will come up with next! 😄

PR #48, opened 2023-06-09
example colab notebook

sandorkonya:
this is an example notebook with chainlit on colab


willydouhard:
This looks awesome. Will test it locally tomorrow and then merge!

PR #54, opened 2023-06-12
feat: enable resizing the side panel

tpatel:
- The child component is hidden when resizing to speed up rendering
- The handle has a 5px+ border around it to help users select it
- Removed the code related to transitions as it wasn't working (transitions were activated on `transform`, `transform` wasn't used)


willydouhard:
looks good! could you provide a minimal gif of the result before merging? Thanks!

tpatel:
done!

willydouhard (approved):

PR #55, opened 2023-06-12
fix importing of os module in chainlit_run

HarshTrivedi:
Without this change starting the app with a specified port doesn't work.

In particular, running `chainlit run demo.py --port 9000` results in the following error:

```
    os.environ["CHAINLIT_PORT"] = port
    ^^
UnboundLocalError: cannot access local variable 'os' where it is not associated with a value
```


willydouhard:
Thank you for your contribution! This error will be fixed by #40 (which we plan to release today). However your were spot on with this one! 

Do not hesitate to contribute again 👍

PR #60, opened 2023-06-14
feat: add Pyplot chart element

tpatel:
- Accepts a pyplot figure, and generates the image
- Reuses the frontend code for the image element

Screenshot:

image


src/pyproject.toml
```diff
@@ -38,6 +38,7 @@ python-dotenv = "^1.0.0"
 auth0-python = "^4.1.1"
 uptrace = "^1.18.0"
 watchfiles="^0.19.0"
+matplotlib = "^3.7.1"
```
willydouhard, line 41: i would not add the dependency in the project. instead the user should have it installed optionally. I would also import matplotlib inside of the Pyplot class so that it only raise the import error when a user tries to instantiate the class but has not installed the module. we can add the module in the requirements-ci.txt for ci purpose

willydouhard: we do the same thing for langchain for instance

willydouhard (changes requested):

src/pyproject.toml
tpatel: Good call!

willydouhard (approved):

PR #65, opened 2023-06-15
fix: prevent substring matches in element links

tpatel:
- Sorting by descending length helps avoiding matching substrings, because the longuest strings are searched first

before

Screenshot 2023-06-15 at 10 23 34

after

Screenshot 2023-06-15 at 10 24 02

### DIFF
--- a/src/chainlit/frontend/src/components/chat/message/content.spec.tsx
+++ b/src/chainlit/frontend/src/components/chat/message/content.spec.tsx
@@ -1,0 +1,53 @@
+import '@testing-library/jest-dom';
+import { render } from '@testing-library/react';
+import { BrowserRouter } from 'react-router-dom';
+import MessageContent from './content';
+import { ITextElement } from 'state/element';
+
+it('renders the message content', () => {
+  const { getByText } = render(
+    <MessageContent
+      authorIsUser={false}
+      actions={[]}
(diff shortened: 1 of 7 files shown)


willydouhard (commented):
Awesome! Can we have a small test for that?

src/chainlit/frontend/src/components/chat/message/content.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+            type: 'text',
+            content: 'hi'
+          } as ITextElement
+        ]}
+        content={`Hello world
+        source_121, source_1, source_12`}
+      />
+    </BrowserRouter>
+  );
+  expect(getByText('source_1')).toBeInTheDocument();
```
willydouhard, line 50: does it check if the text is in the dom or if an element wraps this text specifically (Link in our case)?

willydouhard (commented):
Looks good!

src/chainlit/frontend/src/components/chat/message/content.spec.tsx
tpatel: `getByText` checks there is only one match with the exact text 👌

willydouhard (approved):

PR #70, opened 2023-06-15
perf: move prism to dynamic import

tpatel:
- This splits the code highlighting code into a new chunk
- Speeds up initial loading, only loads code highlighting when needed

before

Screenshot 2023-06-15 at 17 37 51

after

Screenshot 2023-06-15 at 17 39 13

todo

- [ ] add a test showing the code highlighting works

### DIFF
--- a/cypress/e2e/on_chat_start/main.py
+++ b/cypress/e2e/on_chat_start/main.py
@@ -3,4 +3,16 @@ import chainlit as cl
 
 @cl.on_chat_start
 async def start():
-    await cl.Message(content="Hello!").send()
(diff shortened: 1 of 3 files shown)


willydouhard (commented):
That looks good! Maybe the on_chat_start test is not right place to test that a message can render some code? How about making it a frontend test?

tpatel:
I've tried this morning but I couldn't make it work with the async `forceUpdate`. So I'm gonna stick with the e2e test for now, is that ok @willydouhard ?

willydouhard (approved):

PR #76, opened 2023-06-17
add cache decorator to be able to iterate with the file watcher more …

willydouhard:
Typical use case is:

```py
import chainlit as cl


@cl.cache
def load_data():
    import time

    time.sleep(5)
    return 10


data = load_data()


@cl.on_chat_start
async def on_chat_start():
    await cl.Message(content="Hello World!").send()
```

With no cache + file watcher, every change will trigger a reload of the file and we have to wait 5s. With the cache the reload is instant


tpatel (approved):
✨✨✨ AWESOME ✨✨✨

PR #89, opened 2023-06-20
file watcher now listen config changes and should stop correctly

willydouhard:


src/chainlit/config.py
```diff
@@ -103,6 +103,31 @@ class CodeSettings:
...
     lc_rename: Optional[Callable[[str], str]] = None
 
+    def validate(self):
+        requires_one_of = [
+            "lc_factory",
+            "on_message",
+            "on_chat_start",
+        ]
+
+        mutually_exclusive = ["lc_factory"]
```
tpatel, line 113: Is there one item missing in this array?

src/chainlit/server.py
```diff
@@ -46,6 +46,9 @@ async def lifespan(app: FastAPI):
             url = f"http://{host}:{port}"
 
         logger.info(f"Your app is available at {url}")
+
+        # Add a delay before opening the browser
```
tpatel, line 50: Is there a reason for the delay?

tpatel (commented):
I've got one comment on the mutually exclusive code

src/chainlit/config.py
willydouhard: No but there will be more in with the next integrations, typically llama_index

src/chainlit/server.py
willydouhard: Sometimes the server is not fully ready when the app opens in the browser. An error is then displayed for one second before the browser is able to make the connection. To avoid this we wait 1s

PR #96, opened 2023-06-21
Add local db / fs

willydouhard:
New setting `database` in the config.

- If set to local, chats are persisted in a local sqlite db and files in a folder on the local file system
- If set to cloud, chats are persisted on the Chainlit cloud
- If set to custom, the developer can use the [client_factory](https://github.com/Chainlit/chainlit/blob/wd/local-db/src/chainlit/__init__.py#L227) to pass a custom client instance that extends the [BaseClient](https://github.com/Chainlit/chainlit/blob/wd/local-db/src/chainlit/client/base.py) (example with the [LocalClient](https://github.com/Chainlit/chainlit/blob/wd/local-db/src/chainlit/client/local.py))


cypress/support/utils.ts
```diff
@@ -48,32 +86,42 @@ export function installChainlit() {
 }
 
 export function runSpec(test: string) {
-  return runCommand(`npx cypress run --spec cypress/e2e/${test}/spec.cy.ts`);
+  // const recordOptions = ` --record --key ${process.env.CYPRESS_RECORD_KEY} `;
```
tpatel, line 89: We could add a conditional, depending on whether the `process.env.CYPRESS_RECORD_KEY` env variable is set?
That would avoid the need for removing the comment + commit + push to run the CI again when debugging. And that would allow the PR runs to run correctly without needed the variable.

cypress/support/utils.ts
```diff
@@ -9,21 +9,59 @@ const FRONTEND_DIR = join(CHAINLIT_DIR, "chainlit", "frontend");
...
+const runLocal = [
+  "cot",
+  "global_elements",
+  "scoped_elements",
+  "update_message",
+  "remove_message",
+];
+
+function cleanLocalData(testDir: string) {
+  if (existsSync(join(testDir, ".chainlit/fs"))) {
```
tpatel, line 21: Could we use a more explicit folder name? Like `chat_files`

cypress/support/e2e.ts
```diff
@@ -3,8 +3,13 @@ dotenv.config();
 import { installChainlit, runTests, runTest } from "./utils";
 
 async function main() {
-  await installChainlit();
-  const singleTest = process.argv[2];
+  const singleTest = process.env.SINGLE_TEST;
+  const skipBuild = process.env.SKIP_BUILD;
```
tpatel, line 7: ❤️

src/chainlit/element.py
```diff
@@ -1,23 +1,27 @@
...
 
 from chainlit.context import get_emitter
-from chainlit.client import BaseClient
+from chainlit.client.base import BaseClient
 from chainlit.telemetry import trace_event
 from chainlit.types import ElementType, ElementDisplay, ElementSize
 
 type_to_mime = {
-    "image": "binary/octet-stream",
+    "image": "image/png",
```
tpatel, line 13: Not sure this is a good idea, the best would be to detect the image mime type on the server side (for example using https://stackoverflow.com/a/2753385/2831645).
I believe it will be an issue if folks try to upload an SVG.

cypress/support/utils.ts
willydouhard: I agree but this is a secret. It is annoying to add / remove the secret on the github UI before running the CI and could impact other runs

src/chainlit/element.py
willydouhard: I agree but then on the local file system images are not displayed at all since they have no extension. This is also problematic

willydouhard: I mean when a user browse the `/fs` folder on windows or mac

cypress/support/utils.ts
tpatel: Given we're using the `pull_request` github action, the secrets won't be applied to PR CI runs, right?
I don't have a strong opinion about this.

wei-ann-Github:
Hi, from this documentation, https://docs.chainlit.io/cloud/persistence/enterprise, on data persistence, it does not make any mention of the local and custom db option. Does this mean that the local and custom data persistence is not longer supported?

willydouhard:
The code from this PR is no longer supported. In chainlit 1.0.0 you will be able to declare your own custom data layer if you do not want to use ours. It will be a bit time consuming though

wei-ann-Github:
> The code from this PR is no longer supported. In chainlit 1.0.0 you will be able to declare your own custom data layer if you do not want to use ours. It will be a bit time consuming though

@willydouhard , thank you for your reply. I am restricted to using only custom DB. As I am current still using version 0.6.3, may I know if:
1. I can get the documentation for before v1.0.0?
2. there are documentation for my own custom data layer for the current chainlit version?

I am also looking to add a chat login page. May I know if the procedure is still the same in version > 1.0.0 as in v0.6.3?

willydouhard:
It should be pretty similar, here is the[ 1.0.0 documentation](https://docs-prerelease.chainlit.io/get-started/overview) (pre release still)

PR #101, opened 2023-06-24
fix commands order to start UI

xleven:
There is no `buildUi` script in `/src/chainlit/frontend/package.json` so the commands to start UI in CONTRIBUTING.md will not work. Should run `buildUi` at repo root then cd into frontend folder.


willydouhard (approved):

willydouhard:
Nice catch, thank you!

willydouhard:
This is good feedback, thank you! Can you have a look @tpatel ?

PR #112, opened 2023-06-27
Task list

tpatel:
Add the tasklist element

- Enables displaying a structured list of steps taken by the AI
- Useful to add transparency on autonomous agents

The tasklist has a top-level state, as well as a list of tasks that have a title and a status.

## Screenshots

Desktop

image

Mobile

image


src/chainlit/frontend/src/components/tasklist.tsx
```diff
@@ -0,0 +1,258 @@
...
+      >
+        🗒️ Task List
+      </Box>
+      <Chip
+        label={status || '?'}
+        sx={{
+          fontWeight: '500',
+          borderRadius: '4px',
+          backgroundColor: theme.palette.background.default,
+          color: theme.palette.mode === 'dark' ? '#9E9E9E' : '#757575'
```
willydouhard, line 102: I would like to avoid hardcoding colors

src/chainlit/frontend/src/components/tasklist.tsx
```diff
@@ -0,0 +1,258 @@
...
+        }}
+      />
+    </Box>
+  );
+};
+
+const Task = ({ index, task }: { index: number; task: ITask }) => {
```
willydouhard, line 109: we could create a TaskList folder in the components dir and separate the tasklist.ts, tasklistIcon .ts and index.ts for readability

src/chainlit/frontend/src/components/tasklist.tsx
```diff
@@ -0,0 +1,258 @@
...
+            strokeWidth={1.5}
+            d="m14.5 9.5-5 5m0-5 5 5"
+          />
+        </>
+      )}
+    </svg>
+  </SvgIcon>
+);
+
+const Header = ({ status }: { status: ITask['status'] }) => {
```
willydouhard, line 80: the status here could be not correlated to a tasks. Like criticizing the whole task list for instance

src/chainlit/frontend/src/components/tasklist.tsx
tpatel: ✅

src/chainlit/frontend/src/components/tasklist.tsx
tpatel: Good catch! There is no such constraint in the backend, I've just fixed it 👌

src/chainlit/frontend/src/components/tasklist.tsx
tpatel: While I usually agree, these colors are specific to this component (they don't appear anywhere else in the app).
Should we still move it to the global color palette, or wait for another component to use the same colors?

willydouhard (approved):
awesome

PR #119, opened 2023-06-30
improved the Task list component

tpatel:
- add a loading animation to running tasks
- fix TaskStatusIcon alignment
- also refactored colors to use the color palette

### DIFF
--- a/src/chainlit/frontend/src/theme.ts
+++ b/src/chainlit/frontend/src/theme.ts
@@ -1,4 +1,5 @@
 import { createTheme } from '@mui/material/styles';
+import { grey, primary, white } from 'palette';
 
 const typography = {
   fontFamily: ['Inter', 'sans-serif'].join(',')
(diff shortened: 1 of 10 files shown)


src/chainlit/frontend/src/theme.ts
```diff
@@ -82,26 +83,26 @@ const lightTheme = createTheme({
     success,
     error,
     background: {
-      default: '#FAFAFA',
-      paperVariant: '#EEEEEE',
-      paper: '#FFF'
+      default: grey[50],
+      paperVariant: grey[200],
```
willydouhard, line 87: we can get rid of paperVariant et use the grey palette directly

willydouhard (changes requested):

src/chainlit/frontend/src/theme.ts
tpatel: ✅

willydouhard (approved):

PR #124, opened 2023-07-03
fix: get rid of the `serverUrl` manual change when developping

tpatel:
- Introduce a conditional instead of relying on modifying the source code
- Modify the contributing docs based on https://github.com/Chainlit/chainlit/pull/101

## Why am i using ``?

> By default, the dev server (dev command) runs in development mode and the build command runs in production mode.

Source: [Vite modes](https://vitejs.dev/guide/env-and-mode.html#modes)

This means that `import.meta.env.DEV` is:

- `true` when running `vite`
- `false` when running `vite build`


willydouhard (approved):
awesome

PR #127, opened 2023-07-03
Audio element

tpatel:
- Add Audio element
- Enabled automated mime type detection


src/chainlit/frontend/src/components/element/audio.tsx
```diff
@@ -0,0 +1,14 @@
...
+  element: IAudioElement;
+}
+
+export default function AudioElement({ element }: Props) {
+  if (!element.url && !element.content) {
+    return null;
+  }
+  const className = `${element.display}-audio`;
+  const src = element.url || URL.createObjectURL(new Blob([element.content!]));
+  return <audio controls src={src} className={className}></audio>;
```
willydouhard, line 13: Is there a way to display the name of the audio file in the native audio component? Or should we add it ourselves?

src/chainlit/frontend/src/state/element.ts
```diff
@@ -42,6 +49,12 @@ export interface IPdfElement extends IElement {
   url?: string;
 }
 
+export interface IAudioElement extends IElement {
+  type: 'audio';
+  content?: string;
```
willydouhard, line 54: I don't think the content should be of type string, maybe you copy pasted from the TextElement?

willydouhard (changes requested):
looks good! minor things to change

cypress/e2e/audio_element/spec.cy.ts
```diff
@@ -0,0 +1,14 @@
...
+    cy.wait(["@settings"]);
+  });
+
+  it("should be able to display an audio element", () => {
+    cy.get("#welcome-screen").should("exist");
+
+    cy.get(".message").should("have.length", 1);
+    cy.get(".message").eq(0).find(".inline-audio").should("have.length", 1);
+  });
+});
```
willydouhard, line 14: [this](https://github.com/cypress-io/cypress/issues/1750) looks interesting to actually test the audio

willydouhard (changes requested):

src/chainlit/frontend/src/components/element/audio.tsx
tpatel: Good point! Here is how it looks:

image

I took some inspiration from this part of the figma designs:

image

src/chainlit/frontend/src/state/element.ts
tpatel: fixed

src/chainlit/frontend/src/components/element/audio.tsx
```diff
@@ -1,14 +1,31 @@
...
 
 export default function AudioElement({ element }: Props) {
+  const theme = useTheme();
   if (!element.url && !element.content) {
     return null;
   }
   const className = `${element.display}-audio`;
   const src = element.url || URL.createObjectURL(new Blob([element.content!]));
-  return <audio controls src={src} className={className}></audio>;
+  return (
```
willydouhard, line 16: it is better to use the Typography element here like 

```jsx
      <Typography
        color="text.primary"
        sx={{ fontSize: '14px', fontWeight: 700 }}
      >...</Typography>
```

unless the color is different from the main or secondary color in the theme?

src/chainlit/frontend/src/components/element/audio.tsx
willydouhard: nice!

src/chainlit/frontend/src/components/element/audio.tsx
tpatel: I've used a color that's different from the main/secondary color. It's slightly muted, like the Figma design of the image filename (that's the closest thing I've found for this purpose).

image

cypress/e2e/audio_element/spec.cy.ts
tpatel: I've found a way to load the audio file, wait for it to be loaded and then check that it was loaded correctly ✅

willydouhard (approved):

PR #134, opened 2023-07-04
Wd/auth

willydouhard:


src/chainlit/db/prisma/schema.prisma
```diff
@@ -8,11 +8,23 @@ generator client {
   recursive_type_depth = 5
 }
 
+model User {
+  id            Int             @id @default(autoincrement())
+  createdAt     DateTime        @default(now())
```
tpatel, line 13: You could add an `updatedAt` row to help debugging in the future https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#updatedat

tpatel (approved):
✨✨✨ AWESOME ✨✨✨

PR #138, opened 2023-07-06
serve files located in app dir

willydouhard:


src/chainlit/server.py
```diff
@@ -306,11 +306,16 @@ async def serve_file(filename: str):
...
-        return FileResponse(path_to_file)
-    else:
-        return HTMLResponse(content=html_template, status_code=200)
+    app_file_path = os.path.join(config.root, path)
+    ui_file_path = os.path.join(build_dir, path)
+
+    file_paths = [app_file_path, ui_file_path]
+
+    for file_path in file_paths:
+        if file_path != "" and os.path.isfile(file_path):
```
tpatel, line 315: `file_path` can't be empty, right?
I would check whether `path` is empty before doing the `for` loop

tpatel (approved):
✨✨✨ AWESOME ✨✨✨

PR #145, opened 2023-07-07
check if session exists

willydouhard:


src/chainlit/server.py
```diff
@@ -374,9 +374,10 @@ def emit_fn(event, data):
...
     # Function to ask the user a question
     def ask_user_fn(data, timeout):
-        if sessions[sid]["should_stop"]:
-            sessions[sid]["should_stop"] = False
-            raise InterruptedError("Task stopped by user")
+        if sid in sessions:
+            if sessions[sid]["should_stop"]:
+                sessions[sid]["should_stop"] = False
+                raise InterruptedError("Task stopped by user")
         return socket.call("ask", data, timeout=timeout, to=sid)
```
tpatel, line 381: Should we continue and send a message to the session id if it's not in the `sessions` object?

willydouhard: I'd say no but I can see weird edge cases happening

PR #147, opened 2023-07-07
strict compare messageId and highlightedMessage

willydouhard:


tpatel (approved):
✨✨✨ AWESOME ✨✨✨

When squashing the PR, could you change the commit message to share what it's fixing?

PR #155, opened 2023-07-10
Add design page on dev mode

alimtunc:
The purpose of this PR is to create a page that consolidates all existing UI components. 

Actually, I've just added the new designed one. In the near future, we'll add all of them.

Capture d’écran 2023-07-13 à 17 19 23


willydouhard (approved):

willydouhard:
Looks like there is an [import issue](https://github.com/Chainlit/chainlit/actions/runs/5522415019/jobs/10071964587?pr=155#step:9:21)

willydouhard:
Good to me!

Issue #108, opened 2023-06-26
Double output of response with ChainlitCallbackHandler


SimonB97:
Hi, your package is awesome, thank you for this!

Unfortunately, i'm having an issue where the resonses are displayed two times each when using the ChainlitCallbackHandler:

<img width="283" alt="image" src="https://github.com/Chainlit/chainlit/assets/102378134/71a446f2-462e-4a11-a0a7-005cc6280bc5">

i have to use the callback handler to be able to get streaming output with an initially sync agent from langchain wich i made async by applying the fix (last function below) mentioned in your docs:

```
@cl.langchain_factory(use_async=True)
def factory():
    # Initialize the OpenAI language model
    llm = ChatOpenAI(
        temperature=0, 
        model="gpt-3.5-turbo-0613", 
        streaming=True, 
        callbacks=[cl.ChainlitCallbackHandler()]
    )

    # Initialize the SerpAPIWrapper for search functionality
    search = SerpAPIWrapper()

    # Define a list of tools offered by the agent
    tools = [
        Tool(
            name="Search",
            func=search.run,
            description="Useful when you need to answer questions about current events or if you have to search the web. You should ask targeted questions like for google."
        ),
        WriteFileTool(),
        ReadFileTool()
    ]

    # needed for memory with openai functions agent
    agent_kwargs = {
        "extra_prompt_messages": [MessagesPlaceholder(variable_name="memory")],
    }

    memory = ConversationTokenBufferMemory(
        memory_key="memory", 
        return_messages=True,
        max_token_limit=2000,
        llm=llm
    )

    mrkl = initialize_agent(
        tools=tools,
        llm=llm, 
        agent=AgentType.OPENAI_MULTI_FUNCTIONS, 
        verbose=True, 
        agent_kwargs=agent_kwargs, 
        memory=memory,
    )

    return mrkl


@cl.langchain_run
async def run(agent, input):
    # Since the agent is sync, we need to make it async
    res = await cl.make_async(agent.run)(input)
    await cl.Message(content=res).send()
```

How can i stop the double output?

willydouhard:
What happens is that you see the LLM response (first message), and the langchain agent final answer(second message). Usually it is supposed to be indented but since your are passing the callback handler to only the LLM, there is nothing to indent the message in.

remove this line `callbacks=[cl.ChainlitCallbackHandler()]` in the LLM definition and change langchain_run like this:

```py
@cl.langchain_run
async def run(agent, input):
[...]

Issue #61, opened 2023-06-14
How to remove the button under a chat message?


prasoon2211:
Hi, I would like to remove the button that shows up under a user message. See image below - I'd like the remove the 'Took 1 step' button
<img width="504" alt="image" src="https://github.com/Chainlit/chainlit/assets/2565248/3492d55a-9464-4d2f-a458-d17baef9e92a">

Also, when I set `stream=True`, the text that shows up inside the debugging/intermediate/chain of thought textboxes (see image below) does get streamed but the final reply shown to the user doesn't get streamed. Can you please let me know how to stream the reply in real-time to the user?
<img width="1274" alt="image" src="https://github.com/Chainlit/chainlit/assets/2565248/ce77f044-7ace-4e3f-8b33-82dfd9fb5e05">

willydouhard:
You can set the config field `hide_cot` to `true` in the `.chainlit/config.toml` file. Then restart the chainlit server.

prasoon2211:
Thank you! And is there a way to stream the response to the user? I'm already using `streaming=True` in the `ChatOpenAI` constructor:

```python
@cl.langchain_factory(use_async=True)
async def init():
    # Create a chain that uses the pinecone vector store
    chain = ConversationalRetrievalChain.from_llm(
        ChatOpenAI(temperature=0.2, streaming=True, model="gpt-3.5-turbo-0613"),
        chain_type="stuff",
        retriever=ch_vecstore.as_retriever(),
        max_tokens_limit=4097,
        memory=memory,
        combine_docs_chain_kwargs={"prompt": qa_prompt},
    )
    return chain


@cl.langchain_postprocess
async def process_response(res):
    await cl.Message(content=res["answer"]).send()
```

willydouhard:
At the moment the langchain integration does not support final response streaming. Only the intermediary steps are streamed. This is the default behavior with langchain. We are exploring solutions to fix this!

Issue #34, opened 2023-06-04
Streaming=True not working when i integrate Langchain. 


altafr:
import os
from langchain import PromptTemplate, OpenAI, LLMChain
import chainlit as cl

#os.environ["OPENAI_API_KEY"] = "YOUR_OPEN_AI_API_KEY"

template = """Question: {question}

Answer: Let's think step by step."""

llm=OpenAI(temperature=0,streaming=True)

@cl.langchain_factory
def factory():
  
    prompt = PromptTemplate(template=template, input_variables=["question"])
    llm_chain = LLMChain(prompt=prompt, llm=llm , verbose=True)

    return llm_chain

willydouhard:
At the moment the intermediate steps should be streamed and not the final response. Is that the case or even the intermediate steps are not being streamed? You can see https://github.com/Chainlit/chainlit/issues/7#issuecomment-1569969095 for context

gruckion:
> import os from langchain import PromptTemplate, OpenAI, LLMChain import chainlit as cl
> 
> #os.environ["OPENAI_API_KEY"] = "YOUR_OPEN_AI_API_KEY"
> 
> template = """Question: {question}
> 
> Answer: Let's think step by step."""
> 
> llm=OpenAI(temperature=0,streaming=True)
> 
> @cl.langchain_factory def factory():
> 
> ```
> prompt = PromptTemplate(template=template, input_variables=["question"])
> llm_chain = LLMChain(prompt=prompt, llm=llm , verbose=True)
> 
> return llm_chain
> ```

Since you are using an `LLMChain` the streamed content will be inside the working box that appears. If you ask it to tell you a 200 word story and then expand the box. You will see that the content within is streaming.

Please confirm if this is working for you so we can close this issue.

If instead you want to stream the content directly. Below is a simple example of how to feed a message directly back to the UI using streaming.

```py
import chainlit as cl

from langchain.chat_models import ChatOpenAI
from langchain.schema import HumanMessage


llm = ChatOpenAI(streaming=True)

@cl.on_message
def main(message: str):
    cl.Message(content="") # Need to set the Message content to be blank before trigging the llm.
    llm([HumanMessage(content=message)])
```

The above work around using `cl.Message(content="")` is needed and will be fixed in a future release. It is required as of version `0.2.109`

altafr:
thanks

PR #158, opened 2023-07-11
Stream final answer

ramnes:
cc #7 #34 #61 #108

### DIFF
--- a/src/chainlit/lc/callbacks.py
+++ b/src/chainlit/lc/callbacks.py
@@ -1,9 +1,5 @@
 from typing import Any, Dict, List, Optional, Union
 from langchain.callbacks.base import BaseCallbackHandler, AsyncCallbackHandler
-from langchain.callbacks.streaming_aiter_final_only import (
-    DEFAULT_ANSWER_PREFIX_TOKENS,
-    AsyncFinalIteratorCallbackHandler,
-)
 from langchain.schema import (
     AgentAction,
     AgentFinish,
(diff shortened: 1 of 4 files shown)


src/chainlit/lc/callbacks.py
```diff
@@ -121,9 +172,11 @@ def start_stream(self):
         )
         self.stream = streamed_message
 
-    def send_token(self, token: str):
-        if self.stream:
-            run_sync(self.stream.stream_token(token))
+    def send_token(self, token: str, final: bool = False):
+        stream = self.final_stream if final else self.stream
+        if stream:
+            run_sync(stream.stream_token(token))
```
willydouhard, line 178: if final is true then we want to also call .send() on the message, since we want to persist it. Persistence is not called on each streamed token (it would flood the db)

willydouhard (changes requested):

src/chainlit/lc/callbacks.py
ramnes: something like this? 11b58868b38b38157ab82f842fa1a4106cc4bd94

ramnes:
> @tpatel the CI run failed on the recently added file_element test. This PR does not look related to that particular feature. Might be worth it to investigate

most likely fixed on #161, rebasing

src/chainlit/lc/callbacks.py
willydouhard: yes

willydouhard (approved):

Issue #128, opened 2023-07-03
User Session Reset Mobile Safari


mattflo:
Tested on IOS 16.1.1

Anytime you navigate away from the Safari app and then return, the user session is recreated.

willydouhard:
Should be fixed with the `0.6.0` release

Issue #118, opened 2023-06-30
Websocket reconnection


qyj19920704:
How to return the latest results to the front-end when websocket reconnects

qyj19920704:
![image](https://github.com/Chainlit/chainlit/assets/33174475/03f308a8-78f6-468b-aea2-9ae27e1bc0e9)

qyj19920704:
`const socket = io(wsEndpoint, {
      path: '/ws/socket.io',
      extraHeaders: {
        Authorization: accessToken || '',
        'user-env': JSON.stringify(userEnv)
      },
      timeout: 300000,
    });`

`socket = SocketManager(
    app,
    cors_allowed_origins=[],
    async_mode="asgi",
    max_http_buffer_size=max_http_data_size,
    ping_timeout=300,  # Ping timeout in seconds, increase as needed.
    ping_interval=10,  # Ping interval in seconds
)`

qyj19920704:
> 1111

qyj19920704:
` python
socket = SocketManager(
    app,
    cors_allowed_origins=[],
    async_mode="asgi",
    max_http_buffer_size=max_http_data_size,
    ping_timeout=300,  # Ping timeout in seconds, increase as needed.
    ping_interval=10,  # Ping interval in seconds
)
`

qyj19920704:
![image](https://github.com/Chainlit/chainlit/assets/33174475/7ecb07a7-697d-4a2f-b320-7836367c007e)
![image](https://github.com/Chainlit/chainlit/assets/33174475/720800b2-8e24-4fd7-adba-397510ad4781)

willydouhard:
Does adding a timeout fix your issue?

willydouhard:
Should be fixed with the 0.6.0 release.

PR #159, opened 2023-07-11
persist user sessions on socket reconnection

xsyann:
cc #118 #128


src/chainlit/server.py
```diff
@@ -325,17 +325,13 @@ async def serve(path: str):
...
-    session = sessions.get(id)
-    if not session:
-        raise ValueError("Session not found")
-    return session
-
-
 @socket.on("connect")
-async def connect(sid, environ):
+async def connect(sid, environ, auth):
+    session_id = auth and auth.get("sessionId")
```
willydouhard, line 330: so here if a session id leaked somehow I could restore a session from another user. how likely is it to happen?

willydouhard:
Looks really cool!

src/chainlit/session.py
```diff
@@ -1,29 +1,70 @@
...
-    should_stop: bool
-    # Optional client to authenticate users
-    auth_client: Optional[BaseAuthClient]
-    # Optional client to persist messages and files
-    db_client: Optional[BaseDBClient]
-
-
-sessions: Dict[str, Session] = {}
+    sessions_id: Dict[str, str] = {}
+    sessions: Dict[str, "Session"] = {}
```
ramnes, line 11: would look better with this store outside the class I think

src/chainlit/session.py
```diff
@@ -1,29 +1,70 @@
...
+        session_id = Session.sessions_id.get(key)
+        return Session.get_by_id(session_id)
+
+    @classmethod
+    def get_by_id(self, session_id: str):
+        return Session.sessions.get(session_id)
+
+    @classmethod
+    def require(self, key: str):
+        session = Session.get(key)
```
ramnes, line 67: classmethods should take `cls` as a first argument
```suggestion
    def require(cls, key: str):
        session = cls.get(key)
```

src/chainlit/session.py
```diff
@@ -1,29 +1,70 @@
...
+        self.db_client = db_client
+        self.should_stop = False
+        self.id = token_urlsafe()
+
+        Session.sessions[self.id] = self
+        Session.sessions_id[key] = self.id
+
+    def restore(self, new_key: str):
+        if self.key in Session.sessions_id:
+            Session.sessions_id.pop(self.key)
```
ramnes, line 48:
```suggestion
        Session.sessions_id.pop(self.key, None)
```

src/chainlit/session.py
```diff
@@ -1,29 +1,70 @@
...
+
+    def delete(self):
+        if self.key in Session.sessions_id:
+            Session.sessions_id.pop(self.key)
+        Session.sessions.pop(self.id)
+
+    @classmethod
+    def get(self, key: str):
+        session_id = Session.sessions_id.get(key)
+        return Session.get_by_id(session_id)
```
ramnes, line 59: Wouldn't it make more sense to do the opposite, i.e. storing by key rather than id in the main store? We are going to look by key rather than id all the time except when we try to restore a session, aren't we?

ramnes: Also, why not use two session stores, one that's indexed by id and the other one by key, that both hold the references to the same session objects? That would avoid the intermediary step.

src/chainlit/session.py
```diff
@@ -1,29 +1,70 @@
...
-
-
-sessions: Dict[str, Session] = {}
+    sessions_id: Dict[str, str] = {}
+    sessions: Dict[str, "Session"] = {}
+
+    def __init__(
+        self,
+        # Session key
+        key: str,
```
ramnes, line 16: `key` and `id` are very similar abstract words and they don't make the difference between the two usages explicit. It took me quite a while to understand the difference between them even though you explained me your approach a few hours ago. 😅 What would you think of `socket_id` rather than `key`?

src/chainlit/session.py
```diff
@@ -1,29 +1,70 @@
...
+    @classmethod
+    def get_by_id(self, session_id: str):
+        return Session.sessions.get(session_id)
+
+    @classmethod
+    def require(self, key: str):
+        session = Session.get(key)
+        if not session:
+            raise ValueError("Session not found")
+        return session
```
ramnes, line 70:
```suggestion
        return session
```

ramnes, line 70: Lovely place for a Python 3.8 / [PEP572](https://www.python.org/dev/peps/pep-0572/) inline assignment 😎
```suggestion
    @classmethod
    def require(cls, key: str):
        if session := cls.get(key):
            return session
        raise ValueError("Session not found")
```

src/chainlit/server.py
```diff
@@ -410,36 +407,38 @@ async def connection_successful(sid):
...
 async def disconnect(sid):
-    if sid in sessions:
-        # Clean up the session
-        sessions.pop(sid)
+    async def disconnect_on_timeout(sid):
+        await asyncio.sleep(config.project.session_timeout)
+        if session := Session.get(sid):
+            # Clean up the user session
+            if session.id in user_sessions:
+                user_sessions.pop(session.id)
```
ramnes, line 424:
```suggestion
            user_sessions.pop(session.id, None)
```

src/chainlit/server.py
```diff
@@ -410,36 +407,38 @@ async def connection_successful(sid):
...
+            # Clean up the session
+            session.delete()
```
ramnes, line 426: Don't we want to check if the session was active recently before deleting it? It looks more like a scheduled deletion rather than a timeout here, is it what we want?

src/chainlit/server.py
xsyann: I think the attack surface is pretty low, owasp recommends at least 128 bits unpredictable session id, here we are using 256 bits.
In my opinion, the risk of hijacking a session is pretty the same that getting access to the auth token.
However checking the email can eventually mitigate the risk by the fact that an access token is theoretically short lived.

Do you want to add it in this PR or merge this part and do it in another?

https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html#session-id-length

src/chainlit/session.py
xsyann: The idea was to provide a Session utility not linked to the concept of websocket but as it is tightly coupled and unlikely to change, let's use `socket_id` if it clarifies the code reading :)

src/chainlit/server.py
xsyann: Nice one

src/chainlit/server.py
willydouhard: we can do it in a different one

src/chainlit/server.py
xsyann: Should not be issue as we get the session by `socket_id` before deleting it, if a session has been restored (meaning that it is still used), the `socket_id` is replaced and the session will not be found after the timeout. The deletion will be handled when the new socket will be disconnected.

src/chainlit/socket.py
```diff
@@ -85,6 +100,9 @@ def ask_user_fn(data, timeout):
 @socket.on("connection_successful")
 async def connection_successful(sid):
     session = Session.require(sid)
+    if session.restored:
```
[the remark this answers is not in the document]
willydouhard: we could merge but I initially did this for the following reason:

In `connection_successful`  we potentially call `on_chat_start` which might send messages to the UI. If we do that in the `connect` handler, the connection is not yet accepted in my understanding (since it will be accepted if we return True).

Sending messages while the connection is not accepted seems weird to me.

willydouhard (approved):

PR #154, opened 2023-07-10
Update select and slider inputs UI

alimtunc:

PR #160, opened 2023-07-11
Proof of Concept - Settings in drawer synchronized with backend

MathiasSpanhove:
I really like chainlit and it's simplicity to create LLM applications. However I've found it a bit lacking when you want to have more control as a user. The AskUser method makes it a bit difficult to put constraints on what the user can input and may be confusing.

That's why I've tried to create a new element that can contain other elements, right now I've implemented Checkbox, Radio Buttons, SelectBox, Slider, TextInput and NumberInput. This makes it possible to do stuff like this (See hello.py in the PR):

image

Right now you need to use scoped elements to open the drawer but you could add extra functionality where the Agent is able to open the settings drawer and wait for user input. You would also be able to expose LLM settings like 'temperature', 'frequency_penalty', what model you are using, etc.

At first I wasn't planning on creating a PR since I was just tinkering in the code base so I haven't implemented any tests and I'm sure there are still some bugs that need to be fixed (also my first time really programming in python, any feedback welcome). Depending on your feedback and help I'm willing to create smaller PR's so that Chainlit can natively support this.

Let me know what you think!

Kind Regards,
Mathias

### DIFF
--- a/src/chainlit/__init__.py
+++ b/src/chainlit/__init__.py
@@ -34,6 +34,13 @@ from chainlit.element import (
     TaskStatus,
     Text,
     Video,
+    Checkbox,
+    Radio,
+    Slider,
+    SelectBox,
+    TextInput,
+    NumberInput,
+    Container,
 )
 from chainlit.message import Message, ErrorMessage, AskUserMessage, AskFileMessage
 from chainlit.user_session import user_session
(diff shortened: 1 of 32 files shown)


willydouhard:
This looks really cool! We have some in progress work to update the UI inputs components here #154. Those are not elements yet but just nice looking UI elements.

Also rather that using the sidebar maybe we could have a config panel that we could open with a button in the header. Similar to the settings panel we have today? Or even maybe make the settings panel extendable?

Screenshot 2023-07-13 at 14 50 09

MathiasSpanhove:
Closing this one in favor of #229

PR #164, opened 2023-07-12
Wd/message update

willydouhard:
the update method was badly designed, non exhaustively repeating the arguments on the Message class.

Now one can update the message instance (including actions and elements) and then  call `.update()`.


tpatel (approved):
✨✨✨ AWESOME ✨✨✨

PR #169, opened 2023-07-13
fix avatar using the path parameter

willydouhard:
since avatar was overriding the default send, content was not sent to the UI. To avoid this, I now set content in to_dict


tpatel (approved):
✨✨✨ AWESOME ✨✨✨

tpatel (approved):
right, we only store urls in the db, never the full content 👌

PR #171, opened 2023-07-17
Refactor components folder to be atomic

alimtunc:
Purpose of this PR is to have a cleaner components folder in atomic architecture. 

Capture d’écran 2023-07-17 à 10 29 25


src/chainlit/frontend/src/components/organisms/chat/message/inlined.tsx
```diff
@@ -1,16 +1,16 @@
...
-import InlinedFileList from 'components/element/inlined/files';
-import InlinedImageList from 'components/element/inlined/images';
-import InlinedPDFList from 'components/element/inlined/pdfs';
-import InlinedTextList from 'components/element/inlined/texts';
-import InlinedVideoList from 'components/element/inlined/videos';
-
 import { IAction } from 'state/action';
 import { AllElements, ElementType, IElements } from 'state/element';
 
+import InlinedActionList from '../../../atoms/action/inlined';
```
willydouhard, line 6: Great initiative but it looks live the imports were automatically refactored by your IDE. It would be nice to use absolute imports and not relative imports ie:

../../../atoms/element/inlined/audios -> components/atoms/element/inlined/audios

willydouhard (changes requested):

src/chainlit/frontend/src/components/organisms/chat/message/inlined.tsx
alimtunc: It was intentional. I've experienced issues with absolute imports in the past.. But yes, I can do that if nothing breaks.

willydouhard (approved):

PR #177, opened 2023-07-18
make all initial headers available in the user session

willydouhard:
This is a request from users to be able to authenticate users based on custom headers sent during in the initial connection. They typically use an internal proxy to redirect to the chainlit app.

also adding a custom auth client (like custom db client)


src/chainlit/config.py
```diff
@@ -106,8 +106,16 @@ class CodeSettings:
     on_stop: Optional[Callable[[], Any]] = None
     on_chat_start: Optional[Callable[[], Any]] = None
     on_message: Optional[Callable[[str], Any]] = None
+    lc_agent_is_async: Optional[bool] = None
+    lc_run: Optional[Callable[[Any, str], str]] = None
+    lc_postprocess: Optional[Callable[[Any], str]] = None
+    lc_factory: Optional[Callable[[], Any]] = None
+    lc_rename: Optional[Callable[[str], str]] = None
+    llama_index_factory: Optional[Callable[[], Any]] = None
+    langflow_schema: Union[Dict, str] = None
```
ramnes, line 115: looks like a failed rebase

src/chainlit/server.py
```diff
@@ -160,17 +161,19 @@ def get_html_template():
     <meta property="og:image" content="https://chainlit-cloud.s3.eu-west-3.amazonaws.com/logo/chainlit_banner.png">
     <meta property="og:url" content="{url}">"""
 
+    js = f"""<script>
+    window._headers = {headers};
+    </script>"""
```
ramnes, line 166: looks really insecure to me, that means that any script we load can access the initial headers such as Authorization

ramnes: why not put the headers in the user session on the back end, instead?

ramnes:
I'm rather skeptical about this PR. I like most of the changes (and I think they should be merged right away, most likely in another PR) but the headers part really bugs me.

Maybe I just need more context as to the exact use case it is supposed to cover, but right now I fail to understand the problem it's supposed to solve and why it's our responsibility to handle it. I've never seen something like this; what's another open source and on-premise solution that does this?

src/chainlit/server.py
willydouhard: I agree this was the first idea I had. Maybe we should put it in a cookie, way safer. The use case is I have a chainlit app running in a company secured network. I have a proxy in front of the app using my company auth and I then redirect to the chainlit app with a custom auth header. I then want to be able to access those headers as a developer.

src/chainlit/config.py
willydouhard: nice catch

src/chainlit/server.py
ramnes: but why do we need to do anything on the front end? if I have a proxy in front of my app, all the requests I send to the app are going to have my custom headers

willydouhard: because we need to access them in the websocket connection also. I just pushed a change to remove the headers and  just store them in a cookie. Socket io also sends the cookies.

src/chainlit/server.py
```diff
@@ -308,7 +310,15 @@ async def serve(path: str):
...
                     return FileResponse(file_path)
 
-        return HTMLResponse(content=html_template, status_code=200)
+        html_template = get_html_template()
+
+        response = HTMLResponse(content=html_template, status_code=200)
+
+        response.set_cookie(
+            key="chainlit-headers",
+            value=json.dumps(dict(request.headers)),
```
tpatel, line 319: You might want to switch to HttpOnly cookies. This means the cookies won't be able to be read by the frontend code at all.

```suggestion
            value=json.dumps(dict(request.headers)),
            httponly=True,
```

This is compatible with the current way you are getting the cookie in the socket.io `connect` handler.

tpatel (changes requested):
One small change that will prevent js from accessing the cookie contents

src/chainlit/server.py
willydouhard: Done! Also added a test!

src/chainlit/server.py
ramnes: AFAIK there's no reason that the websocket HTTP handshake request wouldn't have the proxy headers as well. Am I missing something?

src/chainlit/socket.py
```diff
@@ -72,10 +84,13 @@ def ask_user_fn(data, timeout):
 
     user_env = environ.get("HTTP_USER_ENV")
     authorization = environ.get("HTTP_AUTHORIZATION")
+    chainlit_headers = load_chainlit_headers(environ.get("HTTP_COOKIE"))
```
ramnes, line 87: what is this black magic? who's modifying our process environment to put HTTP stuff into it? 😦

src/chainlit/server.py
```diff
@@ -160,17 +162,17 @@ def get_html_template():
     <meta property="og:image" content="https://chainlit-cloud.s3.eu-west-3.amazonaws.com/logo/chainlit_banner.png">
     <meta property="og:url" content="{url}">"""
 
+    js = f"""<script></script>"""
+
```
ramnes, line 166: (not needed anymore I guess)

willydouhard: I suspect this will be useful for other features, such as theme customisation. For that feature we want to inject JS variables (which are not sensitive) so that the theme has the correct colors on the first render (instead of doing an additional http call to get the colors)

src/chainlit/socket.py
willydouhard: environ is not the python process environ, it is coming from the socket io connect callback.

ramnes: mdr

ramnes:
IRL discussion:
* we want to try to get the headers directly inside the "connect" handler – there's no reason it could access the cookies and not the headers, and if the server is behind a proxy, this handler should get any custom header
* worst case scenario, regarding the headers, we should still replace what we're doing with the cookie here with a cookie that only contains a session id, and store the headers on the back end in the session object
* in any case, we still want to implement a session id in a cookie to make the session reconnection implementation more robust, but that's outside the scope of this PR

willydouhard:
I cleaned the interface of `get_auth_client` and `get_db_client`. It now accepts both the initial connection headers (useful in websocket context) and the current http request headers (useful for regular http requests).

src/chainlit/client/utils.py
```diff
@@ -1,26 +1,43 @@
...
 
-async def get_auth_client(authorization: Optional[str]) -> BaseAuthClient:
-    # Check authorization
-    if not config.project.public and not authorization:
-        # Refuse connection if the app is private and no access token is provided
-        trace_event("no_access_token")
-        raise ConnectionRefusedError("No access token provided")
-    elif authorization and config.project.id:
+def load_chainlit_cookie(request: Request):
+    cookie_string = request.cookies.get("chainlit-headers")
```
tpatel, line 14: Where is this cookie set?

src/chainlit/socket.py
```diff
@@ -16,6 +17,17 @@
 from chainlit.user_session import user_sessions
 
 
+def load_chainlit_initial_headers(http_cookie):
+    cookie = SimpleCookie(http_cookie)
+    cookie_string = cookie.get("chainlit-initial-headers").value
+    if cookie_string:
+        chainlit_initial_headers = json.loads(cookie_string)
```
tpatel, line 24: We might want to handle `JSONDecodeError`, not sure.

src/chainlit/user_session.py
```diff
@@ -11,10 +24,10 @@ class UserSession:
     Useful for the developer to store user specific data between calls.
     """
 
-    def get(self, key):
+    def get(self, key, default=None):
```
tpatel, line 27: 👍

src/chainlit/client/utils.py
tpatel: Then you should use the same cookie name, right?

```suggestion
    cookie_string = request.cookies.get("chainlit-initial-headers")
```

tpatel (approved):
✨✨✨ AWESOME ✨✨✨

raonigabriel:
Just wanna say thank you, this will eventually help me a lot!
I was trying to read the "cf-ipcountry" header that gets added by Cloudflare to provide more context to the chatbot.
For an example, by leveraging the users country, we can write custom welcome messages like "Good morning, USA" or "Good Night Japan"

https://developers.cloudflare.com/support/network/configuring-ip-geolocation/

willydouhard:
> Just wanna say thank you, this will eventually help me a lot! I was trying to read the "cf-ipcountry" header that gets added by Cloudflare to provide more context to the chatbot. For an example, by leveraging the users country, we can write custom welcome messages like "Good morning, USA" or "Good Night Japan"
> 
> https://developers.cloudflare.com/support/network/configuring-ip-geolocation/

Awesome, will be available in the next release!

PR #179, opened 2023-07-18
remove factories/rename authors

willydouhard:
This is a breaking change. I propose to remove the factory / run / postprocess paradigm as it is not as meaningful as when we introduced them (we automate less things nowadays).

It was confusing for the users. We do not lose any feature though, everything is doable with `on_chat_start`, `on_message` and `user_session`.  The real integrations are at the callback level anyways (which are now better tested).

`langchain_rename` is now `author_rename` and works for every use case.

It also simplifies the code base.


cypress/support/utils.ts
```diff
@@ -50,22 +50,26 @@ export async function runTest(test: string) {
         console.log("Running with local db");
       }
 
-      return new Promise(async (resolve, reject) => {
+      const promise = new Promise(async (resolve, reject) => {
```
tpatel, line 53: We could go a little bit simpler

```suggestion
      await new Promise(async (resolve, reject) => {
```

cypress/support/utils.ts
```diff
@@ -50,22 +50,26 @@ export async function runTest(test: string) {
...
+              resolve(true);
             } else {
               console.log("Process killed successfully");
               resolve(true);
             }
           });
         }
       });
+
+      await promise;
```
tpatel, line 72: And remove this

```suggestion
```

tpatel (approved):
✨✨✨ AWESOME ✨✨✨

I love the cleaner interface (and less code)

src/chainlit/lc/callbacks.py
```diff
@@ -115,6 +115,23 @@ def check_if_answer_reached(self) -> bool:
...
+
+        streamed_message = Message(
+            author=author,
+            indent=indent,
+            llm_settings=llm_settings,
+            prompt=self.consume_last_prompt(),
+            content="",
+        )
+        self.stream = streamed_message
+
```
ramnes, line 134: 😍

src/chainlit/socket.py
```diff
@@ -98,31 +98,30 @@ def ask_user_fn(data, timeout):
...
-    if isinstance(session.auth_client, CloudAuthClient) and config.project.database in [
-        "local",
-        "custom",
-    ]:
-        await session.db_client.create_user(session.auth_client.user_infos)
-
-    if config.code.on_chat_start:
-        """Call the on_chat_start function provided by the developer."""
-        await config.code.on_chat_start()
+    try:
```
ramnes, line 101: let's do that in another PR and wrap all websocket handlers at once

ramnes (approved):
negative diffs are the best diffs

src/chainlit/socket.py
willydouhard: actually I had to do that to debug 🤣

PR #180, opened 2023-07-18
Put side view into a drawer

alimtunc:
Purpose of this PR is to put the side view components to a drawer, wich will add a better UI when elements is showed and fix the responsive for mobile users.

https://github.com/Chainlit/chainlit/assets/18298759/f8be3608-7858-4584-b26b-908b7c404c71

### DIFF
--- a/src/chainlit/frontend/package-lock.json
+++ b/src/chainlit/frontend/package-lock.json
@@ -34,6 +34,7 @@
         "recoil": "^0.7.6",
         "remark-gfm": "^3.0.1",
         "socket.io-client": "^4.6.2",
+        "usehooks-ts": "^2.9.1",
         "yup": "^1.0.2"
       },
       "devDependencies": {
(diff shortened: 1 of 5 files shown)


willydouhard (changes requested):
Good work! One small bug, if I resize the side view on desktop and then shrink the window and open it on mobile, the width is incorrect.

https://github.com/Chainlit/chainlit/assets/13104895/2f46c493-7f63-4b50-8af0-d68252f491fd

willydouhard (approved):

xleven:
@alimtunc Wonderful work!

There is a little issue that send button will sometimes be blocked by the side view drawer:

https://github.com/Chainlit/chainlit/assets/10850975/8476979a-da23-4967-b3e3-ac1f3eebf4e7

By setting variant to always temporary fixed me locally: https://github.com/Chainlit/chainlit/blob/ba72a06ac733dda65de19a5212d262dfbbff4f30/src/chainlit/frontend/src/components/atoms/element/sideView.tsx#L130

But I agree with willy:
> Ideally the side bar would only float on mobile devices

So maybe a few more tweaks are needed for components like Resizable and Drawer.

PR #182, opened 2023-07-19
Implement a parent-child relationship between messages

ramnes:


ramnes:
Hrm *\*cough\**, tests are passing?

src/chainlit/frontend/src/state/chat.ts
```diff
@@ -33,6 +33,7 @@ export interface IMessage {
   humanFeedback?: number;
   language?: string;
   indent?: number;
+  parentId?: number;
```
willydouhard, line 36: i think this could be `number|string` ?

src/chainlit/lc/callbacks.py
```diff
@@ -155,50 +149,66 @@ def consume_last_prompt(self):
...
+            return self.sequence[-1].author
+        return config.ui.name
 
+    def get_last_message(self):
         if self.sequence:
-            author = self.sequence[-1]
-        else:
-            author = config.ui.name
+            return self.sequence[-1]
+        return self.emitter.session.last_message
```
willydouhard, line 160: i think we should store the default last message when the callback is instantiated (using `self.emitter.session.last_message`).

If `self.sequence` becomes empty after being not empty, then `self.emitter.session.last_message` will have changed

src/chainlit/lc/callbacks.py
```diff
@@ -155,50 +149,66 @@ def consume_last_prompt(self):
...
+        )
 
 
 class LangchainCallbackHandler(BaseLangchainCallbackHandler, BaseCallbackHandler):
+    def on_error(self, error, **_):
+        if error := self.create_error(error):
+            run_sync(error.send())
+            self.pop_sequence()
+
+    on_tool_error = on_error
```
willydouhard, line 194: nice

src/chainlit/lc/callbacks.py
```diff
@@ -249,38 +259,30 @@ def on_llm_end(self, response: LLMResult, **kwargs: Any) -> None:
...
-            self.add_message(outputs[output_key], prompt)
-        self.pop_sequence()
-
-    def on_chain_error(
-        self, error: Union[Exception, KeyboardInterrupt], **kwargs: Any
-    ) -> None:
-        if isinstance(error, InterruptedError):
-            return
-        self.add_message(str(error), error=True)
+            parent_id = self.get_last_message().parent_id
```
willydouhard, line 273: in weird situations get_last_message() could be None

src/chainlit/message.py
```diff
@@ -141,6 +143,7 @@ def __init__(
         prompt: str = None,
         llm_settings: LLMSettings = None,
         language: str = None,
+        parent_id: Any = None,
```
willydouhard, line 146: could be `Union[str, int]`

src/chainlit/socket.py
```diff
@@ -158,13 +160,9 @@ async def process_message(session: Session, author: str, input_str: str):
...
-            await session.db_client.create_message(
-                {
-                    "author": author,
-                    "content": input_str,
-                    "authorIsUser": True,
-                }
-            )
+            await session.db_client.create_message(**message)
+
+        session.last_message = Message.from_dict(message)
```
willydouhard, line 165: I think we should also set last_message on the .send method of the Message class

willydouhard (changes requested):

src/chainlit/message.py
ramnes: What if the user as a custom database client that use another type as an id? (e.g. UUID, MongoDB's ObjectId, etc.)

src/chainlit/socket.py
ramnes: The point of that variable is to keep track of the current processing head, so I don't think we want to overwrite it with the messages we send. That being said, I guess its current name is quite poor and we could rename it with something like `last_user_message`, `last_input`, `current_user_message` or anything else. Ideas welcome! :)

src/chainlit/message.py
ramnes: ok, let's assume ids are ints for now

src/chainlit/lc/callbacks.py
```diff
@@ -116,27 +116,21 @@ def check_if_answer_reached(self) -> bool:
...
 
-        streamed_message = Message(
-            author=author,
-            indent=indent,
-            llm_settings=llm_settings,
-            prompt=self.consume_last_prompt(),
-            content="",
-        )
-        self.stream = streamed_message
+        self.stream = self.create_message(prompt=prompt, author=author, parent_id=parent_id)
```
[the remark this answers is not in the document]
willydouhard: perfect

src/chainlit/lc/callbacks.py
ramnes: Mhh, not if last_message only keeps track of the last user message as it currently does; let's discuss this here: https://github.com/Chainlit/chainlit/pull/182#discussion_r1267855310

src/chainlit/socket.py
willydouhard: I don't think this is necessarily the last user message.

Let's consider this small chainlit app:

```py
import chainlit as cl


@cl.on_chat_start
async def start():
    await cl.Message(
        content="Intermediary steps should be attached on this message",
    ).send()
    
    # Run my langchain agent ...
```

Here no user message is every sent, but we still need to attach the intermediary step messages created by the callback to the "last message sent", in this case the message sent in `on_chat_start`.

ramnes: as discussed, let's always use the last root message or the root message we provide to the callback handler

src/chainlit/message.py
```diff
@@ -213,6 +213,8 @@ async def send(self):
         trace_event("send_message")
         id = await super().send()
 
+        self.emitter.session.root_message = self
```
willydouhard, line 216: should we check i that message has not parent id to set it as the root message?

ramnes: good catch, thanks!

willydouhard (approved):

PR #189, opened 2023-07-19
Inject theme settings to custom UI

alimtunc:
Purpose of this PR is to let user custom UI colors, fonts, etc... via the config.toml

### DIFF
--- a/src/chainlit/config.py
+++ b/src/chainlit/config.py
@@ -61,6 +61,27 @@ hide_cot = false
 # Link to your github repo. This will add a github button in the UI's header.
 # github = ""
 
+# Override default MUI light theme. (Check theme.ts)
+[UI.theme.light]
+    #background = "#FAFAFA"
+    #paper = "#FFFFFF"
+
+    [UI.theme.light.primary]
+        #main = "#F80061"
+        #dark = "#980039"
+        #light = "#FFE7EB"
+
+# Override default MUI dark theme. (Check theme.ts)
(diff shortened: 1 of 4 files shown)


src/chainlit/config.py
```diff
@@ -61,6 +61,19 @@
 # Link to your github repo. This will add a github button in the UI's header.
 # github = ""
 
+[UI.theme_settings]
+#font_family = "Inter"
```
willydouhard, line 65: let's not support font_family for now

src/chainlit/config.py
```diff
@@ -94,6 +123,7 @@ class UISettings:
     hide_cot: bool = False
     default_expand_messages: bool = False
     github: str = None
+    theme_settings: ThemeSettings = None
```
willydouhard, line 126: let's just call it theme, we are already in a class named settings

src/chainlit/frontend/src/theme.ts
```diff
@@ -2,8 +2,24 @@ import { grey, primary, white } from 'palette';
...
 
+type Theme = {
+  primary: string;
+  background: string;
+  paper: string;
+};
+
+declare global {
+  interface Window {
+    theme: {
```
willydouhard, line 13: theme should probably optional like `heme?:`

willydouhard (changes requested):

willydouhard (approved):

PR #190, opened 2023-07-19
move conversations, messages and elements to uuid

willydouhard:


src/chainlit/frontend/package.json
```diff
@@ -42,6 +42,7 @@
     "yup": "^1.0.2"
   },
   "devDependencies": {
+    "@types/uuid": "^9.0.2",
```
ramnes, line 45: ah good catch! didn't realize there was devDependencies here as well

src/chainlit/message.py
```diff
@@ -175,11 +174,9 @@ def from_dict(self, _dict: MessageDict):
...
             language=_dict.get("language"),
             parent_id=_dict.get("parentId"),
-            indent=_dict.get("indent"),
+            indent=_dict.get("indent", 0),
         )
 
-        if temp_id := _dict.get("tempId"):
-            message.temp_id = temp_id
         if _id := _dict.get("id"):
             message.id = _id
```
ramnes, line 181: Is there still a reason to have a condition here?

cypress/e2e/cot/main.py
```diff
@@ -2,10 +2,8 @@
 
 
 @cl.on_message
-async def main():
-    root_message = cl.user_session.get("root_message")  # type: cl.Message
-
-    tool1_msg = cl.Message(content="", author="Tool 1", parent_id=root_message.id)
+async def main(message: str, message_id: str):
+    tool1_msg = cl.Message(content="", author="Tool 1", parent_id=message_id)
```
ramnes, line 6: Why not pass a real message object instead? (We'd then use `message.content` and `message.id`)

src/chainlit/emitter.py
```diff
@@ -83,7 +97,7 @@ async def send_ask_user(
             if self.db_client and res:
                 # If cloud is enabled, store the response in the database/S3
                 if spec.type == "text":
-                    await self.db_client.create_message(res)
+                    await self.process_user_message(res)
```
ramnes, line 100: good catch

src/chainlit/frontend/src/components/organisms/chat/index.tsx
```diff
@@ -54,9 +53,7 @@ const Chat = () => {
         createdAt: new Date().toISOString()
       };
 
-      if (!isAuthenticated || !pSettings?.project?.id) {
-        persistChatLocally(msg);
-      }
+      persistChatLocally(msg);
```
ramnes, line 56: ?

PR #191, opened 2023-07-19
Fix for element names containing regex characters

MathiasSpanhove:
The following simple program containing scoped elements would not render the links correctly:

```python
import chainlit as cl

@cl.on_chat_start
async def main():
    element1 = cl.Text(name="Document[1]", content="Testing", display="side")
    element2 = cl.Text(name="Source(45)", content="Testing", display="side")

    await cl.Message(
        content=f"Testing: Document[1] or Source(45)", elements=[element1, element2]
    ).send()
```

I've added code that escapes regex characters so that the element names are matched correctly.


tpatel (approved):
✨✨✨ AWESOME ✨✨✨

ramnes (approved):
LGTM

PR #201, opened 2023-07-20
only serve public and ui dir

willydouhard:


src/chainlit/server.py
```diff
@@ -301,14 +307,14 @@ def register_wildcard_route_handler():
...
-
-            for file_path in file_paths:
-                if os.path.isfile(file_path):
-                    return FileResponse(file_path)
+            ui_file_path = Path(build_dir) / path
+
+            # Check if the file is within the intended directory
+            if (
+                ui_file_path.is_file()
+                and ui_file_path.resolve().parent in ui_file_path.resolve().parents
```
MathiasSpanhove, line 314: I don't think this does anything since the parent will always be in all the parents?

willydouhard: You are correct, I mean `and Path(build_dir).resolve() in ui_file_path.resolve().parents
`

MathiasSpanhove: LGTM!

MathiasSpanhove:
I wonder if it's worth it to mount all static files from chainlit like this:
````python 
app.mount("/static", StaticFiles(packages=[('chainlit', 'frontend/dist')]), name="dist")
```
You would need to change the urls of all the assets and favicon to include '/static' though

willydouhard:
> I wonder if it's worth it to mount all static files from chainlit like this:
> 
> ```python
> app.mount("/static", StaticFiles(packages=[('chainlit', 'frontend/dist')]), name="dist")
> ```
> 
> You would need to change the urls of all the assets and favicon to include '/static' though

I did not know about the package attribute. I'll try it!

MathiasSpanhove:
The following seems to work without any changes:

```python
app.mount("/assets", StaticFiles(packages=[('chainlit', 'frontend/dist/assets')]), name="assets")
app.mount("/favicon.svg", StaticFiles(packages=[('chainlit', 'frontend/dist')]), name="favicon")
```

willydouhard:
Fixed the favicon, the issue was that it was served as a directory where it is actually a file

PR #209, opened 2023-07-21
Improve element type inference

MathiasSpanhove:
I made some changes to the `element.ts` to make typescript smarter causing less `as` casts. 
I also separated Avatar and TaskList from the other elements so that you don't need to write logic to extract them from the other elements.

Open to any feedback!

### DIFF
--- a/src/chainlit/frontend/src/api/index.ts
+++ b/src/chainlit/frontend/src/api/index.ts
@@ -2,7 +2,7 @@ import { IPageInfo, IPagination } from 'components/organisms/dataset/table';
 
 import { IChat, ILLMSettings } from 'state/chat';
 import { IDatasetFilters } from 'state/dataset';
-import { IElement } from 'state/element';
+import { IMessageElement } from 'state/element';
 import { IMember, Role } from 'state/user';
 
 const devServer = 'http://127.0.0.1:8000';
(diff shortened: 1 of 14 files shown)


willydouhard:
LGTM, @tpatel could you review also?

tpatel:
I love that you've moved the Avatar and Tasklist states outside of the element state ❤️ 

Can you share why you've moved the optional url from the `TElement` to each element?

MathiasSpanhove:
> I love that you've moved the Avatar and Tasklist states outside of the element state ❤️
> 
> Can you share why you've moved the optional url from the `TElement` to each element?

I originally did that because I added Input Widget elements (see PoC PR) that didn't have a url property. Right now you can probably put it back into the base class but I imagine that in the future elements will arrive again that don't have a url.

tpatel (changes requested):
Thanks for sharing your explanation! Could you move the url back into the element?
The reason is that at the moment, we are setting a url when we persist an element: https://github.com/Chainlit/chainlit/blob/ba72a06ac733dda65de19a5212d262dfbbff4f30/src/chainlit/element.py#L86
So all elements become files with a url, and you use the element url when viewing a conversation in the `history` tab.

MathiasSpanhove:
> Thanks for sharing your explanation! Could you move the url back into the element? The reason is that at the moment, we are setting a url when we persist an element:
> 
> https://github.com/Chainlit/chainlit/blob/ba72a06ac733dda65de19a5212d262dfbbff4f30/src/chainlit/element.py#L86
> 
> So all elements become files with a url, and you use the element url when viewing a conversation in the `history` tab.

I've added url back into the element.

tpatel (approved):
✨✨✨ AWESOME ✨✨✨

Thanks for your contribution!

Note: One side effect of these changes is that you can't have a link (from a message) to a tasklist element anymore, and this is an improvement 👍

PR #211, opened 2023-07-21
update the guide

willydouhard:


MathiasSpanhove:
LGTM!

Some extra tips that could be added:

If you are on windows, some npm commands like `npm run formatPython` won't work. You can fix this by changing the npm script-shell to bash: `npm config set script-shell "C:\\Program Files\\git\\bin\\bash.exe"` (default x64 install location)

If you don't want to always add `poetry run` before the chainlit command you can start a shell first using `poetry shell`.

willydouhard:
> LGTM!
> 
> Some extra tips that could be added:
> 
> If you are on windows, some npm commands like `npm run formatPython` won't work. You can fix this by changing the npm script-shell to bash: `npm config set script-shell "C:\\Program Files\\git\\bin\\bash.exe"` (default x64 install location)
> 
> If you don't want to always add `poetry run` before the chainlit command you can start a shell first using `poetry shell`.

This should be described in an optional step in the guide right?

ramnes (approved):
LGTM

.github/CONTRIBUTING.md
```diff
@@ -75,75 +80,47 @@ poetry install
...
-## Start the UI
+## Start the Chainlit server from source
+
+You need to build the UI once before starting the server.
 
 ```sh
-pnpm run buildUi
-cd src/chainlit/frontend
-pnpm run dev -- --port 5174
+npm run buildUi
```
ramnes, line 88: pnpm

willydouhard: I think `pnpm` is required for installation. We can still use `npm` other commands. Typically I tried to run `pnpm run dev -- --port 5174` but it was not accepting the port argument.

alimtunc: Yes, you should remove the extra `--`. My bad, did not see this when I replaced npm.
Capture d’écran 2023-07-24 à 13 52 20

PR #212, opened 2023-07-21
Don't share the session id between tabs

ramnes:


src/chainlit/frontend/src/hooks/clearChat.ts
```diff
@@ -1,20 +1,23 @@
...
   const setActions = useSetRecoilState(actionState);
   const setSideView = useSetRecoilState(sideViewState);
   const setTokenCount = useSetRecoilState(tokenCountState);
+  const resetSessionId = useResetRecoilState(sessionIdState);
   const session = useRecoilValue(sessionState);
 
   return () => {
     session?.socket.emit('clear_session');
     session?.socket.disconnect();
+    resetSessionId();
```
willydouhard, line 20: does reset clear the variable or create a new uuid?

willydouhard:
Just thought about it but we should clear the sessions when the[ file watcher picks up a change](https://github.com/Chainlit/chainlit/blob/main/src/chainlit/server.py#L86) in the user developer's code. Currently we just reload the config and the browser, but now the session will be resumed. As a developer I would expect to start from a fresh start every time I change something in my code.

src/chainlit/frontend/src/hooks/clearChat.ts
ramnes: I though it would create a new uuid but that was dumb, it's obviously reusing the same uuid, looking for a workaround right now. :)

ramnes: @willydouhard should be better now

willydouhard (approved):

PR #224, opened 2023-07-25
Parse custom theme json before injecting it to front

alimtunc:
`None` is not supported by JavaScript. We should replace it by `undefined` or the front won't set custom theme if there is one `None` value Capture d’écran 2023-07-25 à 10 03 37

### DIFF
--- a/src/chainlit/server.py
+++ b/src/chainlit/server.py
@@ -1,3 +1,4 @@
+import json
 import mimetypes
 
 mimetypes.add_type("application/javascript", ".js")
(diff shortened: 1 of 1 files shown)


src/chainlit/server.py
```diff
@@ -180,7 +180,7 @@ def get_html_template():
         content = f.read()
         content = content.replace(PLACEHOLDER, tags)
         if js:
-            content = content.replace(JS_PLACEHOLDER, js)
+            content = content.replace(JS_PLACEHOLDER, js.replace("None", "undefined"))
```
willydouhard, line 183: this is not enough as True and False will also be problematic. can we try something like

```py
js = f"""<script>window.theme = {json.dumps(config.ui.theme.to_dict())}</script>"""
```

alimtunc: Looks better!

ramnes (approved):
LGTM

PR #231, opened 2023-07-27
add a Tags input

tpatel:
- One use case is for example configuring the list of stopwords for LLMs
- Also renamed "SelectInput" to "Select" as they are exposed throught "input_widget"
- Moved the ChatSettings in a separate file to enable using the input widgets outside of the chat settings

### DIFF
--- a/src/chainlit/input_widget.py
+++ b/src/chainlit/input_widget.py
@@ -4,7 +4,6 @@ from typing import Any, Dict, List
 
 from pydantic.dataclasses import Field, dataclass
 
-from chainlit.context import get_emitter
 from chainlit.types import InputWidgetType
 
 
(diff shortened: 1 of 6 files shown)


src/chainlit/input_widget.py
```diff
@@ -220,3 +191,36 @@ def to_dict(self) -> Dict[str, Any]:
...
+class Tags(InputWidget):
+    """Useful to create an input for an array of strings."""
+
+    type: InputWidgetType = "tags"
+    initial: List[str] = Field(default_factory=lambda: list)
+    values: List[str] = Field(default_factory=lambda: list)
+
+    def __init__(
+        self,
+        key: str or int,
```
MathiasSpanhove, line 206: Do we want to allow int for the other widget keys as well? When we get the settings back from the front-end is the key still an integer or is it converted to a string by formik?

src/chainlit/input_widget.py
```diff
@@ -220,3 +191,36 @@ def to_dict(self) -> Dict[str, Any]:
...
+        label: str,
+        initial: List[str] = None,
+        tooltip: str = None,
+        description: str = None,
+    ) -> None:
+        self.key = key
+        self.label = label
+        if initial is not None:
```
MathiasSpanhove, line 214: Does this if statement do anything? if it is None it will just stay None no?
If it does something, should we also add this check to the other widgets?

MathiasSpanhove:
Looks great! I like the Tags Input addition!

src/chainlit/input_widget.py
MathiasSpanhove: Nvm looked over the fact that initial is a list in this case

MathiasSpanhove (commented):
Looks great! I like the Tags Input addition!

src/chainlit/input_widget.py
MathiasSpanhove: I originally looked at [streamlit](https://docs.streamlit.io/library/api-reference/widgets/st.text_input) to find inspiration about which properties should be supported and they allow the key to be both int or string, but apparently forgot to remove it properly since I found it confusing

tpatel: Ok, thanks for sharing your process! In that case it makes sense to move to string-only 👌

PR #232, opened 2023-07-27
At/uniformize front widget

alimtunc:
Purpose of this PR : 
 - Uniformize all inputs with common types
 - Add FormInputs component to have a generic component to render inputs


src/chainlit/frontend/src/components/organisms/FormInput.tsx
```diff
@@ -0,0 +1,60 @@
+import Switch, { SwitchProps } from 'components/atoms/switch';
```
MathiasSpanhove, line 1: Shouldn't we group the inputs together in an inputs directory? They are currently in different directories or is there a reason for doing it this way?

src/chainlit/frontend/src/components/organisms/FormInput.tsx
```diff
@@ -0,0 +1,60 @@
...
+  type: T;
+  initial: I;
+}
+
+export type TFormInput =
+  | (SwitchProps & IFormInput<'switch', boolean>)
+  | (SliderProps & IFormInput<'slider', number>)
+  | (SelectInputProps & IFormInput<'select', string>)
+  | (TextInputProps & IFormInput<'textinput', string>)
+  | (TagsInputProps & IFormInput<'tags', string>);
```
MathiasSpanhove, line 23: Shouldn't this one be string[]?

src/chainlit/frontend/src/components/organisms/FormInput.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const FormInput = ({ element }: { element: TFormInput }): JSX.Element => {
+  const renderInput = () => {
+    switch (element.type) {
+      case 'select':
+        return (
+          <SelectInput
+            {...element}
+            items={element?.items?.map((option: SelectItem) => ({
```
MathiasSpanhove, line 32: Is the element ever going to be null here?

src/chainlit/frontend/src/components/organisms/FormInput.tsx
```diff
@@ -0,0 +1,60 @@
...
+              name: element.id
+            }}
+          />
+        );
+      case 'textinput':
+        return <TextInput {...element} placeholder={element.placeholder} />;
+    }
+  };
+
+  return renderInput() || <></>;
```
MathiasSpanhove, line 57: This should never happen and typescript is able to validate this if we add `"strictNullChecks": true,` in the `tsconfig.json` file.

Change the function to this:
```js
const FormInput = ({ element }: { element: TFormInput }): JSX.Element => {  
  switch (element.type) {
    case 'select':
      return (
        <SelectInput
          {...element}
          items={element.items?.map((option: SelectItem) => ({
            label: option.label,
            value: option.value
          }))}
        />
      );
    case 'slider':
      return <Slider {...element} />;
    case 'tags':
      return <TagsInput {...element} />;
    case 'switch':
      return (
        <Switch
          {...element}
          inputProps={{
            id: element.id || undefined,
            name: element.id
          }}
        />
      );
    case 'textinput':
      return <TextInput {...element} placeholder={element.placeholder} />;
  }
};
```

And now when you forget or add a new type, typescript will yell at you to implement the component :D

src/chainlit/frontend/src/components/atoms/switch.tsx
```diff
@@ -3,27 +3,48 @@ import { green } from 'palette';
...
+} & IInput;
 
 export default function Switch(props: SwitchProps): JSX.Element {
-  const { checked, disabled, inputProps, onChange } = props;
+  const {
+    checked,
+    description,
+    disabled,
+    hasError,
+    id,
```
MathiasSpanhove, line 25: I noticed that some current Switch implementations now have errors since id is required

src/chainlit/frontend/src/components/organisms/slider.tsx
```diff
@@ -1,10 +1,35 @@
...
+import { IInput } from 'types/Input';
+
+import InputStateHandler from './inputs/inputStateHandler';
+
+export type SliderProps = IInput & MSliderProps;
+
+const _Slider = ({
+  description,
+  hasError,
+  id,
```
MathiasSpanhove, line 15: I noticed that some current Slider implementations now have errors since id is required

src/chainlit/frontend/src/components/organisms/FormInput.tsx
```diff
@@ -0,0 +1,60 @@
...
+      case 'slider':
+        return <Slider {...element} />;
+      case 'tags':
+        return <TagsInput {...element} />;
+      case 'switch':
+        return (
+          <Switch
+            {...element}
+            inputProps={{
+              id: element.id || undefined,
```
MathiasSpanhove, line 47: element.id is required, is undefined needed here?

src/chainlit/frontend/src/components/organisms/inputs/inputStateHandler.tsx
```diff
@@ -2,23 +2,35 @@ import { Box, FormControl, FormHelperText } from '@mui/material';
...
+    id,
+    label,
+    notificationsCount,
+    tooltip
+  } = props;
 
   return (
     <Box width="100%">
-      {label ? <InputLabel id={id} label={label} tooltip={tooltip} /> : null}
+      {label ? (
```
MathiasSpanhove, line 26: Minor suggestion, I like to use condition && component instead of condition ? component : null

src/chainlit/frontend/src/components/organisms/FormInput.tsx
alimtunc: No, even the items should not be null if we use Select

src/chainlit/frontend/src/components/organisms/inputs/inputStateHandler.tsx
alimtunc: It's a better practice to render null to be sure react will not render unwanted value. https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

src/chainlit/frontend/src/components/organisms/FormInput.tsx
alimtunc: Yes it can be, it can be an issue for later

src/chainlit/frontend/src/components/organisms/FormInput.tsx
MathiasSpanhove: But element should never ever be null, you can have a dynamic list of elements sure but allowing element null to be passed to the forminput is just not going to happen right?

MathiasSpanhove: And if you add a default, typescript won't warn you anymore when you forget to implement a forminput

alimtunc: We can't be sure that backend will send us a valid element type, what if backend accidentely send wrong type ? We can't trust typescript to fallback runtime check.

alimtunc: And I don't see why typescript will not warn if the element is incorrect ? I've just had a fallback on the switch, the types are the same, it isn't null :)

MathiasSpanhove: > 



> We can't be sure that backend will send us a valid element type, what if backend accidentely send wrong type ? We can't trust typescript to fallback runtime check.

Fair enough, I would imagine that is a major bug in the backend and normally I wouldn't consider this case but I can see why you want to do it. Would it be possible to add a typescript guard still like the following: https://medium.com/technogise/type-safe-and-exhaustive-switch-statements-aka-pattern-matching-in-typescript-e3febd433a7a

alimtunc: We can discuss about it yes, why not.

willydouhard (approved):

PR #233, opened 2023-07-27
add backend unit tests

tpatel:
- The unit tests use pytest
- Also fixes running mypy outside of a poetry shell

This is a first step to unlock more unit tests on the python side.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -30,6 +30,8 @@ jobs:
 
       - name: Install Poetry
         run: pip install poetry
+      - name: Install dependencies
+        run: poetry install --with tests
       - name: Run tests
         env:
           CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
--- a/package.json
+++ b/package.json
@@ -12,11 +12,11 @@
   },
   "scripts": {
     "preinstall": "npx only-allow pnpm",
-    "test": "pnpm exec ts-node ./cypress/support/e2e.ts && (cd src/chainlit/frontend && pnpm run test)",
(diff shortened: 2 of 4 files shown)


src/chainlit/client/local.py
```diff
@@ -95,8 +95,13 @@ async def delete_conversation(self, conversation_id):
     async def get_conversation(self, conversation_id: str):
         from prisma.models import Conversation
 
+        try:
```
willydouhard, line 98: conversation id should be a string, why cast it?

willydouhard (changes requested):

src/chainlit/client/local.py
tpatel: I blindly trusted mypy which was telling me the variable needed to be an integer, I shouldn't have trusted it, it was weird to do this change.
I'm gonna dive into the issue and why it complains on my computer and not in the CI

tpatel: It wasn't the mypy cache, nor the pycache, it was the cached prisma schema.
@willydouhard I'm not sure why the cached schema wasn't re-generated. I had to delete the sqlite db to get a new cached schema definition.

willydouhard: When the schema is changed, you have to run chainlit migrate. It will not be done automatically

tpatel: We could add some kind of error to prevent running updated code with an outdated generated prisma setup, that would make it more obvious, wdyt?

willydouhard: Not sure how we would do that. Any idea?

tpatel:
replaced with #235

PR #235, opened 2023-07-28
Add backend unit tests

tpatel:
- The unit tests use pytest
- Also fixes running mypy outside of a poetry shell

This is a first step to unlock more unit tests on the python side.


tpatel:
this is outdated

PR #246, opened 2023-07-31
FormInput improvements

MathiasSpanhove:
This PR contains several changes:

- I changed the InputLabel id={id} to htmlFor={id} to make sure that the labels are linked to the inputs. I was unable to make it work using the id attribute. (Normally MUI does this for us using the id attribute, but this didn't work for some reason)
- Changes to FormInput as discussed to make sure that it is always in a controlled state (and won't switch to uncontrolled if passed null or undefined as a value)
- Some fixes to make sure that formik.handleChange works.
- I had to use formik.setField for the tagsInput since the mui-chips-input doesn't expose an onChange event handle. Not really happy about my solution since now tagsInput 'knows' about Formik. 
- I added 'strictNullChecks' to the tsconfig.json file and added a check to make sure that the FormInput switch case throws an error if a case is not implemented.

Any feedback welcome!

### DIFF
--- a/src/chainlit/frontend/src/components/organisms/FormInput.tsx
+++ b/src/chainlit/frontend/src/components/organisms/FormInput.tsx
@@ -7,46 +7,45 @@ import TagsInput, { TagsInputProps } from './inputs/tagsInput';
 import TextInput, { TextInputProps } from './inputs/textInput';
 import Slider, { SliderProps } from './slider';
 
-export type FormInitial =
-  | string
-  | number
-  | boolean
-  | string[]
-  | number[]
-  | undefined;
+export type TFormInputValue = string | number | boolean | string[] | undefined;
 
-export interface IFormInput<T> extends IInput {
+export interface IFormInput<T, V extends TFormInputValue> extends IInput {
   type: T;
+  value?: V;
+  setField?(field: string, value: V, shouldValidate?: boolean): void;
 }
 
 export type TFormInput =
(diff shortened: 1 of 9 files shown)


src/chainlit/frontend/src/components/organisms/inputs/tagsInput.tsx
```diff
@@ -28,7 +30,16 @@ export default function TagsInput({
       label={label}
       tooltip={tooltip}
     >
-      <MuiChipsInput {...rest} size={size} name={id} id={id} />
+      <MuiChipsInput
+        disabled={disabled}
+        size={size}
```
alimtunc, line 35: Should be in `rest` so we can remove it here and at upper

src/chainlit/frontend/src/components/organisms/FormInput.tsx
```diff
@@ -7,46 +7,45 @@ import TagsInput, { TagsInputProps } from './inputs/tagsInput';
...
-      return <Slider {...element} />;
+      return <Slider {...element} value={element.value ?? 0} />;
     case 'tags':
-      return <TagsInput {...element} />;
+      return <TagsInput {...element} value={element.value ?? []} />;
     case 'switch':
       return (
         <Switch
           {...element}
+          checked={element.value ?? false}
```
alimtunc, line 37:
```suggestion
          checked={!!element.value}
```

alimtunc (approved):
Some request but LGTM, good job :)

PR #252, opened 2023-08-03
Only run pre-commit on staged files (except for mypy)

alimtunc:
This PR consist to reduce the time on pre-commit, since we only need to lint/format staged files, its useless to run it on all files when we want to commit. 

Reducing the time commit by 3.

Before ~15 seconds :
Capture d’écran 2023-08-03 à 16 30 52

After ~6 seconds :
Capture d’écran 2023-08-03 à 16 31 28

And could be better if we run mypy only on staged files.

Had some issues with `tsc` actually.
Also with `mypy` wich need to be on the folder to be run and looks like to not working properly with lint-staged.. any idea @ramnes @willydouhard  ?


src/chainlit/frontend/src/App.tsx
```diff
@@ -97,7 +97,7 @@ function App() {
     if (isAuthenticated && accessToken === undefined) {
       getAccessTokenSilently({
         authorizationParams: {
-          audience: 'chainlit-cloud'
+          audience: 'chainlit-clod'
```
willydouhard, line 100: Why are we changing the audience parameter? Looks like a typo

willydouhard (changes requested):

willydouhard:
where does the `lint-staged` command comes from? eslint directly?

alimtunc:
> where does the `lint-staged` command comes from? eslint directly?

Running with npx will install it on first run, but it should be in the package, my bad. I fix it.

willydouhard (approved):

PR #253, opened 2023-08-03
optimize callback imports

xsyann:


src/chainlit/cache.py
```diff
@@ -2,14 +2,19 @@
...
+    from chainlit.lc import LANGCHAIN_INSTALLED
+
+    return LANGCHAIN_INSTALLED
+
+
 def init_lc_cache():
     use_cache = config.run.no_cache is False and config.run.ci is False
 
-    if LANGCHAIN_INSTALLED and use_cache:
+    if use_cache and is_langchain_installed():
```
ramnes, line 17: I'd just import LANGCHAIN_INSTALLED within that function

xsyann: Doing it so avoids importing and version checking of langchain if cache is not used

willydouhard (approved):

PR #257, opened 2023-08-07
fix: mui warning on Tooltip component wrap a disabled Button

chenjuneking:
When a disabled `Button` was provided to the `Tooltip` component, a mui warning will occur:

```txt
Tooltip.js:276 MUI: You are providing a disabled `button` child to the Tooltip component.
A disabled element does not fire events.
Tooltip needs to listen to the child element's events to display the title.

Add a simple wrapper element, such as a `span`.
```

Noted: This warning text display on development, not on production.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ venv
 
 .DS_Store
 
+.chainlit
 chainlit.md
 
 cypress/screenshots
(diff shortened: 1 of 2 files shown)


willydouhard (approved):
Thanks for the contribution!

PR #258, opened 2023-08-07
feat: action callback handle multiple actions

chenjuneking:
Support the feature: 

Handle multiple actions on one action callback.

I want to generate and display some relative questions dynamically after the user ask a question. On this scene, the actions and callbacks was not 1v1 matched, it's better to design the abilitility of handling all the actions on one action callback.

the application as follow:

```py
import chainlit as cl

@cl.action_callback("ask_for_question")
async def on_action(action):
    the_question = action.value
    await cl.Message(content=f"You ask: {the_question}").send()
    # Ask the question to the LLMs
    # ...
    await action.remove()

@cl.on_chat_start
async def start():
    # The questions
    questions = ["Question one", "Question two", "Question three"]
    actions = []
    for question in questions:
        actions.append(cl.Action(name="ask_for_question", value=question, description="Click to ask the question"))

    await cl.Message(content="Relative questions:", actions=actions).send()
```

In order to support this feature, the `action.name` was no longer to be the `key` of the `Action` component, because the property `action.name` had the same value, so we need a new property `action.id` to identify the action.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ venv
 
 .DS_Store
 
+.chainlit
 chainlit.md
 
 cypress/screenshots
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 Nothing unreleased yet!
 
+## [0.6.2] - 2023-08-06
+
+### Added
+
+- Make the chat experience configurable with Chat Settings
+- Authenticate users based on custom headers with the Custom Auth client
+
+### Fixed
+- Author rename now works with all kinds of messages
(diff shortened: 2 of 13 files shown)


willydouhard:
Interesting! For completion, is the instantiated action when the [server receives an action call](https://github.com/Chainlit/chainlit/blob/main/src/chainlit/socket.py#L225) going to have the correct ID?

chenjuneking:
> Interesting! For completion, is the instantiated action when the [server receives an action call](https://github.com/Chainlit/chainlit/blob/main/src/chainlit/socket.py#L225) going to have the correct ID?

Yes, I had double check, both the server and the client could receive the action with the correct ID, and some unit tests about this feature had beed commited.

willydouhard (approved):
Great work!

PR #260, opened 2023-08-07
add copy to clipboard button to message

pixelkaiser:
I was missing a copy to clipboard button.


ramnes (approved):
LGTM

willydouhard:
Looks like the action test is failing, probably because a new button is added and we naively expect a certain button count

pixelkaiser:
> Looks like the action test is failing, probably because a new button is added and we naively expect a certain button count

I tried running the test suite on macos but Cypress is acting up. Can someone pick this up?

willydouhard:
You can fix the CI by changing the expected number on this [line](https://github.com/Chainlit/chainlit/blob/main/cypress/e2e/action/spec.cy.ts#L10) to `4`. However I would like to set this PR as draft for the time being for two reasons:
- Another PR is introducing a `copy-button` element and I would like to avoid duplicates
- I think it is a good time to think on how actions and buttons are displayed together in the ui. Not entirely satisfied by this:
Screenshot 2023-08-09 at 13 32 06

pixelkaiser:
In my experience keeping small PRs open for an extended period without clear progress usually leads to issues like: Merge conflicts, Clutter in the project tracking, Potential duplication of efforts (which we see here), Alignment with project direction. This was meant to be a small patch to make my downstream life easier and I do not plan to maintain the PR for an extended period of time.

I think closing it will keep your workflow streamlined and focused on actual priorities.

PR #261, opened 2023-08-07
run cypress once

xsyann:
This PR allows to run all Cypress specs at once instead of running them one by one.

It is based on the anti-pattern of re-starting Chainlit with `cy.exec` before running each test and thus come with some limitations which are listed here:
https://docs.cypress.io/guides/references/best-practices#Web-Servers

Unfortunately, as we can't run all tests on the same pre-started Chainlit app, this is the best solution we have for now.

Screenshot 2023-08-09 at 15 34 06

`SKIP_BUILD` and `SINGLE_TEST` have been preserved and `SINGLE_TEST` can even now be set as a regexp to run multiple isolated tests.

### DIFF
--- a/cypress/support/e2e.ts
+++ b/cypress/support/e2e.ts
@@ -1,20 +1,15 @@
 import * as dotenv from "dotenv";
 dotenv.config();
-import { installChainlit, runTests, runTest } from "./utils";
+import { installChainlit, runTests } from "./utils";
 
 async function main() {
-  const singleTest = process.env.SINGLE_TEST;
+  const matchName = process.env.SINGLE_TEST || "*";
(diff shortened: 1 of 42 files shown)


cypress/support/e2e.ts
```diff
@@ -10,11 +10,11 @@ async function main() {
     await installChainlit();
   }
 
-  if (singleTest) {
-    await runTest(singleTest);
-  } else {
+  // if (singleTest) {
```
willydouhard, line 13: do we still support single test runs?

xsyann: It is of course planned to restore this behavior :) I think we can even add regex support without effort.

willydouhard (approved):

PR #263, opened 2023-08-08
fix: format command failed on windows

chenjuneking:
The logs of the error:

```sh
$ cd src/chainlit/frontend/ && pnpm format

> chainlit-ui@0.0.1 format D:\projects\chainlit\src\chainlit\frontend
> prettier --config .prettierrc 'src/**/*.{ts,tsx}' --write --loglevel error

[error] No files matching the pattern were found: "'src/**/*.{ts,tsx}'".
 ELIFECYCLE  Command failed with exit code 2.
(venv)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ venv
 
 .DS_Store
 
+.chainlit
 chainlit.md
 
 cypress/screenshots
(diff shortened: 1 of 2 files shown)


willydouhard:
LGTM, @ramnes @alimtunc @xsyann any reason we wrapped the path with quotes?

chenjuneking:
BTW, my OS was `Windows11` + `Git-Bash`

alimtunc:
> LGTM, @ramnes @alimtunc @xsyann any reason we wrapped the path with quotes?

I don't think so, this just make sense for a path to be in quotes generally

PR #265, opened 2023-08-08
adding litellm integration - support for anthropic, azure, cohere, replicate

krrishdholakia:
Hey @ramnes @alimtunc ,

Chainlit is pretty cool. I'm working on litellm (simple library to standardize LLM API Calls - https://github.com/BerriAI/litellm) and was wondering if we could be helpful.

Added support for Claude, Cohere, Azure and Llama2 (via Replicate) by adding an additional integration file. The code is pretty similar to the OpenAI class - as litellm follows the same pattern as the openai-python sdk.

Would love to know if this helps.

Happy to add additional tests / update documentation, if the initial PR looks good to you.

### DIFF
--- a/cypress/e2e/litellm/.chainlit/config.toml
+++ b/cypress/e2e/litellm/.chainlit/config.toml
@@ -1,0 +1,34 @@
+[project]
+# If true (default), the app will be available to anonymous users.
+# If false, users will need to authenticate and be part of the project to use the app.
+public = true
+
+# The project ID (found on https://cloud.chainlit.io).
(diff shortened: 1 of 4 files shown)


willydouhard:
This is really cool! To me it looks more like an example of how to use litellm with Chainlit than an integration per say (for instance the llama index integration is really about the callback that automatically plugs to llama to display chain of thoughts).

Maybe it will be more suitable to have this in the [cookbook](https://github.com/Chainlit/cookbook)?

PR #270, opened 2023-08-10
At/remove useless useeffect

alimtunc:
This PR consist to remove useless useEffects wich can leads to bugs or useless re-renders.

Good to read : https://react.dev/learn/you-might-not-need-an-effect

### DIFF
--- a/src/chainlit/frontend/src/components/atoms/element/view.tsx
+++ b/src/chainlit/frontend/src/components/atoms/element/view.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useState } from 'react';
 import { useParams } from 'react-router-dom';
 import { useRecoilValue } from 'recoil';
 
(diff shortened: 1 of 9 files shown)


tpatel (commented):
A couple of comments

src/chainlit/frontend/src/components/atoms/element/view.tsx
```diff
@@ -38,32 +38,27 @@ const ElementView = () => {
...
+      if (foundElement) {
+        element = foundElement;
       }
     } else {
       client
         .getElement(conversationId, id)
-        .then((element) => {
-          setElement(element);
-        })
+        .then((res) => (element = res))
```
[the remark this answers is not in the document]
alimtunc: Yes, that's a good point.

src/chainlit/frontend/src/components/organisms/chat/message/message.tsx
```diff
@@ -113,7 +113,13 @@ const Message = ({
             <DetailsButton
               message={message}
               opened={showDetails}
-              onClick={() => setShowDetails(!showDetails)}
+              onClick={() => {
+                setShowDetails(!showDetails);
+                setAppSettings((old) => ({
+                  ...old,
+                  expandAll: false
```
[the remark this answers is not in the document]
alimtunc: You're right. I wanted to fix another side effect I had, but I should not reset it on on click detail.

src/chainlit/frontend/src/components/organisms/chat/message/message.tsx
tpatel, line 1: You might want to revert the changes on this file:

- `setAppSettings` isn't used
- `setShowDetails` doesn't get updated when changing the settings (the settings are in a modal so it would make sense to update the state right away)
image

tpatel (changes requested):
just one more comment 😇

src/chainlit/frontend/src/components/organisms/chat/message/message.tsx
alimtunc: Ok I finally revert it, kind of annoying to get the actual state. But these settings shouldn't be persistant ?

tpatel: I think the settings are persistent. They can be changed while showing messages thought, so it's better if the user sees the change right away.

tpatel (approved):
✨✨✨ AWESOME ✨✨✨

src/chainlit/frontend/src/components/organisms/chat/message/message.tsx
alimtunc: Nop, they're not, if you refresh the page, this will be reset..

tpatel: This might be a bug, only the last setting is saved. I'll ask around for more context on these settings.

PR #277, opened 2023-08-14
Make callbacks available at the main module level again, but lazily

ramnes:
This keeps the imports fast, but with a few advantages over the current implementation:

1. we don't introduce a breaking change to the end users;
2. it's easier to import `chainlit` once and getting all things from there;
3. doing so also makes it easier to read the code and understand what comes from chainlit: `cl.LangchainCallbackHandler` makes it obvious that it's Chainlit's callback handler for Langchain (as opposed to `LangchainCallbackHandler`, without `cl.` in front).

### DIFF
--- a/src/chainlit/__init__.py
+++ b/src/chainlit/__init__.py
@@ -32,7 +32,7 @@ from chainlit.sync import make_async, run_sync
 from chainlit.telemetry import trace
 from chainlit.types import LLMSettings
 from chainlit.user_session import user_session
-from chainlit.utils import wrap_user_function
+from chainlit.utils import make_module_getattr, wrap_user_function
 from chainlit.version import __version__
 
 env_found = load_dotenv(dotenv_path=os.path.join(os.getcwd(), ".env"))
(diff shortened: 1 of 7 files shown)


src/chainlit/__init__.py
```diff
@@ -181,6 +181,15 @@ def sleep(duration: int):
     return asyncio.sleep(duration)
 
 
+__getattr__ = make_module_getattr(
+    {
```
chenjuneking, line 185: Sorry I'm new in python, I wonder why not just import the handlers like this:

```py
from chainlit.langchain.callbacks import LangchainCallbackHandler, AsyncLangchainCallbackHandler, LlamaIndexCallbackHandler, HaystackAgentCallbackHandler
```

Does it has any advantages I didn't know?

willydouhard: This would be working fine but would also import the Langchain library by default, which we want to avoid for performance reasons. This PR will make the Langchain (and other libraries) imports lazy.

PR #279, opened 2023-08-15
fix: Update langflow load_flow_from_json method input parameter

nero408:
As langflow changed the method input parameter name it needs to be adapted to make it work again.

See following change in langflow here: https://github.com/logspace-ai/langflow/commit/008a1c40791310105c653e2e479afc9f1da6f412


ramnes (approved):
LGTM

willydouhard (approved):

PR #288, opened 2023-08-17
At/refactor fetch api

alimtunc:
Purpose of this PR :

1. Using SWR, a smart helper that makes grabbing data super smooth. This makes our GET calls faster and smarter. SWR knows when to get fresh data and when to use cached info.
2. Also, we're cleaning up our code by not storing our API stuff in the recoil store anymore.

I would appreciate to have a double check the API calls to be sure nothing is broken 🙌


src/chainlit/frontend/src/App.tsx
```diff
@@ -13,21 +13,28 @@ import SettingsModal from 'components/molecules/settingsModal';
...
 function App() {
-  const client = useRecoilValue(clientState);
   const { theme: themeVariant } = useRecoilValue(settingsState);
   const [accessToken, setAccessToken] = useRecoilState(accessTokenState);
-  const setRole = useSetRecoilState(roleState);
+  const [role, setRole] = useRecoilState(roleState);
   const { isAuthenticated, getAccessTokenSilently, logout } = useAuth();
   const theme = makeTheme(themeVariant);
 
+  const { data: roleData, error: roleError } = useApi(
```
willydouhard, line 30: the previous version was waiting to receive the access token to fetch the role. Will this version execute before having the access token and a second time after receiving it?

src/chainlit/frontend/src/AppWrapper.tsx
```diff
@@ -1,30 +1,29 @@
...
 
-        setAppSettings((prev) => ({
-          ...prev,
-          expandAll: !!res.ui.default_expand_messages,
-          hideCot: !!res.ui.hide_cot
-        }));
-      });
-    }
-  }, []);
+  if (pSettings === undefined && data) {
```
willydouhard, line 19: why double check if psettings is undefined?

src/chainlit/frontend/src/api/api.tsx
```diff
@@ -0,0 +1,48 @@
...
+    token?: string,
+    data?: any
+  ) => {
+    try {
+      const headers: { Authorization?: string; 'Content-Type': string } = {
+        'Content-Type': 'application/json'
+      };
+      if (token) headers['Authorization'] = `Bearer ${token}`;
+
+      const res = await fetch(httpEndpoint + endpoint, {
```
willydouhard, line 21: let's wait to merge the prompt playground PR first. This will conflict

willydouhard (commented):
Let's merge this after we cut the next release

src/chainlit/frontend/src/components/organisms/playground/submitButton.tsx
```diff
@@ -42,29 +40,31 @@ export default function SubmitButton() {
...
+              }
+            };
+          });
+        }
+      );
     } catch (err) {
       setCompletionController(undefined);
-
-      if (err instanceof ClientError) {
-        toast.error(err.toString());
```
willydouhard, line 66: If a completion request to an LLM provider (in the prompt playground) fails, we will not display the error to the user?

alimtunc: I've added a toast on error in api file, so it should be displayed

src/chainlit/frontend/src/api/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+        headers,
+        signal,
+        body: JSON.stringify(data)
+      });
+
+      if (!res.ok) {
+        const body = await res.json();
+
+        toast.error(res.statusText);
+        throw new Error(
```
willydouhard, line 35: the thrown error will be caught by the catch statement that is just beneath. Then it will only print the status text and not the full error with the details right?

That's why the previous error handling was looking like this:

```tsx
 if (!res.ok) {
      let err: ClientError;
      try {
        const body = await res.json();
        err = new ClientError(res.statusText, body.detail);
      } catch (_) {
        err = new ClientError(res.statusText);
      }
      throw err;
}
```

willydouhard (changes requested):

src/chainlit/frontend/src/api/index.tsx
alimtunc: The previous error handling was only for `const body = await res.json();` because there was no try/catch at all. Now, it's already wrapper by a try/catch, so if `res.json()` isn't it normal to have it catch by the global one ?

src/chainlit/frontend/src/components/organisms/dataset/deleteConversationButton.tsx
```diff
@@ -36,17 +36,20 @@ export default function DeleteConversationButton({
...
-          return <span>{err.message}</span>;
-        } else {
-          return <span></span>;
+    await toast.promise(
+      ChainlitAPI.deleteConversation(conversationId, accessToken),
+      {
+        loading: 'Deleting conversation...',
+        success: 'Conversation deleted!',
+        error: (err) => {
+          if (err instanceof Error) {
```
willydouhard, line 45: We can use ClientError here instead of Error

src/chainlit/frontend/src/api/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+        body: JSON.stringify(data)
+      });
+
+      if (!res.ok) {
+        const body = await res.json();
+        throw new ClientError(res.statusText, body.detail);
+      }
+
+      return res;
+    } catch (error: any) {
```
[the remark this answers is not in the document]
alimtunc: I don't think so, since we can't be sure what will be catched and Typescript can't determine it too. (cf: https://stackoverflow.com/a/69021180)

willydouhard (approved):

PR #289, opened 2023-08-17
Make the context directly accessible from chainlit.context.context

ramnes:


willydouhard (approved):
LGTM, do you want to merge this separately or directly with #290 ?

ramnes:
Let's merge

PR #292, opened 2023-08-18
add support for json objects in the Message component

### LINKED ISSUE #291: Chain of Thought UI crashes when langchain steps output dicts instead of strings
With `[Async]LangchainCallbackHandler`, when my intermediate steps output dicts instead of strings, the UI crashes.

Repro:
- Use the following code.
- Input a message.
- Click on "Took 1 step".
- Get "Something went wrong." with JS console saying "TypeError: r.trim is not a function".

```python
import chainlit as cl
from langchain.chains import TransformChain

@cl.on_message
async def main(message: str):
    def transform(inputs):
        return {
            'output': {
                'structured': {
                    'report': 'Done'
                }
            }
        }
    chain = TransformChain(
        input_variables=['text'],
        output_variables=['output'],
        transform=transform,
    )
    result = chain.run({'text': message}, callbacks=[cl.LangchainCallbackHandler()])
    return await cl.Message(content=result['structured']['report']).send()

```

(Possibly related [typing bug](https://github.com/langchain-ai/langchain/issues/9440) in LangChain 0.0.267, but I'm almost certain their intention is to allow intermediate steps to output whatever.)

willydouhard:
proposal to fix #291 
Screenshot 2023-08-18 at 17 16 42

### DIFF
--- a/src/chainlit/frontend/src/components/organisms/chat/message/content.tsx
+++ b/src/chainlit/frontend/src/components/organisms/chat/message/content.tsx
@@ -49,10 +49,6 @@ function prepareContent({ id, elements, content, language }: Props) {
     ? new RegExp(`(${elementNames.join('|')})`, 'g')
     : undefined;
 
-  if (typeof content === 'object') {
-    content = JSON.stringify(content, null, 4);
-    language = 'json';
-  }
   let preparedContent = content ? content.trim() : '';
   const inlinedElements = elements.filter(
     (e) => isForIdMatch(id, e?.forIds) && e.display === 'inline'
--- a/src/chainlit/frontend/src/components/organisms/chat/message/content.tsx
+++ b/src/chainlit/frontend/src/components/organisms/chat/message/content.tsx
@@ -49,6 +49,10 @@ function prepareContent({ id, elements, content, language }: Props) {
     ? new RegExp(`(${elementNames.join('|')})`, 'g')
     : undefined;
 
+  if (typeof content === 'object') {
+    content = JSON.stringify(content, null, 4);
+    language = 'json';
+  }
   let preparedContent = content ? content.trim() : '';
   const inlinedElements = elements.filter(
     (e) => isForIdMatch(id, e?.forIds) && e.display === 'inline'
--- a/src/chainlit/message.py
+++ b/src/chainlit/message.py
@@ -162,7 +162,7 @@ class Message(MessageBase):
         elif isinstance(content, str):
             self.content = content
         else:
-            raise ValueError(f"Unsupported type {type(content)} for message content")
+            raise TypeError(f"Unsupported type {type(content)} for message content")
 
         self.author = author
         self.prompt = prompt
--- a/src/chainlit/message.py
+++ b/src/chainlit/message.py
@@ -159,7 +159,7 @@ class Message(MessageBase):
         if isinstance(content, dict):
             self.content = json.dumps(content, indent=4)
             self.language = "json"
-        elif type(content) is str:
+        elif isinstance(content, str):
             self.content = content
         else:
             raise ValueError(f"Unsupported type {type(content)} for message content")
--- a/src/chainlit/message.py
+++ b/src/chainlit/message.py
@@ -156,7 +156,7 @@ class Message(MessageBase):
     ):
         self.language = language
 
-        if type(content) is dict:
+        if isinstance(content, dict):
             self.content = json.dumps(content, indent=4)
             self.language = "json"
         elif type(content) is str:
--- a/src/chainlit/message.py
+++ b/src/chainlit/message.py
@@ -1,3 +1,4 @@
+import json
 import uuid
 from abc import ABC, abstractmethod
 from datetime import datetime, timezone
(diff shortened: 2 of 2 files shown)


src/chainlit/frontend/src/components/organisms/chat/message/content.tsx
```diff
@@ -49,6 +49,10 @@ function prepareContent({ id, elements, content, language }: Props) {
     ? new RegExp(`(${elementNames.join('|')})`, 'g')
     : undefined;
 
+  if (typeof content === 'object') {
```
mpartel, line 52: Might technically be a number or boolean too. Perhaps `typeof content !== 'string'` ?

willydouhard: just moved the logic to the server. it has to be a string because the message can be persisted and has to respect the DB schema

src/chainlit/message.py
```diff
@@ -153,10 +154,18 @@ def __init__(
...
+            self.content = json.dumps(content, indent=4)
+            self.language = "json"
+        elif type(content) is str:
+            self.content = content
+        else:
+            raise ValueError(f"Unsupported type {type(content)} for message content")
+
         self.author = author
         self.prompt = prompt
-        self.language = language
```
ramnes, line 159: ?

src/chainlit/message.py
```diff
@@ -153,10 +154,18 @@ def __init__(
         actions: Optional[List[Action]] = None,
         elements: Optional[List[ElementBased]] = None,
     ):
-        self.content = content
+        self.language = language
+
+        if type(content) is dict:
```
ramnes, line 159: let's allow subclasses as well

```suggestion
        if isinstance(content, dict):
```

src/chainlit/message.py
```diff
@@ -153,10 +154,18 @@ def __init__(
...
+            self.content = json.dumps(content, indent=4)
+            self.language = "json"
+        elif type(content) is str:
```
ramnes, line 162: let's allow subclasses as well

```suggestion
        elif isinstance(content, str):
```

src/chainlit/message.py
```diff
@@ -153,10 +154,18 @@ def __init__(
...
+            self.content = content
+        else:
+            raise ValueError(f"Unsupported type {type(content)} for message content")
```
ramnes, line 165:
```suggestion
            raise TypeError(f"Unsupported type {type(content)} for message content")
```

ramnes (approved):
LGTM

mpartel:
It gets a little fiddly.
If I add the following to #291 's code:
```python
from langchain.schema.document import Document
...
                'structured': {
                    'doc': Document(page_content='...'),
                    ...
```
then I get
```
future: <Task finished name='Task-83' coro=<AsyncServer._emit_internal() done, defined at /home/martin/.cache/pypoetry/virtualenvs/chainlit-bug-6EGjyFTw-py3.10/lib/python3.10/site-packages/socketio/asyncio_server.py:427> exception=TypeError('Object of type Document is not JSON serializable')>
  File "/usr/lib/python3.10/asyncio/tasks.py", line 232, in __step
    result = coro.send(None)
  File "/home/martin/.cache/pypoetry/virtualenvs/chainlit-bug-6EGjyFTw-py3.10/lib/python3.10/site-packages/socketio/asyncio_server.py", line 437, in _emit_internal
    await self._send_packet(sid, self.packet_class(
  File "/home/martin/.cache/pypoetry/virtualenvs/chainlit-bug-6EGjyFTw-py3.10/lib/python3.10/site-packages/socketio/asyncio_server.py", line 442, in _send_packet
    encoded_packet = pkt.encode()
  File "/home/martin/.cache/pypoetry/virtualenvs/chainlit-bug-6EGjyFTw-py3.10/lib/python3.10/site-packages/socketio/packet.py", line 64, in encode
    encoded_packet += self.json.dumps(data, separators=(',', ':'))
  File "/usr/lib/python3.10/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/lib/python3.10/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.10/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python3.10/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type Document is not JSON serializable
```

Looks like we'd need to give [`socketio.AsyncServer`](https://python-socketio.readthedocs.io/en/latest/api.html?highlight=json#socketio.AsyncServer) a `json` module (object?) that uses a [`JSONEncoder`](https://docs.python.org/3/library/json.html#json.JSONEncoder) that supports LangChain's [`Serializable`](https://api.python.langchain.com/en/latest/load/langchain.load.serializable.Serializable.html?highlight=serializable#langchain.load.serializable.Serializable).

ramnes:
This is another topic IMO. For this I'd say that we should let users pass whatever JSONEncoder they want to Chainlit rather than trying to handle all cases ourselves.

willydouhard:
I could even arg that a Langchain document should be a text element rather than a message content.

Should we keep the support for dicts in the message class and open another PR for Langchain edge cases or do you think everything should be handled in the Langchain specific solution and the message class should always accept a string?

PR #294, opened 2023-08-18
refactor: Collapse long message

alimtunc:
https://github.com/Chainlit/chainlit/assets/18298759/da55140c-73ab-4f89-b6ef-e6c55e28f87d

### DIFF
--- a/src/chainlit/frontend/src/components/atoms/Collapse.tsx
+++ b/src/chainlit/frontend/src/components/atoms/Collapse.tsx
@@ -1,11 +1,7 @@
 import React from 'react';
 import { useToggle } from 'usehooks-ts';
 
-import {
-  ArrowCircleDown,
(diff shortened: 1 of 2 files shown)


chenjuneking:
It's really a nice work!
I think if we could make the hidden text ellipsis, that would be better.
Like this:

image

image

What do you think?

alimtunc:
> It's really a nice work! I think if we could make the hidden text ellipsis, that would be better. Like this:
> 
> image
> 
> image
> 
> What do you think?

Thank you!

That would be a great idea if the message was only text, since we can have element, text on the same div, would be tricky to have a good ellipsis

tpatel (approved):
✨✨✨ AWESOME ✨✨✨

Are we fine doing the limit on the number of characters rather than the number of lines?

PR #295, opened 2023-08-18
Pydantic v2 support

ramnes:
What's missing:
- [x] Prisma Client Python support for Pydantic v2: https://github.com/RobertCraigie/prisma-client-py/issues/791
- [ ] Running the tests of integrations that do not support Pydantic v2 (Haystack, LlamaIndex) with Pydantic v1


willydouhard:
Could chainlit also support both (like langchain seems to do) so users could use any third party, regardless of their version of pydantic?

ramnes:
yes that's the plan

MathiasSpanhove:
Prisma [v0.10.0](https://github.com/RobertCraigie/prisma-client-py/releases/tag/v0.10.0) has just been released with Pydantic v2 support

PR #304, opened 2023-08-21
✨ feat(chainlit): add support for spontaneous file uploads

alimtunc:
Purpose of this PR is to add support for spontaneous file  :

https://github.com/Chainlit/chainlit/assets/18298759/1da9bf8a-0206-4dac-bd74-1fb03600fba9

App to test

```python
import chainlit as cl
from langchain.agents import  AgentExecutor
import os
import chainlit as cl

from chainlit.langchain.callbacks import LangchainCallbackHandler


@cl.on_file_upload(accept=["text/plain"], max_files=3)
async def upload_file(files: any):
    for file_data in files:
        file_name = file_data['name']
        print(f"Uploaded file: {file_name}\nContent: {file_data['content']}\n\n")

@cl.on_message
async def main(message):
    agent = cl.user_session.get("agent")  # type: AgentExecutor
    cb = LangchainCallbackHandler(stream_final_answer=True)

    await cl.make_async(agent.run)(message, callbacks=[cb])

```


src/chainlit/__init__.py
```diff
@@ -176,6 +176,39 @@ def db_client_factory(
     return func
 
 
+def on_file_upload(
+    accept: Union[List[str], Dict[str, List[str]]],
+    type: Optional[str] = "file",
```
tpatel, line 181: I think we shouldn't offer the option for users to change the type. I believe we should hardcode it in `config.code.on_file_upload_config`.

src/chainlit/frontend/src/components/organisms/chat/message/uploadButton.tsx
```diff
@@ -126,12 +143,12 @@ function _UploadButton({ askUser }: Props) {
   );
 }
 
-export default function UploadButton() {
+export default function UploadButton({ onlyButton }: { onlyButton?: boolean }) {
   const askUser = useRecoilValue(askUserState);
```
tpatel, line 147: Because we are reusing the state from the ask messages, this hides the spontaneous file upload button when we are waiting for an `AskUserMessage`. Example code:

```python
import chainlit as cl


@cl.on_file_upload(accept=["text/plain"], max_files=3)
async def upload_file(files: any):
    for file_data in files:
        file_name = file_data["name"]
        print(f"Uploaded file: {file_name}\nContent: {file_data['content']}\n\n")


@cl.on_chat_start
async def start():
    await cl.AskUserMessage(content="What is your name?", timeout=100).send()
```

My suggestion would be to have two different upload buttons instead.

tpatel (changes requested):
Really cool to see this feature close to be shipped! I've got a couple of comments

src/chainlit/frontend/src/components/organisms/chat/message/uploadButton.tsx
alimtunc: You're right, I've separated them now and added a `useUpload` hook, let me know what you think about this new version 🫡

willydouhard:
Seems to have linting warnings in the latest commit

alimtunc:
> Seems to have linting warnings in the latest commit

Yes, should be good now

src/chainlit/socket.py
```diff
@@ -236,3 +242,12 @@ async def change_settings(sid, settings: Dict[str, Any]):
 
     if config.code.on_settings_update:
         await config.code.on_settings_update(settings)
+
+
+@socket.on("file_upload")
+async def file_upload(sid, files: Any):
+    """Handle change settings submit from the UI."""
```
tpatel, line 249: This doc string from a previous commit isn't up to date

tpatel (approved):
✨✨✨ AWESOME ✨✨✨

One small edge case: a user could use the `@cl.on_file_upload` annotation twice and only the latest config would be saved. We could display a warning when the config already exists? Or ignore because it's too small to consider.

src/chainlit/__init__.py
```diff
@@ -176,6 +176,38 @@ def db_client_factory(
...
+        type (Optional[str]): The type of upload, defaults to "file".
+        max_size_mb (Optional[int]): The maximum file size in megabytes, defaults to 2.
+        max_files (Optional[int]): The maximum number of files allowed to be uploaded, defaults to 1.
+
+    Returns:
+        Callable: The decorated function for handling spontaneous file uploads.
+    """
+
+    def decorator(func: Callable) -> Callable:
+        config.code.on_file_upload_config = {
```
willydouhard, line 199: Why not use a FileSpec here instead of a dict?

src/chainlit/frontend/src/components/organisms/chat/inputBox/input.tsx
```diff
@@ -41,7 +42,7 @@ const Input = ({ onSubmit, onReply }: Props) => {
   const [isComposing, setIsComposing] = useState(false);
 
   const socketOk = session?.socket && !session?.error;
-  const disabled = !socketOk || loading || askUser?.spec.type === 'file';
+  const disabled = !socketOk || loading;
```
willydouhard, line 45: why remove the `askUser?.spec.type === 'file'` condition? If the server is waiting for a file the input box should be disabled

willydouhard (changes requested):

willydouhard (approved):

PR #308, opened 2023-08-22
Add typehint on lazy proxy

xsyann:


src/chainlit/context.py
```diff
@@ -47,4 +47,4 @@ def get_context() -> ChainlitContext:
         raise ChainlitContextException()
 
 
-context = LazyProxy(get_context, enable_cache=False)
+context: ChainlitContext = LazyProxy(get_context, enable_cache=False)
```
ramnes, line 50: Should work I guess, but ideally we'd want something like `LazyProxy[ChainlitContext]`. Happy to take a PR on lazify to support this!

ramnes (approved):

willydouhard (approved):

PR #312, opened 2023-08-24
Automated prompt playground langchain llm provider

tpatel:
- Adding it automatically avoid the need for configuration
- Because the provider list is now dynamic, we need to update the llm provider list in the frontend
- There are no inputs in the automated providers yet

### DIFF
--- a/cypress/e2e/prompt_playground/provider.py
+++ b/cypress/e2e/prompt_playground/provider.py
@@ -1,12 +1,15 @@
 import os
 
 from fastapi.responses import StreamingResponse
+from langchain.llms.fake import FakeListLLM
 
 import chainlit as cl
 from chainlit.input_widget import Select, Slider
 from chainlit.playground.config import BaseProvider, add_llm_provider
+from chainlit.playground.providers.langchain import LangchainGenericProvider
 
 os.environ["TEST_LLM_API_KEY"] = "sk..."
+
 
 class TestLLMProvider(BaseProvider):
     async def create_completion(self, request):
@@ -71,5 +74,26 @@ ChatTestLLM = TestLLMProvider(
     is_chat=True,
 )
 
+responses = ["This is the test completion"]
+llm = FakeListLLM(responses=responses)
+
+LangchainTestLLM = LangchainGenericProvider(
+    id="test-langchain",
+    name="TestLangchain",
+    llm=llm,
+    inputs=[
+        Select(
+            id="model",
+            label="Model",
+            values=["test-model-langchain-1", "test-model-langchain-2"],
+            initial_value="test-model-langchain-2",
+        ),
+    ],
+    is_chat=False,
+    is_automated=False,
+)
+
+
 add_llm_provider(TestLLM)
 add_llm_provider(ChatTestLLM)
+add_llm_provider(LangchainTestLLM)
--- a/src/chainlit/frontend/src/components/organisms/playground/index.tsx
+++ b/src/chainlit/frontend/src/components/organisms/playground/index.tsx
@@ -35,7 +35,7 @@ import SubmitButton from './submitButton';
 export default function PromptPlayground() {
   const [playground, setPlayground] = useRecoilState(playgroundState);
 
-  const { data, error } = useApi<IPlayground>('/project/llm-providers');
+  const { data, error, mutate } = useApi<IPlayground>('/project/llm-providers');
 
   const [restoredTime, setRestoredTime] = useState(0);
   const [isDrawerOpen, toggleDrawer] = useToggle(false);
@@ -45,6 +45,12 @@ export default function PromptPlayground() {
   const isSmallScreen = useMediaQuery<Theme>((theme) =>
     theme.breakpoints.down('md')
   );
+
+  useEffect(() => {
+    // Refresh the providers when the playground is opened
+    if (!playground?.prompt) return;
+    mutate();
+  }, [playground?.prompt]);
 
   useEffect(() => {
     if (!data) return;
--- a/src/chainlit/frontend/src/components/organisms/playground/index.tsx
+++ b/src/chainlit/frontend/src/components/organisms/playground/index.tsx
@@ -65,6 +65,8 @@ export default function PromptPlayground() {
     setPlayground((old) => ({ ...old, prompt: undefined }));
   };
 
+  // Only render the playground if it's open and we have providers.
+  // Prevents the "no provider" error from being thrown and a playground flash
   if (!playground?.prompt) {
     return null;
   }
--- a/src/chainlit/langchain/callbacks.py
+++ b/src/chainlit/langchain/callbacks.py
@@ -1,6 +1,5 @@
-import json
 import importlib
-import re
+import json
 from typing import Any, Dict, List, Optional, Union
 
 from langchain.callbacks.base import AsyncCallbackHandler, BaseCallbackHandler
--- a/src/chainlit/langchain/callbacks.py
+++ b/src/chainlit/langchain/callbacks.py
@@ -1,12 +1,17 @@
 import json
+import importlib
+import re
 from typing import Any, Dict, List, Optional, Union
 
 from langchain.callbacks.base import AsyncCallbackHandler, BaseCallbackHandler
 from langchain.schema import AgentAction, AgentFinish, BaseMessage, LLMResult
 
+from chainlit import input_widget
 from chainlit.config import config
 from chainlit.context import context
 from chainlit.message import ErrorMessage, Message
+from chainlit.playground.config import add_llm_provider, has_llm_provider
+from chainlit.playground.providers.langchain import LangchainGenericProvider
 from chainlit.prompt import Prompt, PromptMessage
 from chainlit.sync import run_sync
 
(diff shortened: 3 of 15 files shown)


src/chainlit/frontend/src/components/organisms/playground/index.tsx
```diff
@@ -37,20 +37,19 @@ export default function PromptPlayground() {
...
-  if (isFirstRender) {
+  useEffect(() => {
     client
       .getLLMProviders()
       .then((res) =>
         setPlayground((old) => ({ ...old, providers: res.providers }))
       )
       .catch((err) => setProvidersError(err));
-  }
+  }, [playground?.prompt?.provider]);
```
willydouhard, line 52: why is this required? everytime we open the prompt playground the firstRender is triggered no?

src/chainlit/langchain/callbacks.py
```diff
@@ -286,6 +293,9 @@ def _on_llm_start(
     invocation_params = kwargs.get("invocation_params")
     provider, settings = get_llm_settings(invocation_params)
 
+    # TODO: path not tested yet
+    provider = ".".join(serialized.get("id", []))
```
willydouhard, line 297: provider is defined just above with get_llm_settings

src/chainlit/langchain/callbacks.py
```diff
@@ -375,6 +385,24 @@ def on_llm_end(self, response: LLMResult, **kwargs: Any) -> None:
     def on_chain_start(
         self, serialized: Dict[str, Any], inputs: Dict[str, Any], **kwargs: Any
     ) -> None:
+        path = serialized.get("kwargs", {}).get("llm", {}).get("id", [])
```
willydouhard, line 388: what is path exactly?

src/chainlit/langchain/callbacks.py
```diff
@@ -500,6 +528,24 @@ async def on_llm_end(self, response: LLMResult, **kwargs: Any) -> None:
...
+        provider = ".".join(path)
+        provider_name = path[-1]
+        provider_path = ".".join(path[:-1])
+
+        if not any(provider == p.id for p in providers):
+            module = importlib.import_module(provider_path)
+            llm_class = getattr(module, provider_name)
```
willydouhard, line 538: This is cool that we can automate this but it has limits:

1. We assume the env variables are available globally (where the llm could have been instantiated with the API key)
2. The inputs are not available

We could keep this approach as a last resort but also create a documentation on "How to Add a Langchain LLM provider" using the same LangchainGenericProvider class

src/chainlit/frontend/src/components/organisms/playground/index.tsx
tpatel: The first render happens before the modal is open, and the component returns `null` (see https://github.com/Chainlit/chainlit/pull/312/files#diff-96df05b06bc651bccb782e3ce335080f28b7cea2308758d45896652e3c2d1ccaR68-R70), this is happening when the page is displayed.

The provider is created by the langchain callback. This means the langchain chain must have started before we have the right providers available (instead of the default ones).
This isn't the perfect way to refresh the llm providers as it creates a couple extra requests, but it does solve our issue with minimal overhead.

src/chainlit/langchain/callbacks.py
tpatel: yes, I had to redefine the provider for the purpose of automation. I still need to clean up the `get_llm_settings` function.

Right now the `get_llm_settings` returns something like `openai-chat`, while we are now using the full langchain class path as the provider id (extracted from the `serialized` dict)
This enables us to access the llm id from places where the `invocation_params` aren't configured.

src/chainlit/langchain/callbacks.py
tpatel: it is the full path of the class in the langchain module, it is used as an ID now (more on that in https://github.com/Chainlit/chainlit/pull/312/files#r1304264859)

src/chainlit/langchain/callbacks.py
tpatel: I think we should have the most automated approach by default. So that it's super smooth to integrate.
And have the "How to Add a Langchain LLM provider" option for more customized use cases, wdyt?

src/chainlit/playground/providers/langchain.py
```diff
@@ -0,0 +1,79 @@
...
+            result = await self.llm._agenerate(
+                messages=[
+                    self.prompt_message_to_langchain_message(i)
+                    for i in request.prompt.messages
+                ],
+                **request.prompt.settings,
+            )
+
+            return PlainTextResponse(
+                result.generations[0].text, media_type="text/plain"
```
willydouhard, line 68: we should support streaming if the llm supports streaming. Also is it working to send a PlainTextResponse if the frontend expects an event stream?

src/chainlit/playground/providers/langchain.py
```diff
@@ -52,8 +52,14 @@ def message_to_string(self, message: PromptMessage) -> str:
...
         await super().create_completion(request)
 
-        # Update the request.prompt.messages with the formatted messages
-        self.create_prompt(request)
+        # Edge case where openai doesn't like empty stop sequences
+        if (
+            self.id in ["azure-openai-chat", "openai", "azure", "openai-chat"]
+            and "stop" in request.prompt.settings
+            and request.prompt.settings["stop"] is not None
+            and len(request.prompt.settings["stop"]) == 0
```
willydouhard, line 60: of stop is an empty array then it works? Also these providers are supposed to be available by default in chainlit

src/chainlit/langchain/callbacks.py
```diff
@@ -16,17 +16,43 @@
...
     if provider.startswith("openai"):
         model_name = invocation_params.pop("model_name")
         invocation_params["model"] = model_name
 
-    return provider, invocation_params
+    merged = {**invocation_params, **serialized.get("kwargs", {})}
+
+    # make sure there is no api key specification nor _type
+    filtered = {
+        k: v for k, v in merged.items() if not k.endswith("_api_key") and k != "_type"
```
willydouhard, line 52: i think we pop _type above so maybe not necessary

src/chainlit/playground/providers/langchain.py
```diff
@@ -0,0 +1,88 @@
...
+        self.llm = llm
+
+    def prompt_message_to_langchain_message(self, message: PromptMessage):
+        content = "" if message.formatted is None else message.formatted
+        if message.role == "user":
+            return HumanMessage(content=content)
+        elif message.role == "assistant":
+            return AIMessage(content=content)
+        elif message.role == "system":
+            return SystemMessage(content=content)
```
willydouhard, line 49: I think we should also handle FunctionMessage. Here is a small snippet to test it:

```py
from langchain import LLMMathChain
from langchain.agents import initialize_agent, Tool
from langchain.agents import AgentType
from langchain.chat_models import ChatOpenAI


llm = ChatOpenAI(temperature=0, model="gpt-3.5-turbo-0613")
llm_math_chain = LLMMathChain.from_llm(llm=llm, verbose=True)
tools = [
    Tool(
        name="Calculator",
        func=llm_math_chain.run,
        description="useful for when you need to answer questions about math"
    )
]

mrkl = initialize_agent(tools, llm, agent=AgentType.OPENAI_FUNCTIONS, verbose=True)
```

src/chainlit/langchain/callbacks.py
```diff
@@ -23,7 +27,51 @@ def get_llm_settings(invocation_params: Union[Dict, None]):
...
+
+    # Automatically add the provider if it doesn't exist yet
+    if not has_llm_provider(provider):
+        path = serialized.get("id", [])
+        provider_name = path[-1]
+        provider_path = ".".join(path[:-1])
+
+        module = importlib.import_module(provider_path)
+        llm_class = getattr(module, provider_name)
+        llm = llm_class()
```
willydouhard, line 45: i think we should instantiate the llm with the current settings, including the API key. If the API key was passed to the llm and is not available in the env variables this won't work

cypress/e2e/prompt_playground/provider.py
```diff
@@ -71,5 +73,24 @@ async def create_event_stream():
...
+            id="model",
+            label="Model",
+            values=["test-model-langchain-1", "test-model-langchain-2"],
+            initial_value="test-model-langchain-2",
+        ),
+    ],
+    is_chat=False,
+    is_automated=False,
+)
+LangchainTestLLM.create_completion = ChatTestLLM.create_completion
```
willydouhard, line 92: Why are we not passing the llm class directly?

src/chainlit/langchain/callbacks.py
tpatel: we've moved back to the langchain id, like `openai-chat` instead of the full class path

src/chainlit/playground/providers/langchain.py
tpatel: we do support streaming now 👌

src/chainlit/playground/providers/langchain.py
tpatel: good point, I've removed the edge case

src/chainlit/langchain/callbacks.py
tpatel: The api key isn't available there, so we will keep instantiating the LLM without parameters for now. User can still declare their own (non automated) provider to go around this issue.

cypress/e2e/prompt_playground/provider.py
tpatel: right the langchain `FakeListLLM` seems like a good choice there 👌

src/chainlit/playground/providers/langchain.py
tpatel: done

willydouhard (approved):

PR #322, opened 2023-08-25
Avoid deadlock in event start

xsyann:
When using streaming, the main event loop is blocked to await tokens, if we try to execute a function (`Message.send()` here) in the same event loop, it leads to a deadlock.

Adding a `new_loop` parameter to the `run_sync` function allows to create a new event loop in the current Thread and allows to send a Message during a stream.

### DIFF
--- a/src/chainlit/sync.py
+++ b/src/chainlit/sync.py
@@ -21,9 +21,19 @@ T_ParamSpec = ParamSpec("T_ParamSpec")
 T = TypeVar("T")
 
 
-def run_sync(co: Coroutine[Any, Any, T_Retval]) -> T_Retval:
+def run_sync(co: Coroutine[Any, Any, T_Retval], force_new_loop=False) -> T_Retval:
+    """Run the coroutine synchronously. If force_new_loop is True,
+    the coroutine is executed in a new event loop, otherwise it is
+    executed in the main event loop."""
+
+    # Execute in a new event loop
+    if force_new_loop:
+        loop = asyncio.new_event_loop()
+        return loop.run_until_complete(co)
+
+    # Execute from the main thread in the main event loop
     if threading.current_thread() == threading.main_thread():
         return sync(co)
-    else:
+    else:  # Execute from a thread in the main event loop
         result = asyncio.run_coroutine_threadsafe(co, loop=context.loop)
         return result.result()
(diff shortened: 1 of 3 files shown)


ramnes (commented):
Good if that fixes an issue.

It looks like a very weird approach to me though, because:
1. It doesn't really make sense to run an event loop for just one function;
2. I don't understand, as a `run_sync()` user, how do I know if I should pass `new_loop=True` or `new_loop=False`? Am I supposed to try one, and if that doesn't work, try the other one? My point being that the event loop handling probably shouldn't be the responsibility of the API user.

xsyann:
> Good if that fixes an issue.
> 
> It looks like a very weird approach to me though, because:
> 
> 1. It doesn't really make sense to run an event loop for just one function;
> 2. I don't understand, as a `run_sync()` user, how do I know if I should pass `new_loop=True` or `new_loop=False`? Am I supposed to try one, and if that doesn't work, try the other one? My point being that the event loop handling probably shouldn't be the responsibility of the API user.

1. We don't really have the choice as we need to execute the fonction in the Thread (as the main event loop is blocked)
I agree that a better solution would have been to make the token generator async to avoid blocking the main event loop but it is not yet implemented in llama_index.

2. I changed it to a `main_thread` boolean and added some doc. As it has a default value, generally a user shouldn't have to deal with it, except in this particular use case.

src/chainlit/sync.py
```diff
@@ -21,9 +21,15 @@
...
+def run_sync(co: Coroutine[Any, Any, T_Retval], main_thread=True) -> T_Retval:
+    """Run the coroutine synchronously. If main_thread is True, the coroutine
+    is executed in the main thread, otherwise it is executed in the current
+    thread."""
+
     if threading.current_thread() == threading.main_thread():
         return sync(co)
     else:
+        if not main_thread:
+            return asyncio.run(co)
```
willydouhard, line 33: why do we use asyncio.run here instead of sync()? Also the main_thread parameter naming might be confusing since we check `threading.current_thread() == threading.main_thread()` in the condition above

willydouhard (approved):

ramnes (approved):
I like `force_new_loop` better 👍

PR #330, opened 2023-08-28
refactor: Replace husky & lint-staged by lefthook

alimtunc:
Running pre-commit fast, concurrently and only on staged files! (Even for mypy now 🔥)

Capture d’écran 2023-08-28 à 10 58 15

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,0 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
-pnpm lint-staged
-pnpm run lintPython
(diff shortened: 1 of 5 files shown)


willydouhard (approved):

alimtunc:
> commands that modify stuff running in parallel is probably a bad idea

Depends on how they synchronize the updated files, I did not see issue like that. But yeah, need to take a long in a long term and modified base. 

But normally, you shouldn't have stuff being modified by lefthook since your dev workspace should be setup and lint/format on each files update 🫣

willydouhard:
I kinda agree with @ramnes. Some developers will lint on save some rely on the pre commit hook.
The concurrency point is also valid, if they the different steps are ran concurrently what happens if they modify the same file at the "same" time?

alimtunc:
Ok, I ended up with something like this. Let me know @ramnes @willydouhard

PR #332, opened 2023-08-28
Compare versions with packaging.version

xsyann:
Fix version comparison from string compare to PEP 440 versions.
`packaging.version` is used instead of `semver` because `semver` doesn't support version like `0.8.11.post3`

### DIFF
--- a/src/chainlit/utils.py
+++ b/src/chainlit/utils.py
@@ -2,6 +2,8 @@ import importlib
 import inspect
 from typing import Callable
 
+from packaging import version
+
 from chainlit.context import context
 from chainlit.logger import logger
 from chainlit.message import ErrorMessage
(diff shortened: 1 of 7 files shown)


willydouhard (approved):

src/chainlit/utils.py
```diff
@@ -63,3 +65,22 @@ def __getattr__(name):
...
+        version (str): Minimum version.
+
+    Returns:
+        (bool): Return False if the module is installed and the version
+            is lower than the minimum version required.
+    """
+    try:
+        module = importlib.import_module(name)
+    except ModuleNotFoundError:
+        return True
```
tpatel, line 85: Shouldn't it be `False` here?

PR #309, opened 2023-08-22
feat: move frontend out of poetry scope

chenjuneking:
After a new release publish, I think it's a good time to move frontend out of the poetry scope to speed up the build time.

Close #276

PR #343, opened 2023-08-31
refactor: Distinct frontend from backend

alimtunc:
Following this https://github.com/Chainlit/chainlit/pull/309


backend/chainlit/sync.py
```diff
@@ -10,9 +10,8 @@
 import threading
 
 from asyncer import asyncify
-from syncer import sync
-
 from chainlit.context import context
+from syncer import sync
```
ramnes, line 14: wrong

alimtunc: This was moved automatically, but I'm curious, why is that wrong?

willydouhard (approved):

PR #350, opened 2023-09-01
prevent langchain to be imported by default

willydouhard:


MathiasSpanhove (approved):
LGTM, thank you for the fast fix!

PR #352, opened 2023-09-04
feat(vertexai): Add GCP Vertex AI providers with generation and chat …

fvaleye:
# Description
- Add the Generation and Chat models from [GCP VertexAI](https://cloud.google.com/vertex-ai?hl=en)
- Using the `GOOGLE_APPLICATION_CREDENTIALS` environment variables, the playground is available for chat and generation LLM models of VertexAI
- Only common parameters are available: `temperature` and `max_output_tokens`

### DIFF
--- a/backend/chainlit/playground/providers/vertexai.py
+++ b/backend/chainlit/playground/providers/vertexai.py
@@ -1,0 +1,126 @@
+import chainlit as cl
+from fastapi import HTTPException
+
+from fastapi.responses import StreamingResponse
+
+from chainlit.input_widget import Select, Slider, Tags
+from chainlit.playground.provider import BaseProvider
+
+vertexai_common_inputs = [
+    Slider(
+        id="temperature",
+        label="Temperature",
+        min=0.0,
+        max=0.99,
+        step=0.01,
+        initial=0.2,
+    ),
+    Slider(
+        id="max_output_tokens",
+        label="Max Output Tokens",
+        min=0.0,
+        max=1024,
+        step=1,
+        initial=256,
+    ),
+]
+
+
+class ChatVertexAIProvider(BaseProvider):
+    async def create_completion(self, request):
+        await super().create_completion(request)
+        from vertexai.language_models import ChatModel, CodeChatModel
+
+        self.validate_env(request=request)
+
+        llm_settings = request.prompt.settings
+        self.require_settings(llm_settings)
+
+        prompt = self.create_prompt(request)
+        model_name = llm_settings["model"]
+        if model_name.startswith("chat-"):
+            model = ChatModel.from_pretrained(model_name)
+        elif model_name.startswith("codechat-"):
+            model = CodeChatModel.from_pretrained(model_name)
+        else:
+            raise HTTPException(
+                status_code=400,
+                detail=f"This model{model_name} is not implemented.",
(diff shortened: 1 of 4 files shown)


backend/chainlit/playground/providers/vertexai.py
```diff
@@ -0,0 +1,117 @@
...
+        self.require_settings(llm_settings)
+
+        prompt = self.create_prompt(request)
+        model_name = llm_settings["model"]
+        if model_name.startswith("chat-"):
+            model = ChatModel.from_pretrained(model_name)
+        elif model_name.startswith("codechat-"):
+            model = CodeChatModel.from_pretrained(model_name)
+        else:
+            raise ValueError(f"This model{model_name} is not implemented.")
```
willydouhard, line 45: I would raise an HTTP 400 error here. It will display the detailed message in the UI

backend/chainlit/playground/providers/vertexai.py
```diff
@@ -0,0 +1,117 @@
...
+        del llm_settings["model"]
+        chat = model.start_chat()
+
+        def create_event_stream():
+            for response in chat.send_message_streaming(
```
willydouhard, line 50: if this is sync implementation I would wrap this in cl.make_async. The reason is that it will block the event loop otherwise. If an async implementation is available, that's even better!

backend/chainlit/playground/providers/vertexai.py
```diff
@@ -0,0 +1,117 @@
...
+        self.require_settings(llm_settings)
+
+        prompt = self.create_prompt(request)
+        model_name = llm_settings["model"]
+        if model_name.startswith("text-"):
+            model = TextGenerationModel.from_pretrained(model_name)
+        elif model_name.startswith("code-"):
+            model = CodeGenerationModel.from_pretrained(model_name)
+        else:
+            raise ValueError(f"This model{model_name} is not implemented.")
```
willydouhard, line 75: same comment as above for the http error

backend/chainlit/playground/providers/vertexai.py
```diff
@@ -0,0 +1,117 @@
...
+        del llm_settings["model"]
+
+        def create_event_stream():
+            for response in model.predict_streaming(prompt, **llm_settings):
```
willydouhard, line 79: same comment as above for async / make_async

willydouhard (changes requested):
This is awesome! Almost ready to merge, few comments to address first

willydouhard:
Screenshot 2023-09-04 at 12 19 44
My py caught an error import, I think it should import from `vertexai.py` and not `openai.py`

backend/chainlit/playground/providers/vertexai.py
fvaleye: Not available, unfortunately, I will include `make_async`.

backend/chainlit/playground/providers/vertexai.py
fvaleye: Same as well.

backend/chainlit/playground/providers/vertexai.py
fvaleye: Same as well.

PR #355, opened 2023-09-04
make favicon customizable

willydouhard:


backend/chainlit/server.py
```diff
@@ -112,8 +120,7 @@ async def watch_files_for_changes():
         os._exit(0)
 
 
-root_dir = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
-build_dir = os.path.join(root_dir, "frontend/dist")
+build_dir = os.path.join(PACKAGE_ROOT, "frontend/dist")
```
tpatel, line 123: Same feedback as my other comment
```suggestion
build_dir = os.path.join(PACKAGE_ROOT, "frontend", "dist")
```

backend/chainlit/db/__init__.py
```diff
@@ -1,9 +1,9 @@
 import os
 
-from chainlit.config import PACKAGE_ROOT, config
+from chainlit.config import BACKEND_ROOT, config
 from chainlit.logger import logger
 
-SCHEMA_PATH = os.path.join(PACKAGE_ROOT, "db/prisma/schema.prisma")
+SCHEMA_PATH = os.path.join(BACKEND_ROOT, "db/prisma/schema.prisma")
```
tpatel, line 6: Not sure if the slash could be a problem when running on Windows, here is a suggested change:
```suggestion
SCHEMA_PATH = os.path.join(BACKEND_ROOT, "db", "prisma", "schema.prisma")
```

backend/chainlit/server.py
willydouhard: Good catch!

clementsirieix (approved):
🚀 LGTM

(Didn't know about the `join` caveat, good to know 👌)

PR #358, opened 2023-09-04
Add an option to disable the prompt playground

clementsirieix:
### Changes 

- The playground display behavior can be handled through the chainlit config (TODO: Update documentation)
- The `prompt` data is filtered from response
- Updated `.gitignore` to allow e2e config tracking

### How to test?

- Run `chainlit run ../cypress/e2e/hide_prompt_playground/main.py` 
- Notice that we are not displaying the playground


cypress/e2e/hide_prompt_playground/provider.py
```diff
@@ -0,0 +1,52 @@
+import os
```
willydouhard, line 1: i think we don't need this file. Since we are never going to open the PP we don't need providers

backend/chainlit/config.py
```diff
@@ -72,6 +72,9 @@
 # Link to your github repo. This will add a github button in the UI's header.
 # github = ""
 
+# Show the prompt playground
+show_playground = true
```
willydouhard, line 76: maybe rename show_prompt_playground?

willydouhard (changes requested):

willydouhard (approved):

willydouhard (approved):

PR #361, opened 2023-09-05
Add feature flags in config

clementsirieix:
## Changes

- Create `features` config table
- Move existing `UI.show_prompt_playground` to `features.prompt_playground`

## How to test?

- run the `e2e/hide_prompt_playground` test

### DIFF
--- a/backend/chainlit/config.py
+++ b/backend/chainlit/config.py
@@ -58,6 +58,10 @@ cache = false
 # Chainlit server address
 # chainlit_server = ""
 
+[features]
+# Show the prompt playground
+prompt_playground = true
+
 [UI]
 # Name of the app and chatbot.
 name = "Chatbot"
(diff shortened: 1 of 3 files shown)


backend/chainlit/config.py
```diff
@@ -233,6 +237,7 @@ class ChainlitConfig:
     chainlit_prod_url = chainlit_prod_url
 
     run: RunSettings
+    features: FeaturesSettings
```
willydouhard, line 240: I would name it `FeatureSettings` for consistency

willydouhard (changes requested):

backend/chainlit/config.py
clementsirieix: Gotcha, I may be a bit nitpicky but I think that the straightforward convention is a PascalCase "{{ header }}Settings" so in our case I would keep the plural. What do you think?

willydouhard (approved):

PR #367, opened 2023-09-06
Added mechanism that allows to inject custom CSS files for UI customi…

onepointconsulting:
This pull request adds another parameter to the configuration file `config.toml`. It allows to add a custom CSS file so that you can customize the CSS of the UI. This CSS file will be then injected into the index.html of the frontend React project.

Here is how you inject the custom CSS file:

```
[UI]
# Name of the app and chatbot.
name = "Chatbot"

...

# Custom CSS file that can be used to customize the UI.
custom_css = '/assets/test.css'
```

### DIFF
--- a/backend/chainlit/langchain/callbacks.py
+++ b/backend/chainlit/langchain/callbacks.py
@@ -1,16 +1,14 @@
 import json
 from typing import Any, Dict, List, Optional, Union
 
-from langchain.callbacks.base import AsyncCallbackHandler, BaseCallbackHandler
-from langchain.schema import AgentAction, AgentFinish, BaseMessage, LLMResult
-
+from chainlit import input_widget
 from chainlit.config import config
 from chainlit.context import context
 from chainlit.message import ErrorMessage, Message
 from chainlit.prompt import Prompt, PromptMessage
 from chainlit.sync import run_sync
-
-from chainlit import input_widget
+from langchain.callbacks.base import AsyncCallbackHandler, BaseCallbackHandler
+from langchain.schema import AgentAction, AgentFinish, BaseMessage, LLMResult
 
 IGNORE_LIST = []  # type: List[str]
 DEFAULT_ANSWER_PREFIX_TOKENS = ["Final", "Answer", ":"]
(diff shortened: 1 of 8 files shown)


willydouhard:
Interesting! Would love to learn a bit more about the use case before merging, can you provide details or a screenshot of your customised app?

onepointconsulting:
Hello, thank you very much for your quick reply. I really like Chainlit and am using to create chatbots. 
My use case is to make an image injected into Markdown to scale nicely horizontally so that it looks OK on all breakpoints. Here is a screenshot of the chatbot:
image
Without scaling the image it looks like this on mobile:
image
However if I manage to inject a custom CSS I can change the behaviour. These are the steps:
1. create a css file in `public/css` named `styles.css` with this content:
```
.markdown-body img {
    width: 100%
}
```
3. change config.toml by adding this line:
```
[UI]
# Name of the app and chatbot.
name = "Chatbot"

...

# Custom CSS file that can be used to customize the UI.
custom_css = '/public/css/styles.css'
```

After these changes the image scales nicely on mobile:
image

But there are many other customization possibilities with CSS, like adding backgrounds to your markdown, changing fonts that allow you to create a more customized experience.

willydouhard:
Makes sense, can you fix the conflict? `show_prompt_playground` has been moved to another config sub category, it should not be under the UI category anymore.

willydouhard:
Just opened https://github.com/Chainlit/chainlit/pull/371 to fix this issue for everyone. It does not invalidate this PR though :)

onepointconsulting:
I have tried to solve the conflict as requested. `show_prompt_playground` is not any longer in `class UISettings`. Please check. Many thanks.

backend/chainlit/langchain/callbacks.py
```diff
@@ -1,16 +1,14 @@
...
-
+from chainlit import input_widget
 from chainlit.config import config
 from chainlit.context import context
 from chainlit.message import ErrorMessage, Message
 from chainlit.prompt import Prompt, PromptMessage
 from chainlit.sync import run_sync
-
-from chainlit import input_widget
+from langchain.callbacks.base import AsyncCallbackHandler, BaseCallbackHandler
```
willydouhard, line 10: why are we updating this file? seems out of scope for this PR

frontend/public/assets/test.css
```diff
@@ -0,0 +1 @@
+img[src*="logo"] { display: none;}
```
willydouhard, line 1: we should not commit this imo

willydouhard (changes requested):

willydouhard:
Looks like the conflict is gone but new modifications appeared. See the changes requested.

onepointconsulting:
I made a mistake of pulling from the main branch. I am sorry. I will close this pull request and create a new one.

willydouhard:
No worries!

PR #373, opened 2023-09-07
Added mechanism to allow injecting custom CSS files for UI customization

onepointconsulting:
This pull request adds another parameter to the configuration file `config.toml`. It allows to add a custom CSS file so that you can customize the CSS of the UI. This CSS file will be then injected into the index.html of the frontend React project.

Here is how you inject the custom CSS file by editing `config.toml`:

```
[UI]
# Name of the app and chatbot.
name = "Chatbot"

...

# Custom CSS file that can be used to customize the UI.
custom_css = '/assets/test.css'
```

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -196,9 +196,7 @@ def get_html_template():
 
     css = None
     if config.ui.custom_css:
-        css = (
-            f"""<link rel="stylesheet" type="text/css" href="{config.ui.custom_css}">"""
-        )
+        css = f"""<link rel="stylesheet" type="text/css" href="{config.ui.custom_css}">"""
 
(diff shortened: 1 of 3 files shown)


backend/chainlit/server.py
```diff
@@ -193,13 +194,21 @@ def get_html_template():
     if config.ui.theme:
         js = f"""<script>window.theme = {json.dumps(config.ui.theme.to_dict())}</script>"""
 
+    css = None
+    if config.ui.custom_css:
+        css = (
+            f"""<link rel="stylesheet" type="text/css" href="{config.ui.custom_css}">"""
```
willydouhard, line 200: why do we need parenthesis here?

willydouhard:

Issue #21, opened 2023-05-31
Speech to text


allisonmorrell:
I'm experimenting with Chainlit and it is awesome. Kudos to you all. 

Future feature request: I would like to incorporate speech to text in the user interface for my application. 

Please label as enhancement.

willydouhard:
I can't contribute in the short term but this is a good first issue (small and scoped to the UI). If any of you wants to make a PR (fork the repo and then open a PR against the Chainlit main branch) I would be more than happy to review it!

allisonmorrell:
> Would you be able to contribute this feature? I would be willing to coach you with adding this functionality.

Unfortunately, I don't know any TypeScript/React. That said, we are hoping to involve more people in our project including someone with more front-end experience so if someone were to outline the recommended approach at a very high level I might be able to find someone who can contribute.

mmnasser2000:
Can someone assign this to me?

sqrt676:
Hi @willydouhard , I am also trying to do, if i come up wuth something good, I will ask you to assign it to me and I will send the request to review it. Thanks !

poojitharamachandra:
is this implemented?

alimtunc:
> poojitharamachandra

On it's way https://github.com/Chainlit/chainlit/issues/21#issuecomment-1760942049

poojitharamachandra:
do you know in which version this will be released?

poojitharamachandra:
is this available in the latest version 0.7.604? if so how to use it? 
https://docs.chainlit.io/backend/config/features

willydouhard:
Yes it is available in `0.7.604`. To enable it just, update your config to:

```toml
[features]
prompt_playground = true
multi_modal = true
# Allows user to use speech to text
[features.speech_to_text]
    enabled = true
    language = "en-US"
```

poojitharamachandra:
where should i make the above changes?

allisonmorrell:
`.chainlit/config.toml` – discussion of features section [here](https://docs.chainlit.io/backend/config/features).

poojitharamachandra:
thanks. where can I find this file in windows system?

willydouhard:
It is located in the same directory you ran `chainlit run ...` at `.chainlit/config.toml`.

poojitharamachandra:
i use the command chainlit run <file.py>  from the dir where my python file is present

poojitharamachandra:
thanks, now I see the option to record.
[...]

PR #375, opened 2023-09-07
Speech to text | Issue #21

mmnasser2000:


willydouhard:
Thank you for your contribution! Will review once we cut the auth release.

frontend/src/components/organisms/chat/inputBox/input.tsx
```diff
@@ -105,6 +121,24 @@ const Input = ({ onSubmit, onReply }: Props) => {
         </IconButton>
       )}
       <HistoryButton onClick={onHistoryClick} />
+      {browserSupportsSpeechRecognition &&
+        (isRecording ? (
+          <StopCircleIcon
```
willydouhard, line 126: Icons should be wrapped in a IconButton (right now the cursor is incorrect)

willydouhard (changes requested):
This looks great, there is a cursor problem that can be solved by wrapping the icons in a `IconButton`.

Also there is a new feature config. I think this should be configurable through that [configuration](https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/config.py#L133).

Finally there are minor conflicts with the current main.

willydouhard:
let me know @mmnasser2000

willydouhard:
Looks like there is a merge issue (this should not change 296 files)

backend/chainlit/config.py
```diff
@@ -76,6 +76,9 @@
 # The CSS file can be served from the public directory or via an external link.
 # custom_css = "/public/test.css"
 
+# Allows user to use speech to text
+# speech_to_text = true
```
willydouhard, line 80: I would put that config under features and not UI

willydouhard (changes requested):
This is awesome! Small comment on the config. Can you attach a small video demoing this new feature?

backend/chainlit/config.py
willydouhard: If you don't have time to change that I can do it, let me know

willydouhard (approved):

willydouhard:

PR #383, opened 2023-09-11
Auth0 inspired UX

clementsirieix:
https://github.com/Chainlit/chainlit/assets/10745953/a1b6db89-3b2b-469f-bb06-fb962249ac1b

⚠️ The issue during signin is fixed on the be

## Changes

- Use `react-password-checklist` to display password validation
- Reset error display on switching form
- Fix display of microsoft provider button

## How to test?

- Go to signup form
- Notice the password creation flow
- Notice that you can still login normally

### DIFF
--- a/libs/components/src/auth/AuthLogin.tsx
+++ b/libs/components/src/auth/AuthLogin.tsx
@@ -82,6 +82,10 @@ const AuthLogin = ({
   const [errorState, setErrorState] = useState(error);
 
   useEffect(() => {
+    setErrorState(undefined);
+    formik.resetForm();
+  }, [showSignIn]);
+  useEffect(() => {
     setErrorState(error);
   }, [error]);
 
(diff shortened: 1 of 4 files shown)


libs/components/src/auth/AuthLogin.tsx
```diff
@@ -91,101 +98,120 @@ const AuthLogin = ({
...
+              fontSize: 14
+            }}
           >
-            Continue
-          </Button>
-        </form>
-      ) : null}
+            Your password must contain:
+            <PasswordChecklist
+              rules={['minLength', 'specialChar', 'number', 'capital']}
```
willydouhard, line 186: this should be configurable, as chainlit app also use password auth and might not have those requiremetns

willydouhard (changes requested):

libs/components/src/auth/AuthLogin.tsx
clementsirieix: Made configurable 👍

PR #385, opened 2023-09-11
Add oauth implementation

tpatel:
- Moved JWT secret from the config file into an env variable


backend/chainlit/auth.py
```diff
@@ -1,18 +1,25 @@
...
 
-from chainlit import config
-
 reuseable_oauth = OAuth2PasswordBearer(tokenUrl="/login", auto_error=False)
 
-# TODO: make this an env variable with CLI command
-secret = random_secret()
+
+def get_jwt_secret():
+    return os.environ.get("CHAINLIT_JWT_SECRET")
```
willydouhard, line 15: maybe rename CHAINLIT_AUTH_SECRET?

backend/chainlit/cli/__init__.py
```diff
@@ -160,3 +164,13 @@ def chainlit_hello(args=None, **kwargs):
...
     init_config(log=True)
+
+
+@cli.command("create-secret")
+@click.argument("args", nargs=-1)
+def chainlit_create_secret(args=None, **kwargs):
+    trace_event("chainlit secret")
+
+    print(
+        f"Copy the following secret into your .env file:\nCHAINLIT_JWT_SECRET={random_secret()}"
```
willydouhard, line 175: maybe add a warning that if it is lost users will not be able to authenticate anymore

frontend/src/hooks/auth.ts
```diff
@@ -17,25 +42,39 @@ export const useAuth = () => {
...
+      logout();
       return;
     }
     try {
-      const { exp, ...AppUser } = JSON.parse(atob(accessToken.split('.')[1]));
+      const { exp, ...AppUser } = JSON.parse(atob(token.split('.')[1]));
+      setToken(token);
+      setAccessToken(`Bearer ${token}`);
       setUser(AppUser as IAppUser);
+      console.log('AppUser', AppUser);
```
willydouhard, line 64: remove console log

PR #398, opened 2023-09-14
Fix/build

willydouhard:


libs/components/package.json
```diff
@@ -14,39 +14,67 @@
   "author": "Chainlit",
   "license": "Apache-2.0",
   "version": "0.0.2",
-  "main": "dist/src/index.js",
+  "type": "module",
+  "main": "dist/src/index.cjs",
```
clementsirieix, line 18: `exports` overrides those properties, but it's contradictory to have both a module type and a main commonjs file.

```suggestion
  "main": "dist/src/index.js",
```

clementsirieix (approved):
I've put 1 change that doesn't impact the build.
Let's go 🚀

PR #401, opened 2023-09-15
Adds support for Okta OAuth provider 

stian-a-johansen:
This PR aims to add support for Okta based logins to chainlit. to use the OktaOauthProvider class set the following environment variables:
OAUTH_OKTA_CLIENT_ID
OAUTH_OKTA_CLIENT_SECRET
OAUTH_OKTA_DOMAIN


backend/chainlit/oauth_providers.py
```diff
@@ -185,7 +185,61 @@ async def get_user_info(self, token: str):
...
 
-providers = [GithubOAuthProvider(), GoogleOAuthProvider(), AzureADOAuthProvider()]
+class OktaOAuthProvider(OAuthProvider):
+    id = "okta"
+    env = [
+        "OAUTH_OKTA_CLIENT_ID",
+        "OAUTH_OKTA_CLIENT_SECRET",
+        "OAUTH_OKTA_DOMAIN",
+    ]
+    authorize_url = f"https://{os.environ.get('OAUTH_OKTA_DOMAIN')}/oauth2/default/v1/authorize"
```
willydouhard, line 195: maybe check trailing slash?

backend/chainlit/oauth_providers.py
```diff
@@ -185,7 +185,61 @@ async def get_user_info(self, token: str):
...
+                if not token:
+                    raise HTTPException(
+                        status_code=400, detail="Failed to get the access token"
+                    )
+                return token
+
+    async def get_user_info(self, token: str):
+        async with aiohttp.ClientSession(raise_for_status=True) as session:
+            async with session.get(
+                f"https://{os.environ.get('OAUTH_OKTA_DOMAIN')}/oauth2/default/v1/userinfo",
```
willydouhard, line 231: same comment for trailing slash

willydouhard:
This is really cool! I think we probably need to add the Okta provider in the [UI](https://github.com/Chainlit/chainlit/blob/main/libs/components/src/auth/ProviderButton.tsx) also. Looks promising!

willydouhard:
Awesome, I think we should add okta to the [provider literal](https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/types.py#L89).

If this is working locally I would also love to see a screen recording!

willydouhard:
> > Awesome, I think we should add okta to the [provider literal](https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/types.py#L89).
> > If this is working locally I would also love to see a screen recording!
> > Once merged we should also add it to the [documentation](https://docs.chainlit.io/authentication/oauth).
> 
> Great stuff, thanks for the help @willydouhard! I have added the provider literal. Right now our okta isn't set up to allow the callback url directly to the chainlit app, so will have to wait until next week for it to be approved, but will send over a video then 👍 Right now it ends in a callback error on the Okta side when using the button.

Awesome, this looks promising!

willydouhard:
Any update @stian-a-johansen ?

stian-a-johansen:
> Any update @stian-a-johansen ?

Hey Willy, it took a while to get the settings updated on the okta side, but here's a video: Apologies for the vibrating black box at the top!


https://github.com/Chainlit/chainlit/assets/9048717/12b466ff-e488-4715-a020-992ca61c9bbc

willydouhard:
It looks great! Are we missing anything or should we do a last review to merge?

stian-a-johansen:
As far as I see I think it's just docs that's missing now @willydouhard 👍  Thank you!

willydouhard:
the `""` is breaking because it is [inside a f string](https://github.com/Serendipity-AI/chainlit/blob/SJ/okta_provider/backend/chainlit/oauth_providers.py#L196). Almost there!

stian-a-johansen:
> the `""` is breaking because it is [inside a f string](https://github.com/Serendipity-AI/chainlit/blob/SJ/okta_provider/backend/chainlit/oauth_providers.py#L196). Almost there!

🤦 now it's better!

stian-a-johansen:
> CI is green! I am not able to test it myself, you confirm everything works on your end?

Yep! Logs in and works here 👍  Very nice the ability to support these!

stian-a-johansen:
Hm, looks like the literals have been removed from types.py in another merge, trying to see where it should go now

willydouhard:
> Hm, looks like the literals have been removed from types.py in another merge, trying to see where it should go now

Yes, I just merged another PR. It goes in client/base.py.

willydouhard:
Looks good! Going to merge. If you can add a paragraph [here](https://docs.chainlit.io/authentication/oauth) with a code example it would be awesome!

stian-a-johansen:
Just merged in the latest from `main` into this branch

libs/components/src/auth/ProviderButton.tsx
```diff
@@ -1,9 +1,11 @@
 import { grey } from 'theme/palette';
 
+
 import GitHub from '@mui/icons-material/GitHub';
 import Google from '@mui/icons-material/Google';
 import Microsoft from '@mui/icons-material/Microsoft';
 import Button from '@mui/material/Button';
+import Okta from './Okta.tsx';
```
willydouhard, line 8: The icon is exported as a const and not default. Either change this line to `import {Okta} from ...` or export the icon as default

willydouhard (changes requested):

libs/components/src/auth/ProviderButton.tsx
willydouhard: This makes the build of libs/components fail

willydouhard (approved):

willydouhard:
Approved! Thank you for your work 🎉

stian-a-johansen:
> Approved! Thank you for your work 🎉

Thanks for the merge and the help!

PR #410, opened 2023-09-21
update to mypy 1.5.1 and fix post_init liskov

willydouhard:
Mypy 1.5.1 was complaining

```
chainlit/input_widget.py:17: error: Signature of "__post_init__" incompatible with supertype "dataclass"  [override]
chainlit/input_widget.py:17: note:      Superclass:
chainlit/input_widget.py:17: note:          def __post_init__(self: InputWidget) -> None
chainlit/input_widget.py:17: note:      Subclass:
chainlit/input_widget.py:17: note:          def __post_init__(self: InputWidget, tooltip: Any, description: Any) -> None
Found 1 error in 1 file (checked 48 source files)
 ELIFECYCLE  Command failed with exit code 1.
```

So i took a look at the input_widget file. To me there is a lot of dead code (plus the post_init was breaking mypy). Would love to have feedback on this to make sure I did not miss on something important.

Tested with prompt playground and chat settings and everything looks ok.

### DIFF
--- a/backend/chainlit/input_widget.py
+++ b/backend/chainlit/input_widget.py
@@ -1,4 +1,4 @@
-from abc import ABC, abstractmethod
+from abc import abstractmethod
 from collections import defaultdict
 from typing import Any, Dict, List, Optional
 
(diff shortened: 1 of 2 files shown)


backend/chainlit/input_widget.py
```diff
@@ -7,20 +7,19 @@
 
 
 @dataclass
-class InputWidget(ABC):
+class InputWidget:
```
MathiasSpanhove, line 10: Is ABC not required when you use @abstractmethod?

https://docs.python.org/3/library/abc.html#abc.abstractmethod

MathiasSpanhove (approved):
LGTM, completely slipped my mind that @dataclass generates the __init__ for you.

backend/chainlit/input_widget.py
willydouhard: It does not break but maybe I should put it back?

MathiasSpanhove: I think that you are now able to instantiate InputWidget, while otherwise you would get an exception that it can't be instantiated since it's abstract.

PR #420, opened 2023-09-24
Added comment to document the usage of the custom_css parameter in th…

onepointconsulting:
Added comment to document the usage of the custom_css parameter in the generated config.toml. This can be helpful for developers to find the parameter.

### DIFF
--- a/backend/chainlit/config.py
+++ b/backend/chainlit/config.py
@@ -64,6 +64,9 @@ hide_cot = false
 # Link to your github repo. This will add a github button in the UI's header.
 # github = ""
 
+# Specify a CSS file that can be used to customize the user interface.
+# custom_css = '/assets/test.css'
+
 # Override default MUI light theme. (Check theme.ts)
 [UI.theme.light]
     #background = "#FAFAFA"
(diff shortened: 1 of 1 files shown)


backend/chainlit/config.py
```diff
@@ -64,6 +64,9 @@
 # Link to your github repo. This will add a github button in the UI's header.
 # github = ""
 
+# Specify a CSS file that can be used to customize the user interface.
+# custom_css = '/assets/test.css'
```
willydouhard, line 68:
```suggestion
# custom_css = '/public/test.css'
```

I think most people will serve the css either from the public dir or an external link

onepointconsulting: I have updated the pull request. Please check.

backend/chainlit/config.py
```diff
@@ -65,7 +65,8 @@
 # github = ""
 
 # Specify a CSS file that can be used to customize the user interface.
-# custom_css = '/assets/test.css'
+# The CSS fle can be served from the public directory or via an external link.
```
willydouhard, line 68:
```suggestion
# The CSS file can be served from the public directory or via an external link.
```

Typo, almost there!

onepointconsulting: Should be fixed. Please check.

willydouhard (approved):

PR #428, opened 2023-09-26
add on chat end

willydouhard:


sumba101 (commented):
this would be a very necessary and helpful addition to the python chainlit library, please consider reviewing and merging

PR #429, opened 2023-09-26
Rework actions to be able to control collapse yourself

MathiasSpanhove:
I saw some feedback from people that they didn't like the fact that actions were force collapsed when there were more than 2. I tried to implement a way where users can try to control which actions are collapsed themselves if they care, otherwise actions will be wrapped. This wrapping is especially useful when the actions are long follow up questions to a previous question.

image

I'm not sure which one is better... I think I like `justify-content: start` more now that I look at it side by side.

`justify-content: center`
image
image

`justify-content: start`
image
image

```python
import chainlit as cl


@cl.on_chat_start
async def main():
    actions = [
        cl.Action(label="Button 0", name="load_index", value="Button 0"),
        cl.Action(label="Button 1", name="load_index", value="Button 1", collapsed=True),
        cl.Action(label="Button 2", name="load_index", value="Button 2"),
        cl.Action(label="Button 3", name="load_index", value="Button 3"),
        cl.Action(label="Button 4", name="load_index", value="Button 4"),
        cl.Action(label="Button 5", name="load_index", value="Button 5"),
        cl.Action(label="Button 6", name="load_index", value="Button 6"),
        cl.Action(label="Button 7", name="load_index", value="Button 7", collapsed=True),
        cl.Action(label="Button 8", name="load_index", value="Button 8"),
        cl.Action(label="Button 9", name="load_index", value="Button 9"),
        cl.Action(label="Button 10", name="load_index", value="Button 10"),
        cl.Action(label="Button 10", name="load_index", value="Button 11"),
        cl.Action(label="Button 10", name="load_index", value="Button 12"),
        cl.Action(label="Button 10", name="load_index", value="Button 13"),
        cl.Action(label="Button 10", name="load_index", value="Button 14")
    ]

    await cl.Message(
        author="System", 
        content="Type a message to start chatting or ask questions about a topic by pressing the corresponding action below.", 
        actions=actions
    ).send()
```

What do you guys think?

I'll fix the failed test tomorrow, `pnpm test` didn't work on my local setup so have to look into that.


willydouhard:
Good work! I like the idea (especially giving control over which action is collapsed). Two comments:
1. We should probably still have a default wrapping policy for mobile
2. Why are the "..." and prompt playground buttons far on the right?

MathiasSpanhove:
>     1. We should probably still have a default wrapping policy for mobile

All actions on mobile are automatically collapsed like before, or are you talking about something else?

>     2. Why are the "..." and prompt playground buttons far on the right?

Consistency from a UI/UX perspective. I like it best when buttons are always on the exact same spot. Right now the human feedback buttons might be moved to the right when the prompt playground button is enabled. And the "..." is now always in the same spot no matter how many actions are shown. 

Had a busy day today, so will try to finish this PR tomorrow with any remaining feedback you have. Would you like the `justify-content: start or end`?

willydouhard:
1. I overlooked it, would be awesome to have a screenshot of the mobile view.
2. I see! In that case should we put the actions on the left and feedback + prompt playground on the right? Also the more action button should be a Vertical "..." rather than Horizontal in that setup.  Plus the feedback + prompt playground icons should not be vertically centered?

Take your time, no rush to merge this 👍

MathiasSpanhove:
I've implemented your suggestions:

image
image
image
image
image

Mobile:
image

willydouhard:
I think it's better! If no actions are provided should the buttons (prompt playground, feedback) still be aligned on the right?

MathiasSpanhove:
> I think it's better! If no actions are provided should the buttons (prompt playground, feedback) still be aligned on the right?

I would say yes for UI/UX consistency. Otherwise the buttons may jump from left to right in one conversation.

willydouhard (approved):

PR #436, opened 2023-09-29
Config/baseurl

willydouhard:


backend/chainlit/server.py
```diff
@@ -206,6 +207,28 @@ def get_html_template():
         return content
 
 
+def get_user_facing_url(url: URL):
+    """
+    Return the user facing URL for a given URL.
+    Handles proxies and subdirectories installs (example: https://example.com/my-chainlit-app/).
```
tpatel, line 213: I think you should remove the subdirectory install mention. I wrote that before I knew that there are other pieces in Chainlit that aren't compatible with subdirectory installs.

tpatel (approved):
✨✨✨ AWESOME ✨✨✨

PR #446, opened 2023-10-03
Rework user history

alimtunc:
This PR offer a new way to browse your chat history.

https://github.com/Chainlit/chainlit/assets/18298759/56c6b9ed-a0ab-4e56-9b53-dd71f94f878a


frontend/src/pages/Page.tsx
```diff
@@ -44,7 +45,10 @@ const Page = ({ children }: Props) => {
       {!isAuthenticated ? (
         <Alert severity="error">You are not part of this project.</Alert>
       ) : (
-        children
+        <Stack direction="row" height="100%" width="100vw" overflow="auto">
```
willydouhard, line 48: can't we do with 100% here? i don't like vw that much

frontend/src/router.tsx
```diff
@@ -25,16 +23,8 @@ export const router = createBrowserRouter([
...
-      <Page>
-        <Conversation />
-      </Page>
-    )
-  },
-  {
-    path: '/dataset',
-    element: <Dataset />
+    path: '/conversations/:id?',
+    element: <Conversations />
```
willydouhard, line 27: shouldnt that component be named Conversation? Or ConversationPage?

libs/components/src/playground/chat.tsx
```diff
@@ -57,7 +57,7 @@ export const ChatPromptPlayground = forwardRef(
         : 'Prompt messages';
 
     return (
-      <Stack sx={{ width: '100%' }}>
+      <Stack sx={{ width: '100%', height: '100%' }}>
```
willydouhard, line 60: why are we touching the prompt playground?

frontend/src/pages/Page.tsx
alimtunc: Actually looks like we don't need the width

willydouhard (approved):

PR #447, opened 2023-10-03
[Prompt Playground] Don't show scrollbars if not needed

MathiasSpanhove:
Before:
image

After:
image


willydouhard (approved):
good catch!

PR #463, opened 2023-10-09
Reduce perceived latency for users

clementsirieix:
Before patch:

https://github.com/Chainlit/chainlit/assets/10745953/c235c382-3fbb-47db-8e04-fa51d0466d67

After patch:

https://github.com/Chainlit/chainlit/assets/10745953/5036c8c3-6c3b-48c6-9c88-2be070d856ed

## Changes

- In `message.py` parallelize message persistence
- In `element.py` parallelize element persistence

## How to test?

- Upload an element
- Notice that element can be uploaded and accessed
- Send a message
- Notice that persistence occurs after

### DIFF
--- a/backend/chainlit/element.py
+++ b/backend/chainlit/element.py
@@ -102,19 +102,19 @@ class Element:
             self.object_key = upload_res["object_key"]
         element_dict = await self.with_conversation_id()
 
-        asyncio.create_task(self._persist(element_dict, client))
+        asyncio.create_task(self._persist(element_dict))
 
         return element_dict
 
-    async def _persist(self, conversation: ElementDict, client: ChainlitCloudClient):
+    async def _persist(self, element: ElementDict):
         if not chainlit_client:
             return
 
         try:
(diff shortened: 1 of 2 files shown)


backend/chainlit/element.py
```diff
@@ -98,18 +100,25 @@ async def persist(self, client: ChainlitCloudClient) -> Optional[ElementDict]:
...
-            )
-            self.persisted = True
-        else:
-            element_dict = await client.update_element(
-                await self.with_conversation_id()
-            )
         return element_dict
 
+    async def _persist(self, conversation: ElementDict, client: ChainlitCloudClient):
+        if not chainlit_client:
```
willydouhard, line 110: what is the client parameter? the code use the global chainlit_client

willydouhard: the naming is confusing, conversation is of type ElementDict

willydouhard (changes requested):
Looks good, few naming changes required.

PR #517, opened 2023-10-31
feat: add the Plotly element

tpatel:
- Enables creating interactive charts using the plotly library
- Fixed the MessageContent width to always fill the full width even if the message is short


backend/chainlit/langchain/callbacks.py
```diff
@@ -344,14 +344,27 @@ def _build_llm_settings(
         return provider, settings
 
 
+DEFAULT_TO_IGNORE = ["RunnableSequence", "RunnableParallel"]
```
willydouhard, line 347: we shoud add these changes in this PR.

libs/components/package.json
```diff
@@ -67,10 +67,12 @@
     "formik": "^2.4.3",
     "lodash": "^4.17.21",
     "mui-chips-input": "2.0.0",
+    "plotly.js": "^2.27.0",
     "react-dropzone": "^14.2.3",
     "react-file-icon": "^1.3.0",
     "react-markdown": "^8.0.7",
     "react-password-checklist": "^1.5.0",
+    "react-plotly.js": "^2.6.0",
```
willydouhard, line 75: do we need both the dependencies?

libs/components/src/messages/components/MessageContent.tsx
```diff
@@ -122,7 +122,7 @@ const MessageContent = memo(
 
     return (
       <Stack width="100%" direction="row">
-        <Box>
+        <Box width="100%">
```
willydouhard, line 125: why is this needed?

willydouhard (changes requested):

libs/components/package.json
tpatel: yes, `plotly.js` is a peer dependency of `react-plotly.js`.

libs/components/src/messages/components/MessageContent.tsx
tpatel: Here is what happens without it: when you have a short message, the container doesn't fill the full message column, so the elements are squeezed (the elements only have `max-width`, no `min-width` nor `width`).

image

tpatel: With the change:

image

willydouhard (changes requested):
remove prisma dependencies?

tpatel:
Thanks for the fix, I think the PR is ready now 👌

PR #539, opened 2023-11-12
Add token field to AppUser and function to update

JayRHa:
Added support to save token in the user_session and add additional scopes the EntraId toke with the optional enviroment var: OAUTH_OKTA_AUTHORIZATION_SERVER_ID

### DIFF
--- a/backend/chainlit/client/base.py
+++ b/backend/chainlit/client/base.py
@@ -41,6 +41,7 @@ class AppUser(DataClassJsonMixin):
     role: Role = "USER"
     tags: List[str] = Field(default_factory=list)
     image: Optional[str] = None
+    token: Optional[str] = None
     provider: Optional[Provider] = None
 
 
--- a/backend/chainlit/oauth_providers.py
+++ b/backend/chainlit/oauth_providers.py
@@ -158,10 +158,16 @@ class AzureADOAuthProvider(OAuthProvider):
     def __init__(self):
         self.client_id = os.environ.get("OAUTH_AZURE_AD_CLIENT_ID")
         self.client_secret = os.environ.get("OAUTH_AZURE_AD_CLIENT_SECRET")
+
+        scope = ["https://graph.microsoft.com/User.Read"]
+        if os.environ.get("OAUTH_AZURE_AD_SCOPE"):
+            scope.append(os.environ.get("OAUTH_AZURE_AD_SCOPE"))
+        scope = " ".join(scope)
+
         self.authorize_params = {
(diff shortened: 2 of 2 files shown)


JayRHa:
This would be very important for us that you merge this change

willydouhard:
In the meantime you can add this at the beginning of your `app.py` file to override the scope:

```py
from chainlit.oauth_providers import providers, AzureADOAuthProvider


azure_provider = next((p for p in providers if isinstance(p, AzureADOAuthProvider)), None)

if azure_provider:
    azure_provider.authorize_params["scope"] = ["..."]
```

hayescode:
I 2nd this PR, it will be very good to perform additional API calls for specific roles/priviledges, for example.

JayRHa:
It would also be awesome if we can cail token for the scope https://database.windows.net/.default

JayRHa:
Yes this works to get other scopes for graph api but not to get other tokens for other microsoft apis like azure or database.

willydouhard:
I see, I agree with the concept but this will clash with the 1.0.0 release. Will revisit this in 1/2 weeks

hayescode:
@willydouhard can we expect this in a future release? My personal goal with this is to 1) limit access to users in certain security groups and 2) give users settings/features based on security groups.

hayescode:
> I see, I agree with the concept but this will clash with the 1.0.0 release. Will revisit this in 1/2 weeks

@willydouhard will this functionality be included in 1.0.0?

willydouhard:
You will be able to set metadata to users and then differentiate access to users based on that metadata. The release will happen on Monday 10AM CET

willydouhard:
Do you confirm this is fixed with the 1.0.0 release?

JayRHa:
How does this work? Have an example

willydouhard:
Following the [doc](https://docs.chainlit.io/authentication/oauth#allow-all-users-who-passed-the-oauth-authentication)

```py
from typing import Dict, Optional
import chainlit as cl

@cl.oauth_callback
def oauth_callback(
    provider_id: str,
    token: str,
    raw_user_data: Dict[str, str],
    default_user: cl.User,
) -> Optional[cl.User]:
    metadata = default_user.metadata.copy()
    metadata["oauth_token"] = token
    metadata["my_custom_field"] = "my_custom_value"
    return cl.User(identifier=default_user.identifier, metadata=metadata)
 ```
 
 Then you can access the user (including the metadata) in any Chainlit decorator

```py
@cl.on_chat_start
async def start():
    user = cl.user_session.get("user")  # type: cl.User
    print(user)
```

willydouhard:
Closing this. Please reopen if the issue is not solved.

PR #600, opened 2023-12-16
various bug fixes

willydouhard:


Zahidpichen1 (commented):
What is the use of adding the code lines from 256 to 261

image

Zahidpichen1 (approved):
It was a good idea to remove `<3.12` in line 22. So that there won't be any conflict in the future

willydouhard:
> What is the use of adding the code lines from 256 to 261
> 
> image

In chainlit 1.0.0 we wait for the first user interaction with the chat to start persisting the messages. Until then the persist operations are stored in a queue.

PR #610, opened 2023-12-21
feat(playground): Add Gemini Pro in the LLM playground

fvaleye:
Add [Gemini](https://cloud.google.com/vertex-ai/docs/generative-ai/model-reference/gemini) Pro from GCP in the LLM Playground

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -85,7 +85,9 @@ module = [
     "socketio.*",
     "uptrace",
     "syncer",
-    "vertexai.language_models"
+    "vertexai.language_models",
+    "vertexai.preview.generative_models",
+    "google.cloud"
 ]
 ignore_missing_imports = true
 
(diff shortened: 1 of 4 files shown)


backend/pyproject.toml
```diff
@@ -85,7 +85,9 @@ module = [
     "socketio.*",
     "uptrace",
     "syncer",
-    "vertexai.language_models"
+    "vertexai.language_models",
```
willydouhard, line 88: I would not add those as dependencies. We should never import the llm providers at top level of the file but always in methods. Then if a user wants to use it he/she will have to manually install this package.

The reason is that we don't want to bloat the package with too many dependencies.

fvaleye: It's for mypy missing import error only.

willydouhard: Then it should probably go with the rest of the [optional dependencies](https://github.com/Chainlit/chainlit/blob/main/backend/pyproject.toml#L72C1-L73C1)?

willydouhard (approved):

PR #621, opened 2023-12-31
Adding streaming functionality to output for haystack callback manager

datapay-ai:
I probably need to adjust this this so that there is a toggle turn streaming on and off, but just wanted to get an idea as to whether or not the way I implemented the code is in line with what you would typically be looking for. What I found is basically that Haystack does stream tokens, but it just doesn't have any obvious implementation for capturing the final answer as a stream (at least not for < 2.0). I played around with this locally and managed to get this to work, although I did notice a few differences between my version and the current main branch (which I adjusted for).

In short:
1. Is this a feature you would be interested in merging?
2. Are there any tests I have to put in place before something like this could be merged?
3. Are there any code style changes or implementation concerns that I should fix?

Kindly let me know. Cheers!

### DIFF
--- a/backend/chainlit/haystack/callbacks.py
+++ b/backend/chainlit/haystack/callbacks.py
@@ -71,10 +71,11 @@ class HaystackAgentCallbackHandler:
         self.stack.push(run_step)
 
     def on_agent_finish(self, agent_step: AgentStep, **kwargs: Any) -> None:
-        run_step = self.last_step
-        run_step.end = datetime.utcnow().isoformat()
-        run_step.output = agent_step.prompt_node_response
-        run_sync(run_step.update())
+        if self.last_step:
+            run_step = self.last_step
+            run_step.end = datetime.utcnow().isoformat()
+            run_step.output = agent_step.prompt_node_response
+            run_sync(run_step.update())
 
     # This method is called when a step has finished
     def on_agent_step(self, agent_step: AgentStep, **kwargs: Any) -> None:
--- a/backend/chainlit/haystack/callbacks.py
+++ b/backend/chainlit/haystack/callbacks.py
@@ -57,7 +57,7 @@ class HaystackAgentCallbackHandler:
 
         if self.stream_final_answer:
             self.final_stream = Message(author=self.stream_final_answer_agent_name, content="")
-            self.last_tokens: list[str] = []
+            self.last_tokens: List[str] = []
             self.answer_reached = False
 
         root_message = context.session.root_message
--- a/backend/chainlit/haystack/callbacks.py
+++ b/backend/chainlit/haystack/callbacks.py
@@ -36,7 +36,7 @@ class HaystackAgentCallbackHandler:
     stack: Stack[Step]
     last_step: Optional[Step]
 
-    def __init__(self, agent: Agent):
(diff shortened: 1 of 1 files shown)


willydouhard:

1. We are interested in merging this feature! Will it still be relevant with Haystack 2.0?
2. You can try to add a test but testing 3rd party libraries is sometimes painful, as it often conflicts with other 3rd party library requirements
3. Code style seems ok! As long as it passes our CI (mypy + black).

willydouhard:
Could you fix this [mypy warning](https://github.com/Chainlit/chainlit/actions/runs/7370797896/job/20058144445?pr=621)?

datapay-ai:
> Thank you for your contribution!
> 
> 1. We are interested in merging this feature! Will it still be relevant with Haystack 2.0?
> 2. You can try to add a test but testing 3rd party libraries is sometimes painful, as it often conflicts with other 3rd party library requirements
> 3. Code style seems ok! As long as it passes our CI (mypy + black).

2. Ah I see. Well let me do some more extensive manual testing to make sure there aren't any silly edge cases, and I will also add a flag so that at the very least it's an opt-in (because it probably should + it means that if there is any bug / issue, it won't break existing stuff)
3. Ok, sounds good! I think I need to add a couple types, but that should be easy.

Will let you guys know when I am done!

datapay-ai:
> Could you fix this [mypy warning](https://github.com/Chainlit/chainlit/actions/runs/7370797896/job/20058144445?pr=621)?

Yep, will do.

datapay-ai:
I have now tested it locally and it seems to be working fine. I have added two flags (one for streaming the final answer output, and another for choosing the name of the agent of the streamed final answer output. I fixed what seemed to be a bug in the code (before my changes were added) where running it would cause a crash due to an attempt from a piece of code to pop from an empty list:
image

I also added a type to self.last_tokens so that the mypy error should hopefully stop.

Let me know if this looks fine. Also, if you have some time at some point, it would be great to get some help as to how to install a fork of chainlit as a package with poetry + pyproject.toml. The pyproject.toml being in the backend folder of the chainlit project seems to be causing me some issues when trying to install my chainlit fork as a library, and if I have time to make more MRs in the future, being able to do that will make it a lot easier to test

datapay-ai:
Fixed another issue related to stack popping that would cause the program to fail due to an empty stack when `not agent_step.is_last()` (noticed this when using it with my haystack agent that has tools, and thus has multiple steps)

willydouhard:
Thank you for the bug fixes! There is still a [mypy issue ](https://github.com/Chainlit/chainlit/actions/runs/7376572932/job/20072505909?pr=621#step:6:1). Let me know if you need help on this one

datapay-ai:
> Thank you for the bug fixes! There is still a [mypy issue ](https://github.com/Chainlit/chainlit/actions/runs/7376572932/job/20072505909?pr=621#step:6:1). Let me know if you need help on this one

Seems like this is because I was using the : list[str] syntax rather than : List[str] (guessing this must be due to the python or mypy version). I have changed it to List[str] now so it should work now I believe.

datapay-ai:
Ok, so I used mypy locally, and the problem should be fixed now.

willydouhard (approved):
really cool! Thank you for your contribution!

PR #622, opened 2024-01-01
Add Chat Settings to Generic Langchain Provider

tylertitsworth:


willydouhard:
Thank you for your contribution! Can you provide a small code example demoing this change? Thank you and happy new year!

tylertitsworth:
Hi, thank you for the kind welcome! Here is a simplified version of my workflow: https://github.com/tylertitsworth/multi-mediawiki-rag/blob/tylertitsworth/example/main.py

This workflow assumes you already have the Ollama installed and serving, as well as a vectordb to retrieve from. I'm sure there are simpler workflows that don't involve RAG, but this is the closest approximation to what I am hosting.

Langchain has many different provider types, the type that I use is Ollama. The issue found in https://github.com/langchain-ai/langchain/issues/14980 is avoided by using langchain v0.0.349. This issue may not be present in other providers.

willydouhard:
This is a cool use case. I did not even know that you could update a provider on the fly with the chat settings! The CI is failing because you did not add the input parameter [here](https://github.com/tylertitsworth/chainlit/blob/main/cypress/e2e/prompt_playground/provider.py#L82). Maybe the simplest fix is to make the `inputs` arg default to `[]`?

willydouhard:
There is a last issue to fix, you are using the <1.0.0 code for the langchain generic provider. On the main branch we are at 1.0.0. The code changed slightly for LLM providers. For instance [this line](https://github.com/Chainlit/chainlit/blob/cadb0cb103598d287e9bf91e4cd439c479bee4a6/backend/chainlit/playground/providers/langchain.py#L68) is problematic compared to [main](https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/playground/providers/langchain.py#L66)

tylertitsworth:
Thanks for the help! I didn't realize that `require_settings()` was the line blocking the test from succeeding.

backend/chainlit/playground/providers/langchain.py
```diff
@@ -65,7 +67,11 @@ async def create_completion(self, request):
 
         messages = self.create_generation(request)
 
-        stream = make_async(self.llm.stream)
+        # https://github.com/langchain-ai/langchain/issues/14980
+        stream = make_async(self.llm.stream)(
+            input=messages,
+            **request.prompt.settings
```
willydouhard, line 73: This will also fail. There is no longer a `prompt` attribute in the request. See the [anthropic](https://github.com/Chainlit/chainlit/blob/cadb0cb103598d287e9bf91e4cd439c479bee4a6/backend/chainlit/playground/providers/anthropic.py#L32) example.

willydouhard (changes requested):

backend/chainlit/playground/providers/langchain.py
willydouhard: Also can we pass the settings directly to the stream method? Does this work for all Langchain LLMs? Just curious

tylertitsworth: This works so long as the `id` of the setting matches a parameter name in the chat model function. Langchain seems to use the same name convention for all of their extra parameters for chat models

- https://api.python.langchain.com/en/latest/chat_models/langchain_community.chat_models.anthropic.ChatAnthropic.html?highlight=chatanthropic#
- https://api.python.langchain.com/en/latest/chat_models/langchain_community.chat_models.ollama.ChatOllama.html?highlight=chatollama#langchain_community.chat_models.ollama.ChatOllama

However, this is not always the case, like in https://api.python.langchain.com/en/latest/chat_models/langchain_community.chat_models.openai.ChatOpenAI.html?highlight=chatopenai#langchain_community.chat_models.openai.ChatOpenAI.

backend/chainlit/playground/providers/langchain.py
```diff
@@ -65,7 +67,11 @@ async def create_completion(self, request):
 
         messages = self.create_generation(request)
 
-        stream = make_async(self.llm.stream)
+        # https://github.com/langchain-ai/langchain/issues/14980
+        stream = make_async(self.llm.stream)(
```
willydouhard, line 71: The CI fails because stream is already a coroutine instead of a callable. We can do

```py
        result = await make_async(self.llm.stream)(
            input=messages, **request.generation.settings
        )
```
to replace stream in result in one.

willydouhard (changes requested):

willydouhard:

PR #648, opened 2024-01-09
Set oauth cookie to secure=True when using same-site=none

hayescode:
When setting the env var `CHAINLIT_COOKIE_SAMESITE=none` the `secure` parameter in the browser cookie needs to be set to `secure=True`. By default, `secure=False` so adding a boolean conditional and explicitly passing in the `secure` variable.

[Discord Discussion](https://discord.com/channels/1088038867602526210/1179183494124019852/1184663263070388244)

image

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -318,8 +318,9 @@ async def oauth_login(provider_id: str, request: Request):
         url=f"{provider.authorize_url}?{params}",
     )
     samesite = os.environ.get("CHAINLIT_COOKIE_SAMESITE", "lax")  # type: Any
+    secure = samesite.lower() == 'none'
     response.set_cookie(
-        "oauth_state", random, httponly=True, samesite=samesite, max_age=3 * 60
+        "oauth_state", random, httponly=True, samesite=samesite, secure=secure, max_age=3 * 60
     )
     return response
 


backend/chainlit/server.py
```diff
@@ -318,8 +318,9 @@ async def oauth_login(provider_id: str, request: Request):
         url=f"{provider.authorize_url}?{params}",
     )
     samesite = os.environ.get("CHAINLIT_COOKIE_SAMESITE", "lax")  # type: Any
+    secure = samesite.lower() == 'none'
```
willydouhard, line 321: both none and strict should be secure and only lax should be secure false?

hayescode: @willydouhard this code change only sets secure=True when `none` but i can easily update it if you'd like. I'm not sure of other people's use cases, but it could cause issues on localhost since it's http.

From GPT4:
For each SameSite option, here's what the Secure attribute should be set to:
- SameSite=Strict: The Secure attribute can be set to true to ensure the cookie is only sent over HTTPS, but it is not required to be true because the cookie is not sent in cross-site requests. It can be false if the site is not using HTTPS, but using HTTPS is highly recommended for security.
- SameSite=Lax: Similar to SameSite=Strict, the Secure attribute can be set to true to ensure the cookie is only sent over HTTPS, but it is not strictly necessary. However, setting it to true is a best practice for enhanced security. It can be false if the site is not using HTTPS, but again, using HTTPS is highly recommended.
- SameSite=None: The Secure attribute must be set to true. This is required because SameSite=None allows the cookie to be sent in cross-site requests, and without the Secure attribute being true, the cookie would be vulnerable to interception over an unencrypted connection (HTTP).

willydouhard (approved):

Issue #144, opened 2023-07-07
Make browser query parameters available in `@on_chat_start`


kevinwmerritt:
Feature request for query params to be made available in the `on_chat_start` decorator. From there you could add them to user session or take another action.

```python
@on_chat_start
async def main(query_params):
    document_url = query_params.get('document_url')
    if document_url:
        ....
```

Imagine an application that lists multiple documents and you want to design a chainlit that can Q&A one of those documents.

Possible workflow today:

1. The user downloads the document from the application.
2. In a separate browser tab they open the chainlit.
3. In the chainlit they upload the file with a `AskFileMessage` dialog.
4. Process the file and chat.

Possible workflow enabled by query params:

1. User clicks a [Chat] button next the document they want to Q&A and a new browser tab opens to mychainlit.xyz?document_url=/my/doc.pdf
3. The chainlit begins processing the file referenced in the document.

Additional scenarios could include:
- passing a reference to an ID to an API accessible resource
- chaining chainlit apps by passing fragments of one chat to another chainlit as part of a larger workflow

laurentiupiciu:
@willydouhard do you have any updates on this matter?

willydouhard:
Did not have the time to add it yet. I am unsure about the behaviour though. Let's say you connect to `http://localhost:8000?params={...}`. In this case the json would probably encoded in b64 but that's not the point.

So we send that to the socket connection headers. Now let's say I navigate in the app a few times and create a new session. Is it expected that the params are sent again? In that case that means we should store them in the local storage or something. Also what happens if I refresh the page without the `?params`?

laurentiupiciu:
from my perspective it should be something like in streamlit: https://docs.streamlit.io/library/api-reference/utilities/st.experimental_get_query_params

laurentiupiciu:
got it. thank you!

bachi76:
@willydouhard just give either the request object or at least the GET/POST/Headers as optional param to the`@cl.on_chat_start` method

Alternatively or additionally, do the same here:

```
@app.post("/auth/header")
async def header_auth(request: Request):
    if not config.code.header_auth_callback:
        raise HTTPException(
            status_code=status.HTTP_400_BAD_REQUEST,
            detail="No header_auth_callback defined",
        )

    app_user = await config.code.header_auth_callback(request.headers, add the request here)
    ...
```
 This is what I'm doing right now (have to patch library code, which I totally only do if no other options are possible. But we need this).
 
[...]

PR #682, opened 2024-01-22
Add query parameters

postonsundays:
Initial proposal for the enhancement proposed in #144 to use the browser's query parameters as part of the app (e.g. process initial query on app launch with "?q=[QUERY]").

This implementation loads the query parameter as part of the initial page data being fetched in the `useChatSession` hook, then directly into the `on_chat_start` call of the app as it may be used in the initial setup call. Open to suggestions about how else it be called.

### DIFF
--- a/backend/chainlit/session.py
+++ b/backend/chainlit/session.py
@@ -64,6 +64,8 @@ class BaseSession:
         root_message: Optional["Message"] = None,
         # Chat profile selected before the session was created
         chat_profile: Optional[str] = None,
+        # Query Params
+        query_params: Optional[Dict[str, str]] = None,
     ):
         if thread_id:
             self.thread_id_to_resume = thread_id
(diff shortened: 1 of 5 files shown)


willydouhard:
Thank you for the proposal! I think this will not work with authentication though:

1. The user hits https://my-chainlit-app.com?q...
2. The user is redirected to /login
3. The user authenticates and is redirected to https://my-chainlit-app.com
4. The websocket session is created without the query params

willydouhard:
The thing is that the decorators are ran once the websocket connection is established (hence after authentication). The easiest way to do it would be to store query params in the local storage to send it as ws connection time.

postonsundays:
That could work; another solution for this use case might be adding the original url as a redirect query parameter to the login / signup pages, like auth provider use as a common pattern. I'll do some more research and take a look into the current setup for auth navigation in another ticket.

tpatel:
Closing as the discussion has stopped, we can still reopen the PR in the future

PR #686, opened 2024-01-23
add youtube video player

postonsundays:
implementation can scale to add other hosted providers like Facebook / Tiktok, or use an outside library like React Player
https://github.com/cookpete/react-player


willydouhard:
Thank you! Can you provide a screenshot of the final result?

tpatel:
I really like the `react-player` idea, is this something you would like to pursue @postonsundays ?

PR #690, opened 2024-01-24
Added X-Chainlit-Session-Id header to uploadFile

JamesLloydPartingtonSerendipityai:
Adds the header to the upload file request


willydouhard:
This only fixes the issue partially for uploads. You will have the same issue for file reads and we can't pass a header easily on that request since we use the url directly in images tags and other stuff.

The real solution here is to implement sticky session with a cookie at the load balancer level

PR #702, opened 2024-01-27
Adds allow_origins config value to websocket cors

kvnn:
I have a Chainlit-based app set up with the following AWS setup:
- S3 static website using custom frontend
- load balancer
- ec2 instance(s) running docker container with the app

The S3 -> load balancer XML/HTTP requests work, using HTTPS, but the Websocket requests are failing because of CORS issues.

I'm looking to solve this for myself, and hopefully I can provide a simple and useful solution for the project as a whole.

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -183,7 +183,7 @@ app.add_middleware(
 
 socket = SocketManager(
     app,
-    cors_allowed_origins=[],
+    cors_allowed_origins=config.project.allow_origins,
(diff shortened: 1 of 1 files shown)


kvnn:
So, I just realized that this was just modified here: https://github.com/Chainlit/chainlit/pull/681/files
Why are you removing support for custom backends?

kvnn:
Closing this, I've dug way further into SocketIO / Starlette than I wanted to, I may re-open later with some more clarity

PR #708, opened 2024-01-30
adds custom_js

kvnn:
Adds a `custom_js` config option which will load the filename in the same manner as `custom_css`.

Tested on a personal project.

### DIFF
--- a/backend/chainlit/config.py
+++ b/backend/chainlit/config.py
@@ -101,6 +101,10 @@ hide_cot = false
 # The CSS file can be served from the public directory or via an external link.
 # custom_css = "/public/test.css"
 
+# Specify a Javascript file that can be used to customize the user interface.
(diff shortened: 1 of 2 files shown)


backend/chainlit/config.py
```diff
@@ -221,9 +226,9 @@ class CodeSettings:
     on_message: Optional[Callable[[str], Any]] = None
     author_rename: Optional[Callable[[str], str]] = None
     on_settings_update: Optional[Callable[[Dict[str, Any]], Any]] = None
-    set_chat_profiles: Optional[
-        Callable[[Optional["User"]], List["ChatProfile"]]
-    ] = None
+    set_chat_profiles: Optional[Callable[[Optional["User"]], List["ChatProfile"]]] = (
```
kvnn, line 229: This is from the auto-linter, FYI.

PR #735, opened 2024-02-09
update literalai dependency

tpatel:
update chainlit based on the future changes from the literalai dependency


backend/chainlit/playground/providers/openai.py
```diff
@@ -142,8 +142,8 @@ async def create_completion(self, request):
 
             llm_settings["stop"] = stop
 
-        if request.generation.functions:
-            llm_settings["functions"] = request.generation.functions
+        if request.generation.tools:
+            llm_settings["functions"] = request.generation.tools
```
willydouhard, line 146: we should set tools instead of function here

backend/chainlit/playground/providers/openai.py
```diff
@@ -309,8 +309,8 @@ async def create_completion(self, request):
 
         llm_settings["model"] = env_settings["deployment_name"]
 
-        if request.generation.functions:
-            llm_settings["functions"] = request.generation.functions
+        if request.generation.tools:
+            llm_settings["functions"] = request.generation.tools
```
willydouhard, line 313: same

backend/chainlit/langchain/callbacks.py
```diff
@@ -350,7 +213,14 @@ def _build_llm_settings(
         # make sure there is no api key specification
         settings = {k: v for k, v in merged.items() if not k.endswith("_api_key")}
 
-        return provider, settings
+        model_keys = ["model", "model_name", "deployment", "deployment_name"]
```
willydouhard, line 216: correct order is             model_keys = ["azure_deployment", "deployment_name", "model", "model_name"]

willydouhard (commented):
looks good, we can release the sdk version (should not break previous chainlit versions) to make sure this works end to end

PR #778, opened 2024-02-28
Add OpenAI integration

tpatel:
- This simplifies instrumenting OpenAI calls
- Reuses the literalai OpenAI instrumentation
- Creates a new step for each OpenAI call, with the call details in the generation property

Depends on https://github.com/Chainlit/python-client/pull/43

### DIFF
--- a/backend/chainlit/openai/__init__.py
+++ b/backend/chainlit/openai/__init__.py
@@ -27,9 +27,7 @@ def instrument_openai():
             parent_id = context.session.root_message.id
 
         step = Step(
-            name=generation.settings.get("model", generation.provider)
-            if generation.settings
-            else generation.provider,
+            name=generation.model if generation.model else generation.provider,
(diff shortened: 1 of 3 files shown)


backend/chainlit/openai/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+        context = get_context()
+
+        parent_id = None
+        if context.current_step:
+            parent_id = context.current_step.id
+        elif context.session.root_message:
+            parent_id = context.session.root_message.id
+
+        step = Step(
+            name=generation.settings.get("model", generation.provider)
```
willydouhard, line 30: generation now have a model field directly

willydouhard (changes requested):
LGTM, small comment on the generation.model

willydouhard (approved):

PR #781, opened 2024-02-29
Add custom tabs to FE

TomBers:
Hi All, First want to say excellent work on ChainLit.  

So this a PoC PR to see how to add tabs to the NavBar.  The use case for this would be people wanting a full site with ChainLit at its core.  Having the option to navigate to other full pages would deepen the customisation options.  (I am thinking about Pages such as About Us, or even fuller analytics view.

In `config.toml` it would look like:

```
injected_tabs = [
    { to = "/custom-tab", label = "Custom Tab" },
    { to = "/about", label = "About" }
]
```

### DIFF
--- a/backend/chainlit/config.py
+++ b/backend/chainlit/config.py
@@ -174,6 +174,12 @@ class Theme(DataClassJsonMixin):
     dark: Optional[Palette] = None
 
 
+@dataclass()
+class Tab(DataClassJsonMixin):
+    to: Optional[str] = None
+    label: Optional[str] = None
+
+
 @dataclass
 class SpeechToTextFeature:
     enabled: Optional[bool] = None
@@ -193,6 +199,7 @@ class FeaturesSettings(DataClassJsonMixin):
 class UISettings(DataClassJsonMixin):
     name: str
     show_readme_as_default: bool = True
+    injected_tabs: Optional[List[Tab]] = None
     description: str = ""
     hide_cot: bool = False
     # Large size content are by default collapsed for a cleaner ui
@@ -213,8 +220,10 @@ class CodeSettings:
     # Module object loaded from the module_name
     module: Any = None
     # Bunch of callbacks defined by the developer
-    password_auth_callback: Optional[Callable[[str, str], Optional["User"]]] = None
-    header_auth_callback: Optional[Callable[[Headers], Optional["User"]]] = None
+    password_auth_callback: Optional[Callable[[
+        str, str], Optional["User"]]] = None
+    header_auth_callback: Optional[Callable[[
+        Headers], Optional["User"]]] = None
     oauth_callback: Optional[
         Callable[[str, str, Dict[str, str], "User"], Optional["User"]]
     ] = None
@@ -309,7 +318,8 @@ def init_config(log=False):
                     translation = json.load(f)
                     with open(dst, "w", encoding="utf-8") as f:
                         json.dump(translation, f, indent=4)
-                        logger.info(f"Created default translation file at {dst}")
+                        logger.info(
+                            f"Created default translation file at {dst}")
 
 
 def load_module(target: str, force_refresh: bool = False):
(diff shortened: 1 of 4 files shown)


sandangel:
Hi, I'm super interested in this feature. May I ask can you add a screenshot for how it may look like? I want to keep all the components at the top and only change things below.

TomBers:
@sandangel So this is still work in progress, so it was meant just to see if anyone was open to the idea.

This is it running on my local version:

image

sandangel:
Hi @TomBers , that looks exactly what we need. Do you have an idea how can we build the UI for those injected tabs?

TomBers:
@sandangel So my idea would be that you just have the the control of the tabs.  There is already the ability to add new FastAPI routes to return a HTML document from the server.  

Personally I would just let the users link to pages and let them worry about the styling.  Open to opposing thoughts.

sandangel:
@TomBers is it possible to use client side router like react router? so the web still have the header displaying the tab you're on, while the content below can be different?

sandangel:
@TomBers I think chainlit support loading custom js file. Maybe we can set some elements' id for injected tabs, navigating by react-router. Then as users, we can build a new react app that mount to those elements. WDYT?

TomBers:
@sandangel Sorry for the late reply I was away for a few days.  

So quick thoughts - chainlit does support custom JS, I think it could be done like that, but not sure off the top of my head how to go about doing it, I would need to experiment.  The issue I see is that we are expanding the complexity, as in just adding a tab to a new page is relatively simple, then the user can have full control of the page and do whatever they want.  It maybe my bias, but React is quite the commitment if what I want is a few pages??

TomBers:
@sandangel I think it would be worth talking to other devs on the project as this is more of a directional decision than a technical one??

sandangel:
Yeah I think so. I hope this PR can get more attention from the project maintainer to see their opinion

TomBers:
This is my first PR to Chainlit, do you know someone in the main team who we could ask?

sandangel:
@willydouhard hi, can you add a comment here?

tpatel:
Hello, @TomBers and @sandangel. Thanks for suggesting this new feature with this draft PR 🙏 
I've talked to Willy about this and our thoughts are that it's going to far from the core focus of Chainlit.

We would like to keep Chainlit focused on the conversations and this might dilute our focus. Every new features will need maintenance in the future, and this feature opens the door for more website-oriented feature requests.

One good alternative is to use copilot with the custom build feature. This way you can host a totally custom site, and have chainlit as a widget.

Thanks for your understanding.

TomBers:
@tpatel @sandangel That makes sense, I appreciate starting the discussion, and can see it is potentially a maintenance issue.  

Just FYI @sandangel, while this was in discussion, I did go ahead an make an separate menu in pure JS. 
image

image


Thanks both.

sandangel:
Hi @TomBers, I'm interested with this feature and would greatly appreciate it if you can share the code.

TomBers:
@sandangel I put this in a custom JS file.

`
function openNav() {
    document.getElementById("myNav").style.width = "100%";
}

/* Close when someone clicks on the "x" symbol inside the overlay */
function closeNav() {
    document.getElementById("myNav").style.width = "0%";
}

function addTheMenu(addMenuToFn) {
    const header = document.querySelector('header');
    if (header) {
        addMenuToFn(header.nextElementSibling);
    } else {
        setTimeout(() => {
            addTheMenu(addMenuToFn);
        }, 200);
    }
}

function addMenuTo(element) {
    if (document.getElementById("myNav")) {
        return;
    }
    menuHtml =
        `<div id="myNav" class="overlay">

        
        <a href="javascript:void(0)" class="closebtn" onclick="closeNav()">&times;</a>
    
        
        <div class="overlay-content">
        <a href="/hello">About</a>
        <a href="#">Services</a>
        <a href="#">Clients</a>
        <a href="#">Contact</a>
        </div>
    
    </div>`;
    buttonHtml = `<span onclick="openNav()" class="menuBtn">☰ open</span>`;
    element.insertAdjacentHTML('beforebegin', buttonHtml);
    element.insertAdjacentHTML('beforebegin', menuHtml);
}


window.onload = function () {
    var observer = new MutationObserver(function (mutations) {
        // Run your function
        addTheMenu(addMenuTo);
    });

    var config = { attributes: true, childList: true, subtree: true };

    observer.observe(document.body, config);
}
`

sandangel:
Awesome. Thank you very much. I think this is a great starting point. We can go ahead and even build a react page that mount to the tab. THanks a lot for sharing.

tpatel:
Given it's not something we're gonna merge into chainlit, I'm closing the PR, thanks for your understanding.

PR #787, opened 2024-03-05
Multi-modal support for file size, mime types and number of files

jpolvto:
We've implemented support for restricting multi modal inputs by mime type, file size and number of files, similar to AskFileMessage, giving developers a bit more flexibility in choosing what inputs to allow for the application.

This will not return any error, but will restrict the user from uploading files that don't match the specifications.


tpatel (approved):

PR #808, opened 2024-03-12
Changed Azure AD Login Flow

shabirjan:
Added Hybrid Flow for Authorization grant to reterive user id_token. 
Changed Redirect from Get to Post as Hybrid Flow needs form_post instead of query and returns the response as Form body.

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -380,13 +380,14 @@ async def oauth_login(provider_id: str, request: Request):
     return response
 
 
-@app.get("/auth/oauth/{provider_id}/callback")
+@app.post("/auth/oauth/{provider_id}/callback")
 async def oauth_callback(
     provider_id: str,
     request: Request,
     error: Optional[str] = None,
-    code: Optional[str] = None,
-    state: Optional[str] = None,
+    code: Annotated[str, Form()] = None,
+    id_token: Annotated[str, Form()] = None,
+    state:Annotated[str, Form()] = None,
 ):
     if config.code.oauth_callback is None:
         raise HTTPException(
@@ -421,11 +422,11 @@ async def oauth_callback(
 
     # Check the state from the oauth provider against the browser cookie
     oauth_state = request.cookies.get("oauth_state")
-    if oauth_state != state:
-        raise HTTPException(
-            status_code=status.HTTP_401_UNAUTHORIZED,
-            detail="Unauthorized",
-        )
+    #if oauth_state != state:
+     #   raise HTTPException(
+      #      status_code=status.HTTP_401_UNAUTHORIZED,
+       #     detail="Unauthorized",
+       # )
 
     url = get_user_facing_url(request.url)
     token = await provider.get_token(code, url)
@@ -433,7 +434,7 @@ async def oauth_callback(
     (raw_user_data, default_user) = await provider.get_user_info(token)
 
     user = await config.code.oauth_callback(
-        provider_id, token, raw_user_data, default_user
+        provider_id, token, raw_user_data, default_user, id_token
     )
 
     if not user:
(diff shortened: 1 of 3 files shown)


backend/chainlit/server.py
```diff
@@ -380,13 +380,14 @@ async def oauth_login(provider_id: str, request: Request):
     return response
 
 
-@app.get("/auth/oauth/{provider_id}/callback")
+@app.post("/auth/oauth/{provider_id}/callback")
```
tpatel, line 383: This will break other oauth providers, I think you might want to add a `@app.post("/auth/oauth/azure-id/callback")` on top of the existing `@app.get("/auth/oauth/{provider_id}/callback")` route?

backend/chainlit/server.py
```diff
@@ -421,19 +422,19 @@ async def oauth_callback(
...
+       # )
 
     url = get_user_facing_url(request.url)
     token = await provider.get_token(code, url)
 
     (raw_user_data, default_user) = await provider.get_user_info(token)
 
     user = await config.code.oauth_callback(
-        provider_id, token, raw_user_data, default_user
+        provider_id, token, raw_user_data, default_user, id_token
```
tpatel, line 437: Could you update the signature of the function in https://github.com/Chainlit/chainlit/blob/3179d2928a4bafbc2d6e46b220ad44072753082c/backend/chainlit/__init__.py#L118 ?

tpatel (changes requested):

I've suggested some changes to make sure we don't break other oauth providers.

backend/chainlit/server.py
shabirjan: Thanks for suggestion, I have pushed the changes.

tpatel:
hello @shabirjan, I think there was a misunderstanding, in your code in `backend/chainlit/server.py`, you are changing the behavior for folks who are doing `GET /auth/oauth/{provider_id}/callback`, and this isn't something we can do.

Could you separate the code from `GET /auth/oauth/{provider_id}/callback` and `POST /auth/oauth/azure-ad/callback` to make sure we don't break other oauth providers?

shabirjan:
> hello @shabirjan, I think there was a misunderstanding, in your code in `backend/chainlit/server.py`, you are changing the behavior for folks who are doing `GET /auth/oauth/{provider_id}/callback`, and this isn't something we can do.
> 
> Could you separate the code from `GET /auth/oauth/{provider_id}/callback` and `POST /auth/oauth/azure-ad/callback` to make sure we don't break other oauth providers?

Hi @tpatel , Can you please explain a bit, what exactly you need me to do? What code should I separate?

tpatel:
hi @shabirjan, I've been working on this PR so that we can merge your changes. However I tried to look for the reason behind the PR and I could gather that you're interested by getting the user role?

Would this following snippet (without the code changes from this PR) work for your needs?

```
@cl.oauth_callback
def oauth_callback(
  provider_id: str,
  token: str,
  raw_user_data: Dict[str, str],
  default_user: cl.User,
) -> Optional[cl.User]:
  
  import httpx
  res = httpx.get("https://graph.microsoft.com/v1.0/me/appRoleAssignments", headers={"Authorization": f"Bearer {token}"})
  print(res.json()) # This contains the user "App Roles Assignments"

  return default_user
```

shabirjan:
hi @tpatel , We already discussed almost 2-3 month ago on the call, that we can't use that endpoint as it returns user roles for all the AAD Applications in the tenant for that specific user, which we are not looking for. Our use case for that implementation was check the role assigned to the currently logged in user on behalf of the Specific AAD App Registration.

tpatel:
closing in favor of https://github.com/Chainlit/chainlit/pull/1046 (I've implemented the azure hybrid flow as needed in a separate oauth provider and route so that it doesn't brake the integration for other users).

PR #813, opened 2024-03-14
move react-components code into react-client

tpatel:
- simplifies the dependencies and maintenance
- add cypress eslint configuration


tpatel:
cancelled in favor of #829

PR #822, opened 2024-03-18
Fix typo in provider name

giulioottantotto:
`Coginto` -> `Cognito`


tpatel (approved):

I just had to update your change because we've recently moved the file you had modified.

PR #836, opened 2024-03-21
Support SQLAlchemy for custom data layer

hayescode:
# Overview
- Adds custom, direct database, data layer using SQLAlchemy with support for a wide-range of SQL dialects
- Improved loading speeds of `list_threads` and `get_threads` dramatically versus previous recursion. 
  - The code is kinda wacky/long, but the end performance is worth it. 
- Optionally configures ADLS as the blob storage provider (designed to allow other providers in the future)
  - I only have access to Azure so I can't test/configure other providers
- Duplicated `PageInfo` and `PaginatedResponse` from literal SDK into `backend/chainlit/types.py` and updated typing.  
- _Note_: ~~I had to add locks because the backend uses `asyncio.gather` and sometimes `create_step` is attempted before `update_thread` causing a foreign key violation and race condition.~~
  - ~~I prefer this approach in order to maintain referential integrity of the database.~~
  - **April 2, 2024**: Removed most foreign key constraints because Chainlit backend executes tasks concurrently via `.gather()` causing `create_step` to be called before `update_thread` resulting in foreign key violations. I assume this was done for performance reasons. The backend would have to be re-worked to handle this and a custom solution adds complexity that should not exist.
- Added tests to new `cypress/e2e/custom_data_layer.py`.  

# Shout-Outs
- Thank you @nixent for suggesting to use SQLAlchemy for a broader solution to support a larger # of developers!
- Inspired from @sandangel who showed that this can be done and shared their work
- For testing I straight up ganked your `cypress` tests @tjroamer . I had never done these before.

# Test System
- **OS**: Windows 11
- **Python**: 3.11.4
- **Postgres**: v11

# How to configure
Install necessary dependencies to use this custom data layer.
```bash
pip install chainlit[custom-data] --upgrade
```   
Run this SQL DDL in your sql database
<details>
<summary> <b>DDL</b></summary>

```sql
CREATE TABLE users (
    "id" UUID PRIMARY KEY,
    "identifier" TEXT NOT NULL UNIQUE,
    "metadata" JSONB NOT NULL,
    "createdAt" TEXT
);

CREATE TABLE IF NOT EXISTS threads (
    "id" UUID PRIMARY KEY,
    "createdAt" TEXT,
    "name" TEXT,
    "userId" UUID,
    "userIdentifier" TEXT,
    "tags" TEXT[], 
    "metadata" JSONB,
    FOREIGN KEY ("userId") REFERENCES users("id") ON DELETE CASCADE
);

CREATE TABLE IF NOT EXISTS steps (
    "id" UUID PRIMARY KEY,
    "name" TEXT NOT NULL,
    "type" TEXT NOT NULL,
    "threadId" UUID NOT NULL,
    "parentId" UUID,
    "disableFeedback" BOOLEAN NOT NULL,
    "streaming" BOOLEAN NOT NULL,
    "waitForAnswer" BOOLEAN,
    "isError" BOOLEAN,
    "metadata" JSONB,
    "tags" TEXT[], 
    "input" TEXT,
    "output" TEXT,
    "createdAt" TEXT,
    "start" TEXT,
    "end" TEXT,
    "generation" JSONB,
    "showInput" TEXT,
    "language" TEXT,
    "indent" INT
);

CREATE TABLE IF NOT EXISTS elements (
    "id" UUID PRIMARY KEY,
    "threadId" UUID,
    "type" TEXT,
    "url" TEXT,
    "chainlitKey" TEXT,
    "name" TEXT NOT NULL,
    "display" TEXT,
    "objectKey" TEXT,
    "size" TEXT,
    "page" INT,
    "language" TEXT,
    "forId" UUID,
    "mime" TEXT
);

CREATE TABLE IF NOT EXISTS feedbacks (
    "id" UUID PRIMARY KEY,
    "forId" UUID NOT NULL,
    "value" INT NOT NULL,
    "comment" TEXT
);
```

</details>  

Add `SQLALCHEMY_CONNINFO` environment variable in your `.env` file  

```env
SQLALCHEMY_CONNINFO = <dialect>+<driver>://<user>:<password>:<port>/<database>
ADLS_SAS_TOKEN = <your_SAS_token> # Optional
```

Add this to your `app.py`
```python
import chainlit.data as cl_data
from chainlit.data.storage_clients import AzureStorageClient

storage_client = AzureStorageClient(
    account_url="https://<your_account>.dfs.core.windows.net",
    container="<your_container>",
    credential=credential,
    sas_token=ADLS_SAS_TOKEN
    )
cl_data._data_layer = SQLAlchemyDataLayer(
    conninfo=SQLALCHEMY_CONNINFO,
    ssl_require=True,
    storage_provider=storage_client,
    user_thread_limit=100)
```


backend/chainlit/sql_alchemy.py
```diff
@@ -0,0 +1,449 @@
...
+from chainlit.types import Feedback, FeedbackDict, Pagination, ThreadDict, ThreadFilter
+from literalai import PageInfo, PaginatedResponse
+from chainlit.step import StepDict
+
+if TYPE_CHECKING:
+    from chainlit.element import Element
+    from chainlit.step import StepDict
+
+class SQLAlchemyDataLayer(BaseDataLayer):
+    def __init__(self, conninfo, ssl_require=False):
```
nixent, line 27: Suggest using url for the parameter name as it is more recognizable 
```suggestion
    def __init__(self, url, ssl_require=False):
```

backend/chainlit/sql_alchemy.py
```diff
@@ -0,0 +1,449 @@
...
+from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy.ext.asyncio import create_async_engine, AsyncSession
+from sqlalchemy.orm import sessionmaker
+from azure.storage.filedatalake import FileSystemClient, ContentSettings  # type: ignore
+from chainlit.context import context
+from chainlit.logger import logger
+from chainlit.data import BaseDataLayer, queue_until_user_message
+from chainlit.user import User, PersistedUser, UserDict
+from chainlit.types import Feedback, FeedbackDict, Pagination, ThreadDict, ThreadFilter
+from literalai import PageInfo, PaginatedResponse
```
nixent, line 19: would be good to move `PageInfo` and `PaginatedResponse` definitions to chainlit

nixent (commented):
Thank you for turning to SQLAlchemy that fast! I have couple of comments, primarily about DB init instead of running DDL scripts

backend/chainlit/sql_alchemy.py
```diff
@@ -0,0 +1,449 @@
...
+        if ssl_require:
+            # Create an SSL context to require an SSL connection
+            ssl_context = ssl.create_default_context()
+            ssl_context.check_hostname = False
+            ssl_context.verify_mode = ssl.CERT_NONE
+            ssl_args['ssl'] = ssl_context
+        self.engine = create_async_engine(self._conninfo, connect_args=ssl_args)
+        self.async_session = sessionmaker(self.engine, expire_on_commit=False, class_=AsyncSession)
+        self.thread_update_lock = asyncio.Lock()
+        self.step_update_lock = asyncio.Lock()
```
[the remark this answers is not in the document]
hayescode: I expect this configuration documentation to be added to docs.chainlit.io once merged. I actually specifically left the DDL scripts out of the code change because this my DDL is Postgres specific and will not work with all dialects. I want this to be as widely accessible as possible.

backend/chainlit/sql_alchemy.py
```diff
@@ -0,0 +1,449 @@
...
+    async def create_user(self, user: User) -> Optional[PersistedUser]:
+        logger.info(f"Postgres: create_user, user_identifier={user.identifier}")
+        existing_user: Optional['PersistedUser'] = await self.get_user(user.identifier)
+        user_dict: Dict[str, Any]  = {
+            "identifier": str(user.identifier),
+            "metadata": json.dumps(user.metadata) or {}
+            }
+        if not existing_user: # create the user
+            logger.info("Postgres: create_user, creating the user")
+            user_dict['id'] = str(uuid.uuid4())
```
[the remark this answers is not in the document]
hayescode: I have no strong opinions on this. I'm not using Table objects but I could add easily enough. I believe python's `uuid` module follows the same standard. I haven't had an issue at least.

backend/chainlit/sql_alchemy.py
hayescode: I 100% agree!! @willydouhard @tpatel what do you think? Honestly I'm not clear why all dataclasses aren't in `chainlit/types.py`. Classes are duplicated in both Chainlit/Literal and they really belong in Chainlit as it actually handles theses classes. Currently I can't import `chainlit/element.py` or `chainlit/step.py` because of circular import errors with chainlit/data.

backend/chainlit/sql_alchemy.py
hayescode: I feel like this could be confusing for people constructing their own conninfo using its parts, one of which is url.

backend/chainlit/sql_alchemy.py
```diff
@@ -0,0 +1,449 @@
...
+
+        return PaginatedResponse(
+            pageInfo=PageInfo(hasNextPage=has_next_page, endCursor=end_cursor),
+            data=paginated_threads
+        )
+    
+    ###### Steps ######
+    @queue_until_user_message()
+    async def create_step(self, step_dict: 'StepDict'):
+        logger.info(f"Postgres: create_step, step_id={step_dict.get('id')}")
```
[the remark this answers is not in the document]
hayescode: I'll remove "Postgres" once I get more feedback. I like the idea of passing in the dialect to the class init, but would be better to automatically configure different dialects as well (instead of just updating logger messages).

backend/chainlit/sql_alchemy.py
willydouhard: We could duplicate the definitions if needed, but we need them in the literalai sdk also. For the circular import issue, did you check https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports ?

backend/chainlit/sql_alchemy.py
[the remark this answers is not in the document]
hayescode: @nixent updated

backend/chainlit/sql_alchemy.py
hayescode: @willydouhard @nixent @tpatel I copy/pasted the remaining literal definitions to `backend/chainlit/types.py`. Yes I'm type checking the other necessary types using your method. Testing shows this resolves the issue, thanks!

backend/chainlit/sql_alchemy.py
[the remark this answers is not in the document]
hayescode: literalai uses the python uuid version.
image

backend/chainlit/sql_alchemy.py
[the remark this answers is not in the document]
hayescode: If this becomes problematic, we can PR to do a CREATE IF NOT EXISTS upon instantiation.

backend/chainlit/data/sql_alchemy.py
```diff
@@ -0,0 +1,484 @@
...
+import aiofiles
+import aiohttp
+import asyncio
+from dataclasses import asdict
+from sqlalchemy import text
+from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy.ext.asyncio import create_async_engine, AsyncSession
+from sqlalchemy.orm import sessionmaker
+from azure.storage.filedatalake import FileSystemClient, ContentSettings  # type: ignore
+from literalai import PageInfo, PaginatedResponse
```
sandangel, line 15: update to import from chainlit

sandangel:
can we add an interface/adapter for blob storage client?
For example: the client needs to implement upload_file, delete_file, download_file. then user can just have their own client and implement that interface instead of having to make a PR to chainlit to support their blob storage choice?

sandangel:
I also suggest using memcached or session storage, but this is a good first starting point already.

backend/chainlit/data/sql_alchemy.py
hayescode: Tried but ChainlitDataLayer expects the literal version and mypy complains. I don't want to touch that class or literal. If there's another way please lmk!

hayescode:
> can we add an interface/adapter for blob storage client? For example: the client needs to implement upload_file, delete_file, download_file. then user can just have their own client and implement that interface instead of having to make a PR to chainlit to support their blob storage choice?

@sandangel I was thinking about make another file called `chainlit/backend/data/blob.py` or something where we could have classes for each blob storage provider. I only have access to Azure though, so cannot implement the others.

I added the functions based on the requirements from the [chainlit documentation](https://docs.chainlit.io/api-reference/data-persistence/custom-data-layer), except for delete_user_session because I cannot see a purpose. Where are you seeing delete_file and download_file?

sandangel:
@hayescode 

> Where are you seeing delete_file and download_file?

I think they are hidden in literalai API client. That is why I chose to implement the literalai API instead of BaseDataLayer

sandangel:
> I was thinking about make another file called chainlit/backend/data/blob.py or something where we could have classes for each blob storage provider. I only have access to Azure though, so cannot implement the others.

I think this does not scale well for the reason I mentioned earlier:

> having to make a PR to chainlit to support their blob storage choice

As the number of providers grow, it will add more maintenance burden for us. That is why I suggest having a common interface on the chainlit side, and on the users side they can use their own blob storage client.

hayescode:
> I think they are hidden in literalai API client.

I am not looking in literal because it's irrelevant for a custom data layer. Devs advise in the project ticket is to inherit from BaseDataLayer so that's what this implements.

> That is why I suggest having a common interface on the chainlit side, and on the users side they can use their own blob storage client.

@sandangel I don't understand what you mean I guess. If you have code for this please share.

In any case, this isn't a fully feature complete PR and we should expect (and welcome!) more enhancements. I can only test some SQL dialects and storage providers. Tbh I am in no rush as I'm already live with this. Data persistence has blocked me and my team for months so I made this out of frustration mostly haha.

sandangel:
@hayescode sorry for the confusion. Here is some pseudocode:

```python
# chainlit code

from typing import Protocol


class BlobStorageClient(Protocol):
    def upload_file(self, key: str, data: bytes) -> str:
        pass


class S3StorageClient:
    def upload_file(self, key: str, data: bytes) -> str:
        import boto3

        s3_client = boto3.client("s3")
        s3_client.put_object(Bucket="my-bucket", Key=key, Body=data)

        return f"s3://my-bucket/{key}"

class SQLAlchemyDataLayer(BaseDataLayer):

    async def add_blob_storage_client(self, blob_storage_client: BlobStorageClient) -> None:
        self.blob_storage_client = blob_storage_client
        logger.info("Blob Storage client initialized")


    @queue_until_user_message()
    async def create_element(self, element: 'Element'):
        # ...

        element.url = self.blob_storage_client.upload_file(key=element.key, data=element.data)


# user code:

cl._data_layer = SQLAlchemyDataLayer()
cl._data_layer.add_blob_storage_client(S3StorageClient())
```

hayescode:
@sandangel ok that's what i was thinking here

> I was thinking about make another file called chainlit/backend/data/blob.py or something where we could have classes for each blob storage provider. I only have access to Azure though, so cannot implement the others.

I have added this and made an Azure client and a S3 client (using your code). If you need to adjust the S3 client i suggest making a PR once this is merged.

I also added a `user_thread_limit` when initializing SQLAlchemyDataLayer to limit the number of threads returned by `get_all_user_threads` since this number will only grow so limiting this gives the developer the ability to customize `list_threads` and chat history performance.

hayescode:
Ok, this is ready. @willydouhard @tpatel It would be good to merge this before the Tuesday update (also i hate mypy, sorry for all the commits).

backend/chainlit/data/storage_clients.py
```diff
@@ -0,0 +1,58 @@
...
+
+class S3StorageClient(BaseStorageClient):
+    """
+    Class to enable Amazon S3 storage provider
+    """
+    def __init__(self, service_name: str, bucket: str):
+        try:
+            self.bucket = bucket
+            self.client = boto3.client(service_name)
+            logger.info("S3StorageClient initialized")
```
sandangel, line 47:
```suggestion
    def __init__(self, bucket: str):
        try:
            self.bucket = bucket
            self.client = boto3.client('s3')
            logger.info("S3StorageClient initialized")
```

sandangel: I think the service name is always s3

cypress/e2e/custom_data_layer/sql_alchemy.py
```diff
@@ -0,0 +1,28 @@
+from typing import List, Optional
+
+import chainlit.data as cl_data
+from chainlit.data.sql_alchemy import SQLAlchemyDataLayer
+from literalai.helper import utc_now
+
+import chainlit as cl
+
+cl_data._data_layer = SQLAlchemyDataLayer(conninfo="<your conninfo>")
```
sandangel, line 9: Shall we add blob storage client for testing?

tpatel:
> Ok, this is ready. @willydouhard @tpatel It would be good to merge this before the Tuesday update (also i hate mypy, sorry for all the commits).

hello @hayescode, thanks for your work on the custom data layer! There was a breaking change in literal that had to be pushed this chainlit release, could you update your PR?
I see you have copied some literal types into Chainlit, would there be a way to avoid it?

backend/chainlit/data/storage_clients.py
hayescode: I generally avoid hardcoding where possible. I suggest submitting a PR if passing "S3" as a parameter is a problem.

cypress/e2e/custom_data_layer/sql_alchemy.py
hayescode: done!

hayescode:
@tpatel I have updated this PR for Chainlit 1.0.500 compatibility. I have also updated the code to use the copied Chainlit versions of `PaginatedResponse` and `PageInfo`. These classes really belong in Chainlit and is an unnecessary risk for literal to be a dependency for a custom data layer. I added code [here](https://github.com/Chainlit/chainlit/pull/836/files#diff-1539dba35ca2b6051a16d07800529f285cc0046ec50a3af0a5fd49c3cb757a18R413-R426) to convert LiteralAI version to Chainlit version.

One odd difficulty that I've been battling, and ultimately gave up on, is the referential integrity of the underlying tables because of race conditions. For example, Chainlit sends `create_step` and `update_thread` concurrently and sometimes `create_step` is executed 1st. Steps should only exist if the thread exists so I initially made "threadId" a foreign key in the steps table. This race condition therefore caused SQL foreign key violations. It's possible to make this custom data layer await `update_thread`, but that wouldn't improve Literal. I suggest backend changes be made to handle this. I assume your DDL doesn't enfore foreign keys. Do you know why? Performance?

Please let us know when we can expect a review on this by. Thanks!

backend/chainlit/data/storage_clients.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self, account_url: str, container: str, credential: Optional[Union[str, Dict[str, str], "AzureNamedKeyCredential", "AzureSasCredential", "TokenCredential"]], sas_token: Optional[str] = None):
+        try:
+            self.data_lake_client = DataLakeServiceClient(account_url=account_url, credential=credential)
+            self.container_client: FileSystemClient = self.data_lake_client.get_file_system_client(file_system=container)
+            self.sas_token = sas_token
+            logger.info("AzureStorageClient initialized")
+        except Exception as e:
+            logger.warn(f"AzureStorageClient initialization error: {e}")
+        
+    async def upload_file(self, object_key: str, data: Union[bytes, str], mime: str = 'application/octet-stream', overwrite: bool = True) -> Dict[str, Any]:
```
sandangel, line 28:
```suggestion
    async def upload_file(self, object_key: str, data: Union[bytes, str], **kwargs: Any) -> Dict[str, Any]:
```

I think we should simplify the common signature to only common required params. Other params can be put in kwargs.

backend/chainlit/data/storage_clients.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    def __init__(self, service_name: str, bucket: str):
+        try:
+            self.bucket = bucket
+            self.client = boto3.client(service_name)
+            logger.info("S3StorageClient initialized")
+        except Exception as e:
+            logger.warn(f"S3StorageClient initialization error: {e}")
+
+    async def upload_file(self, object_key: str, data: Union[bytes, str], mime: str = 'application/octet-stream', overwrite: bool = True) -> Dict[str, Any]:
```
sandangel, line 51:
```suggestion
    async def upload_file(self, object_key: str, data: Union[bytes, str], **kwargs: Any) -> Dict[str, Any]:
```

backend/pyproject.toml
```diff
@@ -45,6 +45,11 @@ lazify = "^0.4.0"
 packaging = "^23.1"
 python-multipart = "^0.0.9"
 pyjwt = "^2.8.0"
+asyncpg = "^0.29.0"
+SQLAlchemy = "^2.0.28"
+boto3 = "^1.34.73"
+azure-identity = "^1.14.1"
+azure-storage-file-datalake = "^12.14.0"
```
sandangel, line 52: Should be optional deps

backend/chainlit/data/__init__.py
```diff
@@ -460,6 +469,10 @@ async def update_thread(
             tags=tags,
         )
 
+class BaseStorageClient(Protocol):
+    """Base class for non-text data persistence like Azure Data Lake, S3, Google Storage, etc."""
+    async def upload_file(self, object_key: str, data: Union[bytes, str], mime: str = 'application/octet-stream', overwrite: bool = True) -> Dict[str, Any]:
```
sandangel, line 474:
```suggestion
    async def upload_file(self, object_key: str, data: Union[bytes, str], **kwargs: Any) -> Dict[str, Any]:
```

backend/chainlit/data/storage_clients.py
hayescode: Priority is getting this into as many devs hands as possible who can test and improve on this in ways you and I cannot. You have enough ideas for a follow up PR, and I e already added several. I also plan to add OpenAI assistant thread/run/message synchronization in a future PR. Who knows what else the community will do to improve and extend this. That's why we need to get it in their hands.

backend/pyproject.toml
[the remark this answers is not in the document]
hayescode: Good idea.

- Moved to new poetry dependency called `custom-data`
- Updated `.github/workflows/mypy.yaml` to install these dependencies
- Updated the "How To Configure" section of my PR description

hayescode:
> Fantastic PR ✨
> 
> One blocker is to move the requirements to an optional dependency, so that we don't push these new requirements to all Chanlit deployments.
> 
> I'm gonna look into ways to have your test file run in the test suite, that would be a good way to bring more confidence in this change and in future changes.

Thanks @tpatel !!

Please let me know if I can assist in any way. For the test suite, maybe we just replicate [this test](https://github.com/hayescode/chainlit/blob/main/cypress/e2e/data_layer/main.py)?

What sort of timeline can we expect for this? It's growing cumbersome updating this PR for new main commits and fixing these pesky mypy issues 😊

sandangel:
Perhaps we should split the dependencies to s3 vs azure blog storage and import them only in the init function?

backend/chainlit/data/sql_alchemy.py
```diff
@@ -0,0 +1,482 @@
...
+        else: # update the user
+            logger.info("SQLAlchemy: update user metadata")
+            query = """UPDATE users SET "metadata" = :metadata WHERE "identifier" = :identifier"""
+            await self.execute_sql(query=query, parameters=user_dict) # We want to update the metadata
+        return await self.get_user(user.identifier)
+        
+    ###### Threads ######
+    async def get_thread_author(self, thread_id: str) -> str:
+        logger.info(f"SQLAlchemy: get_thread_author, thread_id={thread_id}")
+        query = """SELECT u.* FROM threads t JOIN users u ON t."user_id" = u."id" WHERE t."id" = :id"""
```
tpatel, line 117: I think it should be `userId` instead of `user_id` here, right?

backend/chainlit/data/sql_alchemy.py
```diff
@@ -0,0 +1,482 @@
...
+            FROM threads
+            WHERE "userId" = :user_id
+            ORDER BY "createdAt" DESC
+            LIMIT :limit
+        """
+        user_threads = await self.execute_sql(query=user_threads_query, parameters={"user_id": user_id, "limit": self.user_thread_limit})
+        if not isinstance(user_threads, list):
+            return None
+        thread_ids = "('" + "','".join(map(str, [thread['thread_id'] for thread in user_threads])) + "')"
+        if not thread_ids:
```
tpatel, line 359: This is always `True`, and it creates an error when there are no `user_threads` yet, you could check the length of `user_threads`.

backend/chainlit/data/storage_clients.py
```diff
@@ -0,0 +1,58 @@
...
+            return {"object_key": object_key, "url": url}
+        except Exception as e:
+            logger.warn(f"AzureStorageClient, upload_file error: {e}")
+            return {}
+
+class S3StorageClient(BaseStorageClient):
+    """
+    Class to enable Amazon S3 storage provider
+    """
+    def __init__(self, service_name: str, bucket: str):
```
tpatel, line 43: Isn't the service_name always `s3`?

tpatel (changes requested):
We can merge this as soon as it's ready!

I've got some errors when trying it locally:
- I had to use `postgresql+asyncpg://` in the connection string to make use of the `asyncpg` module (something that we'll need to add to the docs later)
- It seems like anonymous thread recording didn't work, there were errors about not having an existing user, although the steps were added to the DB
- I've commented inline for other errors.

backend/chainlit/data/sql_alchemy.py
hayescode: nice catch. I forgot to update `get_thread_author` after userIdentifier was added directly to threads in the latest update. 

Now we can simply do `query = """SELECT "userIdentifier" FROM threads "id" = :id"""`

Fixed

backend/chainlit/data/sql_alchemy.py
hayescode: fixed

backend/chainlit/data/storage_clients.py
hayescode: i have no idea, never used s3, but you're the 2nd person who has used it that wants to make it s3 so i did it.

Thanks for the review!

sandangel:
I think we should disable writing to DB at all if users are not signed in.

hayescode:
Yes, of course. I didn't think Chainlit allowed that [based on the docs](https://docs-prerelease.chainlit.io/data-persistence/history#how-to). Users can't even use my app unless authenticated. If it's true the backend logic doesn't enforce authentication for data persistence,how does Literal handle this? Foreign keys were acting as a guard (among other undesirable DB activity) but per my earlier comments race conditions caused me to remove that.

tpatel:
Literal does capture non-logged-in users queries too for observability purposes (so you can see anonymous feedback for example).

I think it's fine to capture data only for logged-in users, because the goal here is to keep the message history of logged-in users 👍

hayescode:
@tpatel interesting and news to me! I agree that this custom data layer should be limited to authenticated users. Do you have any guidance on how we should enforce authentication? Adding a `get_user` call to each function as assert authentication seems non-optimal so I wonder if there's a cleaner way you can think of?

hayescode:
@tpatel I ended up leveraging chainlit context by adding this to each of the functions to ensure DB writes only occur if a valid user is in context.

```python
if not getattr(context.session.user, 'id', None):
    raise ValueError("No authenticated user in context")
```

If you know of a better way please let us know. If not I believe all actions have been addressed?

tpatel:
@hayescode looks good, however I'm getting a `Chainlit context not found` because the `POST /project/threads` is called before the websocket session is established, so the context isn't defined yet.

I'm looking into how we could make this work.

hayescode:
@tpatel I removed the chainlit.context check from all except for create/upsert methods. This gives us the desired behavior where non-authenticated users cannot write to the DB, while maintaining functionality for authenticated users.

tpatel:
Neat, it makes a ton of sense to remove the readonly checks because they are called from already-authenticated routes (so they won't be called if the user isn't correctly authenticated) 👍 

Could we switch to a silent ignore (like `return None`) instead of raising an exception when `getattr(context.session.user, 'id', None)` is `None`? This would prevent folks who use Chainlit without authentication from having the exception stack traces in their logs.

backend/chainlit/data/sql_alchemy.py
```diff
@@ -0,0 +1,503 @@
...
+                "userId" AS user_id,
+                "userIdentifier" AS user_identifier,
+                "tags" AS thread_tags,
+                "metadata" AS thread_metadata
+            FROM threads
+            WHERE "userId" = :user_id
+            ORDER BY "createdAt" DESC
+            LIMIT :limit
+        """
+        user_threads = await self.execute_sql(query=user_threads_query, parameters={"user_id": user_id, "limit": self.user_thread_limit})
```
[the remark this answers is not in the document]
hayescode: @tpatel good idea! added!

backend/chainlit/data/sql_alchemy.py
```diff
@@ -0,0 +1,496 @@
...
+    async def delete_element(self, element_id: str):
+        logger.info(f"SQLAlchemy: delete_element, element_id={element_id}")
+        query = """DELETE FROM elements WHERE "id" = :id"""
+        parameters = {"id": element_id}
+        await self.execute_sql(query=query, parameters=parameters)
+
+    async def delete_user_session(self, id: str) -> bool:
+        return False # Not sure why documentation wants this
+
+    async def get_all_user_threads(self, user_id: Optional[str] = None, thread_id: Optional[str] = None) -> Optional[List[ThreadDict]]:
```
sandangel, line 346: I'm not sure how chainlit is using this function, but if they are using this for displaying the list of threads only, then I think we should not query all the information inside each thread. That should happen when user click to resume the chat.

hayescode: This function expects a List[ThreadDict] which is basically everything to your point. I was able to speed it up a lot by querying it all at once per user instead of recursively calling each thread, steps in that thread, etc to build it.

This is also used for the search and feedback.

I added a user_thread_limit to cap threads since it'll only grow. I have several users filling up my limit (100 threads) and performance is fine. Are you seeing performance problems or just asking the Chainlit devs why it's like this?

tpatel: This function isn't part of `BaseDataLayer` so it can be defined and used from within this custom data layer as needed.

@sandangel I agree that there might be some performance improvements to be done, I think it's fine for the first release of this custom data layer 👍

backend/chainlit/data/sql_alchemy.py
```diff
@@ -0,0 +1,496 @@
...
+        parameters = {"id": step_id}
+        await self.execute_sql(query=feedbacks_query, parameters=parameters)
+        await self.execute_sql(query=elements_query, parameters=parameters)
+        await self.execute_sql(query=steps_query, parameters=parameters)
+    
+    ###### Feedback ######
+    async def upsert_feedback(self, feedback: Feedback) -> str:
+        logger.info(f"SQLAlchemy: upsert_feedback, feedback_id={feedback.id}")
+        if not getattr(context.session.user, 'id', None):
+            raise ValueError("No authenticated user in context")
```
tpatel, line 256: You need to remove this check. The `upsert_feedback` is only called from a route, so it's outside of the user session (so attempting to use `context` will always through).

I don't have a good solution to prevent adding feedback to the DB for non-authenticated chainlit apps for now.

```suggestion
```

hayescode: Maybe we're overthinking this. I don't see a scenario where authenticated and non-authenticated users coexist in the same app/db. With authentication enabled users can't use my app with authentication. If someone doesn't have authentication set up I don't know why they'd set up a custom data layer.

I just don't see how this scenario we're trying to guard against would be possible, but maybe I'm missing something?

tpatel: Yes this is a edge edge case I agree. I can sadly think about a least one scenario where it could happen: a app owner disables the authentication for a few weeks.
To be fair this is the only thing that I've found in today's testing session, and getting this fixed would enable me to do a last code review / testing session.

An option that I've just though about is to use the `require_login` method instead of checking the context (https://github.com/Chainlit/chainlit/blob/59398777b7f04048ddd8961a47924a6a58c23be0/backend/chainlit/auth.py#L31-L37).

hayescode: I like this idea! Then we can just block initialzation of the custom data layer itself if no authentication. Great solution!

Where would we call require_login()? I've tried the app.py before the decorators and in sqlalchemy.py but it is returning `None` for me even when I do have authentication enabled.

```python
from chainlit.auth import require_login
is_authentication_enabled = require_login()

class SQLAlchemyDataLayer(BaseDataLayer):
    def __init__(self, conninfo: str, ssl_require: bool = False, storage_provider: Optional[BaseStorageClient] = None, user_thread_limit: Optional[int] = 1000):
        if not is_authentication_enabled:
            print(f'is_authentication_enabled: {is_authentication_enabled}')
            raise PermissionError("Authentication is required to use SQLAlchemyDataLayer")
```

intc-hharshtk:
Is it possible to continue from previous thread?

tpatel (approved):
✨✨✨ AWESOME ✨✨✨

Thanks for your work @hayescode ! I believe this feature is ready now, we can still disable the persistence and display a proper warning for chainlit apps without authentication in a future PR.

JeanRessouche:
Hi, it seems that this implementation does not support Azure Sql Databases: i get no error at boot but a Bad Request: 'User not persisted' 400 error, while the user is correctly stored in the user table. Also, on thread creation, an error appears to store thread.tags because this datatype (TEXT[]) does not exist in Azure Sql. 

```py
from sqlalchemy import create_engine
import urllib
import chainlit.data as cl_data
from sqlalchemy.engine import URL

driver = 'ODBC Driver 17 for SQL Server'

params = urllib.parse.quote_plus(
    'Driver=%s;' % driver +
    'Server=tcp:%s,1433;' % os.environ["AZURE_DB_HOST"] +
    'Database=%s;' % os.environ["AZURE_DB_DATABASE"] +
    'Uid=%s;' % os.environ["AZURE_DB_USERNAME"] +
    'Pwd={%s};' % os.environ["AZURE_DB_PASSWORD"] +
    'Persist Security Info=False;' +
    'MultipleActiveResultSets=False;' +
    'Encrypt=yes;' +
    'TrustServerCertificate=no;' +
    'Connection Timeout=30;')

conn_str = 'mssql+pyodbc:///?odbc_connect={}'.format(params)
engine_azure = create_engine(conn_str)
engine_azure.connect()
print('Connection to the Azure Db is ok')

connection_url = URL.create(
    drivername='mssql+aioodbc',
    username=os.environ["AZURE_DB_USERNAME"],
    password=os.environ["AZURE_DB_PASSWORD"],
    host=os.environ["AZURE_DB_HOST"],
    database=os.environ["AZURE_DB_DATABASE"],
    query={
        'driver': driver,
        'Encrypt': 'yes',
        'TrustServerCertificate': 'no',
        'Connection Timeout': '30'
    }
)

cl_data._data_layer = SQLAlchemyDataLayer(
    conninfo=connection_url,
    ssl_require=True,
    storage_provider=engine_azure,
    user_thread_limit=100)
```

Did i made a mistake somewhere ?

tpatel:
@JeanRessouche I've added [some docs yesterday](https://docs.chainlit.io/data-persistence/custom#sql-alchemy-data-layer) to clarify the scope of this release. For now this has only been tested with PostgreSQL.

I'll be happy to guide you if you choose to add support for Azure SQL!

JeanRessouche:
Thanks a lot @tpatel, the doc is helping but it's still a little bit cloudy for me.

Adding support for Azure Sql is definitely something that i'm willing to do, but not in the short term, so I'm trying the Azure Datalake way. 

Based on the doc, i'm facing a blocker with the conninfo variable.
For me an Azure Datalake Gen 2 is basically a storage account (that can provide with the account name & sas token), so i have no idea how to get the conninfo content that look like sql server details here. Does it require an Azure Synapse on top of the Data lake gen 2 ?

hayescode:
For the tags I'm not sure what that azure SQL equivalent is but it's a list of strings in that column.

For the data lake I am using ADLS gen2. User not persisted sounds like this isn't getting configured properly. Pass the account url and credential (managed identity or access key). Try a test script to write some test data.

JeanRessouche:
I'm curious about how you created the DDL in ADLS gen 2 on your side, i was only able to connect to it with sql server driver, thus the postgre ddl won't work.

So far i was only able to make it work properly with a postgre server (which is great already!), failed with ADSL gen2 & Azure Sql database.

JeanRessouche:
Hum, ok, i have to retry then, did that but it wasn't working, certainly because i did not figure out what to put in conninfo when we use ADLS.

hayescode:
you don't put it in the conninfo you instantiate the ADLS class and pass it `storage_provider`.

```python
from chainlit.data.storage_clients import AzureStorageClient

storage_client = AzureStorageClient(
    account_url="https://<your_account>.dfs.core.windows.net",
    container="<your_container>",
    credential=credential,
    sas_token=ADLS_SAS_TOKEN
    )
cl_data._data_layer = SQLAlchemyDataLayer(
    conninfo=SQLALCHEMY_CONNINFO,
    ssl_require=True,
    storage_provider=storage_client,
    user_thread_limit=100)
```

hayescode:
There are 2 clients. Pass the ALDS client to the SQLAlchemy via storage_client. I think you're conflating these.

Reymond190:
In next commits, pls fix this ->
i tested the sql alchemy datalayer using azure AD Oauth. there is an validation error when whole dict is used in "PersistedUser(**user_data)", works after metadata is removed from the dict.


 ERROR:
  File "D:\git_projects\mvenv_504\lib\site-packages\chainlit\data\sql_alchemy.py", line 114, in get_user
    return PersistedUser(**user_data)
  File "D:\git_projects\mvenv_504\lib\site-packages\pydantic\_internal\_dataclasses.py", line 140, in __init__
    s.__pydantic_validator__.validate_python(ArgsKwargs(args, kwargs), self_instance=s)
pydantic_core._pydantic_core.ValidationError: 1 validation error for PersistedUser
metadata
  Input should be a valid dictionary [type=dict_type, input_value='{"image": "data:image/jp..."provider": "azure-ad"}', input_type=str]
    For further information visit https://errors.pydantic.dev/2.7/v/dict_type

Tirupathiraopatnala:
> There are 2 clients. Pass the ALDS client to the SQLAlchemy via storage_client. I think you're conflating these.

This is a bit confusing tbh, can you please add detailed steps to setup data persistance with azure

Tirupathiraopatnala:
> There are 2 clients. Pass the ALDS client to the SQLAlchemy via storage_client. I think you're conflating these.

from chainlit.data.storage_clients import AzureStorageClient

storage_client = AzureStorageClient(
    account_url="https://<your_account>.dfs.core.windows.net",
    container="<your_container>",
    credential=credential,
    sas_token=ADLS_SAS_TOKEN
    )
    
 If only this part of the code is responsible for data persistance in azure, will it create the required tables by default? or do we need to spin a db there?

hayescode:
Data lake only stores elements (non-text). It's optional. SQL database handles everything else and the DDL is in the pr description.

Issue #837, opened 2024-03-21
react-client: Provide clean way to get the current thread id


jbeckerdm:
**Is your feature request related to a problem? Please describe.**
We're using the react-client in our own custom frontend for a while now and have generally been very happy with it. It's great that it exposes all of chainlit's functionality in a very well thought-out and easy to use manner. Even things it doesn't expose can mostly be accessed anyway by just accessing the recoil state ourselves.
However, we did not see any way to access the thread id of the currently active chat. Yes, there's `idToResume` but it's only set when we're actually resuming a thread, but not if we created a new chat and started chatting in it.

We hacked together our own way for now, a hook that reads the thread id from the first response of a new chat, so that we have access to it as soon as possible, or return `idToResume` if that one is set. But it would be great to have react-client expose it itself

**Describe the solution you'd like**
react-client should expose a single `threadId` that equals 
- the threadId of a just started chat as soon as the thread has been created in the backend
- the threadId of the resumed chat while a chat is currently being resumed (=idToResume)
- undefined when the chat session is 'empty', aka nothing has been written yet (or clear() has been called right before)

This could take idToResume's place as well - if the user needs to know if the current chat is a resumed one, maybe expose it as a boolean `isResuming`.

**Describe alternatives you've considered**
Nothing yet. As mentioned, we're currently using a hacky workaround to get such information, and having a single state that holds the current chat's threadId whenever there is one has been very helpful. I think it would benefit other users of the react-client as well to simplify this.

Issue #824, opened 2024-03-18
New thread not in the Past Chats until the page is refresshed


BrittleFoot:
**Describe the bug**

**To Reproduce**
1. Create a new thread, add messages
2. Observe the past chats side view
3. **See that this new thread is not there**
4. Refresh the page
5. **See that created thread is there**

**Expected behavior**
Thread is shown in PastChats as soon as it's been created

**Screenshots**

> Ignore the error responses, I just disabled the model for cost's sake

There is no thread in the side view
![image](https://github.com/Chainlit/chainlit/assets/14301123/f43df8dc-e8e7-49b5-b7e7-5a745c05ffcd)

Thread is here, after the refresh
![image](https://github.com/Chainlit/chainlit/assets/14301123/d4fb7abc-32f3-4089-90d9-9a0d545fdc5e)

**Desktop (please complete the following information):**

OS: both windows and linux clients, server on linux
Browser chrome
Version 1.0.400

**Additional context**

I have my custom data layer, but all methods works fine afaik :)

Issue #823, opened 2024-03-18
New messages does not shown in past chat until the page is refreshed


BrittleFoot:
**Describe the bug**
Hi! I am using the persistence data layer and I noticed somewhat unclear UI behaviour

When I resume a thread and add a few messages, then switch to the same thread from the Past Chats view, the new messages I added are not shown.

From the network log, it may be due to threads being cached client-side. 

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a new thread with some messages
2. Refresh the page so this new thread will appear in the PastChats
    > _strange behaviour too, I expect a new chat to appear there immediately_ 
3. Select the thread from Past Chats
4. Click "Resume chat"
5. Add new messages
6. Select other thread and Select this thread in Past Chats again
7. **See that new messages didn't show there**
8. Refresh the page
9. Select this thread in Past Chats again
10. **See that new messages from step 5 are there**

**Expected behaviour**
I expect to see the updated version of the thread in the past messages.

**Screenshots**

> ignore the error responses, I just disabled the model for cost's sake
 
New Chat
![image](https://github.com/Chainlit/chainlit/assets/14301123/768e3ca4-8644-400f-8857-ed96330f6457)

See see it in past chats
![image](https://github.com/Chainlit/chainlit/assets/14301123/f043105f-0eac-4d2a-aaf1-ca324ad08b5f)

Rasume and add new messages
![image](https://github.com/Chainlit/chainlit/assets/14301123/a71148d0-1dfd-43d1-b5a1-e00708d3df17)

Click to other chat and click back to this chat
**Notice there are no** `message2`
![image](https://github.com/Chainlit/chainlit/assets/14301123/5ce48cb0-4d93-45f9-a744-3643489d3e24)

Refreshed the page
![image](https://github.com/Chainlit/chainlit/assets/14301123/777b943d-d60e-4bce-ac8d-eb3160c5ca34)

**Desktop (please complete the following information):**
 - OS: both windows and linux clients, server on linux
 - Browser chrome
 - Version 1.0.400

**Additional context**

I have my custom data layer, but all methods works fine afaik :)

PR #838, opened 2024-03-21
Newly created thread displayed in the side bar; Resumed threads content is revalidated when navigating in the side bar

BrittleFoot:
Hi!

I recently posted two issues:
- https://github.com/Chainlit/chainlit/issues/824
- https://github.com/Chainlit/chainlit/issues/823

Here's my proposal for solving it

There are two main ideas:

- when the new thread is created we refresh the list of the threads
  > I detect it with a combination of `firstInteraction` and `messages` states, since, as mentioned in https://github.com/Chainlit/chainlit/issues/837 there is no clear way to determine a newly created thread
- when the thread is resumed - we set the current thread page data as '' with `mutate()` so it'll be invalidated on the next render.
- also, when the already resumed/opened thread is clicked in the thread history sidebar the user is redirected to this opened chat, instead of a suggestion to resume the thread again

Here's a demo of how it works

https://github.com/Chainlit/chainlit/assets/14301123/99f90677-5854-4925-b815-30c6e0dedde4

### DIFF
--- a/frontend/src/components/organisms/threadHistory/sidebar/ThreadList.tsx
+++ b/frontend/src/components/organisms/threadHistory/sidebar/ThreadList.tsx
@@ -17,6 +17,7 @@ import Typography from '@mui/material/Typography';
 import {
   ThreadHistory,
   useChatInteract,
+  useChatMessages,
   useChatSession
 } from '@chainlit/react-client';
 
@@ -41,6 +42,7 @@ const ThreadList = ({
 }: Props) => {
   const { idToResume } = useChatSession();
   const { clear } = useChatInteract();
+  const { messages } = useChatMessages();
   const navigate = useNavigate();
   if (isFetching || (!threadHistory?.timeGroupedThreads && isLoadingMore)) {
     return (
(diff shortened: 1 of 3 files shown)


tpatel:
Hello, I've tried to rebase to test these changes but made a mistake and it close the PR, I'm working on fixing it

tpatel:
@BrittleFoot Would you mind adding me temporarily as a contributor to your fork? I believe when I updated your code it failed in the middle, close the PR which cut my access to your repo 😓 

After some testing your code has a lot of potential, I just had an issue with the thread list that is reloaded more often than it needs, which sometimes sets the `currentThreadId` to the wrong thread.

tpatel:
After looking into your work, we decided to go with an alternative implementation, where the current chat is moved to `/thread/:id` after being persisted, instead of staying at `/`. The current PR is located at https://github.com/Chainlit/chainlit/pull/923

BrittleFoot:
Works for me, thanks! More robust, even.

my pleasure contributing to it

PR #842, opened 2024-03-22
Do not reload installed packages on file change

### LINKED ISSUE #790: Investigate and fix file watcher issues
Related issues:

- https://github.com/Chainlit/chainlit/issues/655
- https://github.com/Chainlit/chainlit/issues/745
- https://github.com/Chainlit/chainlit/issues/784

rickythefox:
Closes #790 .

The issue was present when the venv was created in the same dir as the project using chainlit. On reload the code that was supposed to clear the modules related to the app from `sys.modules` also cleared chainlit itself which led to a reload and a new config object being initialized when `config.py` module was loaded. The `socket.py` module still had the reference to the old config module though, which resulted in reload failing. The lesson here is - module reloading in python is HARD and should really not be attempted. ;)

The fix checks where the modules are installed (venv or regular installation) and prevents their reload. This fixes all cases EXCEPT the `hello.py` in the chainlit repo - as the chainlit code is in the same dir as the demo file it's imporsible to say what's what. A possible fix would be to move `hello.py` to a subdir.

### DIFF
--- a/backend/chainlit/config.py
+++ b/backend/chainlit/config.py
@@ -1,5 +1,6 @@
 import json
 import os
+import site
 import sys
 from importlib import util
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


willydouhard:
This looks awesome! Going to do a bit of manual testing before merging. Thanks!

rickythefox:
So did you get this to work? Any corner cases I need to take into account?

tpatel (approved):

I couldn't replicate the file watcher issue, so I can't be sure it's fixing all the issues mentioned, but the explanation makes sense and I see no bad things that could come form this!

PR #850, opened 2024-03-27
fix BaseDataLayer create_element type

wapa5pow:
There is argument type inconsistency between BaseDataLayer.create_element and ChainlitDataLayer.create_element.
I checked the type running on my local environment and confirmed that it is Element type.


tpatel (approved):

PR #851, opened 2024-03-27
migrate to literal score

willydouhard:


backend/chainlit/types.py
```diff
@@ -33,8 +33,8 @@ class Pagination(BaseModel):
 
 
 class ThreadFilter(BaseModel):
-    feedback: Optional[Literal[-1, 0, 1]] = None
-    userIdentifier: Optional[str] = None
+    feedback: Optional[Literal[0, 1]] = None
+    userId: Optional[str] = None
```
hayescode, line 37: @willydouhard why is `-1` being removed?

backend/chainlit/types.py
```diff
@@ -146,16 +149,16 @@ class ChatProfile(DataClassJsonMixin):
...
 
 
 @dataclass
 class Feedback:
     forId: str
-    value: Literal[-1, 0, 1]
-    strategy: FeedbackStrategy = "BINARY"
+    value: Literal[0, 1]
     id: Optional[str] = None
     comment: Optional[str] = None
```
hayescode, line 163: @willydouhard why are these changing?

backend/chainlit/types.py
willydouhard: -1 becomes 0 and we introduce the concept of "feedback deletion" to replace the current 0 purpose. Having 3 values was a workaround to avoid to handle feedback deletion

backend/chainlit/types.py
willydouhard: Chainlit is only supporting binary feedback so this was never used throughout the project.

PR #861, opened 2024-04-02
modify config.py

raihanou1:
We are a group of students conducting a security audit on an application named TaskWeaver, which utilizes Chainlit, as part of our final year project. During our analysis, we discovered that the default host configuration is set to 0.0.0.0. 

To enhace the application's security, the default host configuration should be altered to 127.0.0.1. The current setting, 0.0.0.0, permits connections from any external IP address, thereby widening the attack surface. This could lead to several security vulnerabilities, such as:

- Denial of Service (DoS) Attacks: Attackers can inundate the system with an overwhelming number of requests, leading to service interruptions for legitimate users by exhausting the system's resources.
- Man-in-the-Middle (MitM) Attacks: The open access makes it feasible for attackers to intercept and manipulate communications between two parties covertly.

By changing the default host to 127.0.0.1, we limit connections exclusively to the local machine. This adjustment drastically minimizes these security risks and reinforces the application's defenses against potential cyber threats.


dokterbob:
This is a welcome and necessary security patch!

... but also a breaking change for some users! 

Would you be so kind as to add this as a breaking change in the [changelog's unreleased section](https://github.com/Chainlit/chainlit/blob/df334f2be15c9d5e57b9c854cc9a5f291bcc766d/CHANGELOG.md?plain=1#L7)?

dokterbob (changes requested):
Looks good to me. Please update the changelog and we'll merge!

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 Nothing unreleased!
 
+## [1.1.404] - 2024-08-22
+
+### Changed
+
+- changing the default host from 0.0.0.0 to 127.0.0.1  
+
```
dokterbob, line 16: Needs to be in `## Unreleased` and we're good to go!

dokterbob (changes requested):

dokterbob (approved):

PR #870, opened 2024-04-02
rename the Literal env variable to LITERAL_API_URL

tpatel:


backend/chainlit/data/__init__.py
```diff
@@ -462,7 +462,7 @@ async def update_thread(
 
 
 if api_key := os.environ.get("LITERAL_API_KEY"):
-    server = os.environ.get("LITERAL_SERVER")
+    server = os.environ.get("LITERAL_API_URL")
```
willydouhard, line 465: we should be backward compatible with LITERAL_SERVER also

willydouhard (changes requested):

backend/chainlit/data/__init__.py
tpatel: good point, this would prevent a breaking change 👍

tpatel: fixed

willydouhard (approved):

PR #876, opened 2024-04-03
add model and chunks to llama index callback handler

willydouhard:


backend/chainlit/llama_index/callbacks.py
```diff
@@ -152,10 +171,15 @@ def on_event_end(
             step.output = content
 
             token_count = self.total_llm_token_count or None
-
+            raw_response = response.raw if response else None
+            model = raw_response.get("model", None) if raw_response else None
+            provider = "openai"
```
tpatel, line 176: Is it always `openai`?

tpatel: I couldn't find a good way to extract the correct provider, so let's merge it like this.

tpatel (approved):

PR #882, opened 2024-04-04
feat: allow using UVICORN_LOG_CONFIG env var to be used to configure uvicorn logging.

LasseGravesenSaxo:
# Motivation
The reason for this PR is that I'm having issues configuring all the loggers, and this causes problems with some log messages being in a different format than expected; We want all our logging to be json such that it can be ingested into our logging aggregator as a single blob, but with the default logging config that cannot be overridden currently we instead sometimes get a bunch of separate log messages that can't be ingested as a single blob and are missing important fields.

# Implementation
We let `UVICORN_LOG_CONFIG` be loaded from the environment, its default value is `uvicorn.config.LOGGING_CONFIG` (unsure how to handle this best really, either we need a reference to `uvicorn.config.LOGGING_CONFIG` or it should instead not set the `log_config` kwarg in the `uvicorn.Config` if it's not set, which would then fallback to the `uvicorn` default value; because if we set it to None then it won't use the default logging config). This config value is then passed to `uvicorn.Config` which is used to configure the `uvicorn` server. 

The value of `UVICORN_LOG_CONFIG` could be `logging.conf`, and `logging.conf` could look something like this:

```
[loggers]
keys = uvicorn

[logger_uvicorn]
level = DEBUG
handlers = custom
propagate = 0
qualname = uvicorn

[handlers]
keys = custom

[handler_custom]
class = StreamHandler
level = DEBUG
formatter = json
args = (sys.stdout,)

[formatters]
keys = json

[formatter_json]
class = ...
```

# Questions
1) It might be that this PR should also modify this file: https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/logger.py

To ensure that logging isn't configured multiple times.

2) It might be that there are other uvicorn configs that should passthru like this. I don't know if we can start uvicorn in such a way that it respects all environment variables without just duplicating what they've done here: https://github.com/encode/uvicorn/blob/master/uvicorn/main.py

### DIFF
--- a/backend/chainlit/cli/__init__.py
+++ b/backend/chainlit/cli/__init__.py
@@ -55,8 +55,6 @@ def run_chainlit(target: str):
     ws_protocol = os.environ.get(
         "UVICORN_WS_PROTOCOL", "auto"
     )
-    log_config_env = os.environ.get("LOG_CONFIG", uvicorn.config.LOGGING_CONFIG)
-
     config.run.host = host
     config.run.port = port
     config.run.root_path = root_path
@@ -75,6 +73,7 @@ def run_chainlit(target: str):
     init_lc_cache()
 
     log_level = "debug" if config.run.debug else "error"
+    log_config = config.run.log_config
 
     # Start the server
     async def start():
@@ -83,7 +82,7 @@ def run_chainlit(target: str):
             host=host,
             port=port,
             ws=ws_protocol,
-            log_config=log_config_env,
+            log_config=log_config,
             log_level=log_level,
             ws_per_message_deflate=ws_per_message_deflate,
             ssl_keyfile=ssl_keyfile,
@@ -127,7 +126,7 @@ def run_chainlit(target: str):
 @click.option(
     "-l",
     "--log-config",
-    default=None,
+    default=uvicorn.config.LOGGING_CONFIG,
     is_flag=False,
     envvar="LOG_CONFIG",
     help="Set the Uvicorn log config file path. Supported formats: .ini, .json, .yaml.",
@@ -189,8 +188,6 @@ def chainlit_run(
         os.environ["CHAINLIT_SSL_KEY"] = ssl_key
     if root_path:
         os.environ["CHAINLIT_ROOT_PATH"] = root_path
-    if log_config:
-        os.environ["LOG_CONFIG"] = log_config
     if ci:
         logger.info("Running in CI mode")
 
@@ -210,6 +207,7 @@ def chainlit_run(
     config.run.watch = watch
     config.run.ssl_cert = ssl_cert
     config.run.ssl_key = ssl_key
+    config.run.log_config = log_config
 
     run_chainlit(target)
 
--- a/backend/chainlit/cli/__init__.py
+++ b/backend/chainlit/cli/__init__.py
@@ -55,6 +55,7 @@ def run_chainlit(target: str):
     ws_protocol = os.environ.get(
         "UVICORN_WS_PROTOCOL", "auto"
     )
+    uvicorn_log_config_env = os.environ.get("UVICORN_LOG_CONFIG", uvicorn.config.LOGGING_CONFIG)
 
     config.run.host = host
     config.run.port = port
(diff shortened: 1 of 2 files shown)


LasseGravesenSaxo:
Hope someone can take a look at this!

smueller18:
We are also interested in having an option to customize the Uvicorn logging. @LasseGravesenSaxo why did you choose a plain environment variable? For me, it would make a lot more sense to introduce a new CLI option.

Like this:

```python
@click.option(
    "-l",
    "--log-config",
    envvar="LOG_LEVEL",
    help="Uvicorn logging configuration file. Supported formats: .ini, .json, .yaml.",
)
def chainlit_run(target, watch, headless, debug, ci, no_cache, host, port, log_config):
    pass
```

If the option is not given, the default value of Uvicorn will be used anyway, so no need to define a custom default here.

LasseGravesenSaxo:
> We are also interested in having an option to customize the Uvicorn logging. @LasseGravesenSaxo why did you choose a plain environment variable? For me, it would make a lot more sense to introduce a new CLI option.
> 
> Like this:
> 
> ```python
> @click.option(
>     "-l",
>     "--log-config",
>     envvar="LOG_LEVEL",
>     help="Uvicorn logging configuration file. Supported formats: .ini, .json, .yaml.",
> )
> def chainlit_run(target, watch, headless, debug, ci, no_cache, host, port, log_config):
>     pass
> ```
> 
> If the option is not given, the default value of Uvicorn will be used anyway, so no need to define a custom default here.

@smueller18 LOG_LEVEL and --log-config should be different.

I picked an environment variable because of this bit already using an environment variable for Uvicorn:

```
    ws_per_message_deflate_env = os.environ.get(
        "UVICORN_WS_PER_MESSAGE_DEFLATE", "true"
    )
    ws_per_message_deflate = ws_per_message_deflate_env.lower() in [
        "true",
        "1",
        "yes",
    ]  # Convert to boolean
```

But I can easily modify it such that it uses an option too.

LasseGravesenSaxo:
@smueller18, I made a change now that makes it possible to set log config as a click option.

backend/chainlit/cli/__init__.py
```diff
@@ -128,19 +139,21 @@ async def start():
 )
 @click.option("--host", help="Specify a different host to run the server on")
 @click.option("--port", help="Specify a different port to run the server on")
-def chainlit_run(target, watch, headless, debug, ci, no_cache, host, port):
+def chainlit_run(target, watch, headless, debug, log_config, ci, no_cache, host, port):
     if host:
         os.environ["CHAINLIT_HOST"] = host
     if port:
         os.environ["CHAINLIT_PORT"] = port
+    if log_config:
```
smueller18, line 147: This makes not really sense here. Instead, the log_config should be added to class `RunSettings` and be set below:

```
config.run.debug = debug
config.run.log_config = log_config
```

backend/chainlit/cli/__init__.py
```diff
@@ -47,6 +47,8 @@ def run_chainlit(target: str):
         "yes",
     ]  # Convert to boolean
 
+    log_config_env = os.environ.get("LOG_CONFIG", uvicorn.config.LOGGING_CONFIG)
```
smueller18, line 50: This should be removed. Instead use `log_config=config.run.log_config` below:

```
        config = uvicorn.Config(
            app,
            host=host,
            port=port,
            log_config=config.run.log_config,
            log_level=log_level,
            ws_per_message_deflate=ws_per_message_deflate,
        )
```

LasseGravesenSaxo:
@smueller18 I applied changes based on your comments.

Atheuz:
Interested in seeing this get attention from maintainers as it's causing an issue with logging for us.

dokterbob:
Hey all, we're finally getting to give love to some of these long-pending contribs.

As in the meantime, we've landed the possibility to mount chainlit in FastAPI apps, so that we can run and configure our own uvicorn, is this still a relevant and necessary feature?

I'm asking as I'd imagine that anyone running chainlit in production will want to do so with their own uvicorn, where they have full control over it's config.

This way, we won't end up having to maintain wrappers for the plethora of uvicorn options without blocking devs accesss to lower-level stuff. The same goes for FastAPI. 

In this perspective, the chainlit CLI is merely a wrapper to land quick PoC's and a full-fledged ASGI would be the production-grade wrapper for scale-out (similar to how e.g. Django does it).

Love to hear your feedback @LasseGravesenSaxo and @smueller18.

dokterbob:
@smueller18 Thanks for the head's up. I will close this one for now.

LasseGravesenSaxo:
We are good also now using what smueller18 said.

Issue #862, opened 2024-04-02
i get error result for show Markdown Table


wac81:
I want to show html form
The table code is shown below:
[...]

PR #893, opened 2024-04-10
Allow html in text elements

### LINKED ISSUE #597: Allow Unsafe Html does NOT impact Text Element
I would have expected the config option `allow_unsafe_html` to affect the content in `Text` element as well. However, it seems to only permit raw markdown. It would be great if the `Text` element could respect this config option.

jdb78:
This fixes #597 and #862 

Text elements can now use html to render content vs just messages.

### DIFF
--- a/frontend/src/components/atoms/elements/Text.tsx
+++ b/frontend/src/components/atoms/elements/Text.tsx
@@ -1,3 +1,5 @@
+import { useRecoilValue } from 'recoil';
+
 import Box from '@mui/material/Box';
 
 import { Markdown } from 'components/molecules/Markdown';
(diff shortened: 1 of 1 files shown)


willydouhard (approved):

willydouhard:
Looks like the build is failing @jdb78

frontend/src/components/atoms/elements/Text.tsx
```diff
@@ -1,17 +1,23 @@
+import { useRecoilValue } from 'recoil';
+
 import Box from '@mui/material/Box';
 
 import { Markdown } from 'components/molecules/Markdown';
 
 import { useFetch } from 'hooks/useFetch';
 
 import { type ITextElement } from 'client-types/';
+import { projectSettingsState } from '@chainlit/app/src/state/project';
```
tpatel, line 10: You can use a direct import because this file is already within the `@chainlit/app` package.

```suggestion
import { projectSettingsState } from 'state/project';
```

jdb78: done

jdb78:
should be fixed - think you can change settings so you can avoid having to approv every time checks need to be run

PR #896, opened 2024-04-10
 Newly created thread displayed in the side bar; Resumed threads content is revalidated when navigating in the side bar

tpatel:
Original PR by @BrittleFoot : https://github.com/Chainlit/chainlit/pull/838.


tpatel:
Sorry for the mess, I won't be force pushing contributor branches again anytime soon, I'm pushing your code back into your repo ✅

PR #900, opened 2024-04-12
fix: mypy typings for decorators

sandangel:
Fixed mypy typings of decorators.
Formatted with black/isort. 


backend/chainlit/__init__.py
sandangel, line 1: This is the only file that I made change. Other files were formatted using black

dahifi:
Looks like main had no black pass on it? I assume your code didn't force all those changes, might be better to skip the lint on this one and pass just your changes. Twenty files is a lot to review, and the linting can be done through CI on the PR.

PR #905, opened 2024-04-13
Adding SSL certificate + key file parameters to run command (passthrough to uvicorn webserver)

shanelenagh:
I added optional parameters to pass an SSL certificate file and key file through to the webserver in the run command, as either a CLI or env var using the existing pattern, so that traffic can be secured from the app source (installing behind a reverse-proxy might not be desirable in all cases, and if the proxy is outside the host or container of this app there might be security constraints requiring that traffic to be encrypted as well, which was the reason I implemented this for my use case).  This passes through to the webserver (uvicorn) config object, and I have confirmed it provides an SSL connection from the source when proper cert+key files are used.

### DIFF
--- a/backend/chainlit/cli/__init__.py
+++ b/backend/chainlit/cli/__init__.py
@@ -37,6 +37,8 @@ def cli():
 def run_chainlit(target: str):
     host = os.environ.get("CHAINLIT_HOST", DEFAULT_HOST)
     port = int(os.environ.get("CHAINLIT_PORT", DEFAULT_PORT))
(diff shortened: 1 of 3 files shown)


tpatel:
Could you share more about your use case where the reverse proxy isn't an option?

shanelenagh:
Sorry @tpatel , life got busy and I missed this thread. Glad to see another PR with similar functionality is being merged, pending those changes requested. For the record we have requirements where the content needs to be encrypted at the source, even as it's traversing over our local network to the distinct proxy server (not a local node container), per our network security policy.

PR #906, opened 2024-04-14
Refactor socket.py to handle task interruption gracefully

jpolvto:
When the user presses the stop button, the backend raises an InterruptedError. In my mind this should be logged instead. Let me know if I've missed something!


jpolvto:
Will fix the test.

willydouhard:
Then it won't actually stop the task I believe. Currently it is the error that will interrupt an async task

jasperpolaxinter:
This is solved in the latest version, so I'll remove the pull request

jpolvto:
This is resolved in the latest version, I'll close the pull request

PR #916, opened 2024-04-17
feat: Add node metadata in element content for CBEventType.RETRIEVE in Llama_index callback.

amait41:
Commit non testé.


dokterbob:
@amait41 Thanks for the contrib!

Sorry it took a long time to get back to you, we've been quite overwhelmed by contributions...

Could you please provide a clear description of the what and the why of your contribution?

What's the use case, why would other users want this?

In addition, to get your contribution merged, please resolve any merge conflicts!

If you're able to make requested changes in the coming weeks, that'd be great. We're trying to clean house a bit so either we get it in or we'll drop it. 

Thanks for understanding!

PR #921, opened 2024-04-18
Fix SQLAlchemyDataLayer in sql_alchemy.py

mohamedalani:
Fixing 2 bugs: 
- adding `self.storage_provider = None` in `def __init__()` so that it does not break method create_element. 

- adding generation parameter in create_step as it's also a JSON type

### DIFF
--- a/backend/chainlit/data/sql_alchemy.py
+++ b/backend/chainlit/data/sql_alchemy.py
@@ -1,29 +1,45 @@
-import uuid
-import ssl
-from datetime import datetime, timezone
 import json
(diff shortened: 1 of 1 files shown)


intc-hharshtk:
THanks for the fix in sql_alchemy.py, been stuck with parameters generation.

mohamedalani:
@hayescode @willydouhard what do you guys think?

tpatel (approved):

willydouhard:
Looks like you have a mypy issue:
```
chainlit/data/sql_alchemy.py:42: error: Incompatible types in assignment (expression has type "None", variable has type "BaseStorageClient")  [assignment]
```

PR #925, opened 2024-04-19
Allow None for timeout in AskSpec to support indefinite waits

AidanShipperley:
## Overview

Updated the AskSpec data class to support `None` as a valid value for the `timeout` field, allowing developers to force a wait without a timeout constraint. Previously, passing `timeout = None` would trigger a Pydantic validation error, making the only solution to pass an extremely large integer to achieve a pseudo-infinite wait.

## Changes

- Adjusted the timeout field type from `int` to `Optional[int]` within the AskSpec data class.
  - This should allow developers to add indefinite waits when using `AskUserMessage`, `AskFileMessage`, and `AskUserAction` by passing `timeout = None`.
- The `types.py` file has several adjustments to it's spacing and import structure as a result of the pre-commit hook set up in your repository, the only thing I actually changed was one line.

## Rationale

When asking the user for a file, for example, before a model can answer question, it's possible to not want the operation to timeout ever. The user may leave their computer and move onto something else for hours, but the file input should still work. Setting the variable `timeout = None` is very intuitive and clearly shows the intention of the developer to not allow a timeout to occur.

## Testing

- I followed all steps in the contribution guidelines and all tests passed locally.
- I additionally tested that `AskUserMessage`, `AskFileMessage`, and `AskUserAction` worked with passing `timeout = None` with this change, and they all did as far as I could tell.

### DIFF
--- a/backend/chainlit/types.py
+++ b/backend/chainlit/types.py
@@ -1,5 +1,17 @@
 from enum import Enum
-from typing import TYPE_CHECKING, Dict, List, Literal, Optional, TypedDict, Union, Generic, TypeVar, Protocol, Any
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Generic,
+    List,
+    Literal,
+    Optional,
+    Protocol,
+    TypedDict,
+    TypeVar,
+    Union,
+)
 
 if TYPE_CHECKING:
     from chainlit.element import ElementDict
@@ -37,6 +49,7 @@ class ThreadFilter(BaseModel):
     userId: Optional[str] = None
     search: Optional[str] = None
 
+
 @dataclass
 class PageInfo:
     hasNextPage: bool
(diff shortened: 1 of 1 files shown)


tpatel:
Thanks for this PR, this is a valid feature. You will need to change more code to enable the no-timeout behavior.

One example: when you pass `timeout = None` to `AskUserMessage`, the `timeout` value is set to 60 seconds, instead of `None`. 

Could you make the change in the `Ask` classes to allow for a `None` timeout and to enable the disable the timeout?

tpatel:
Closing as the discussion has stopped, we can still reopen the PR in the future

AidanShipperley:
Hi @tpatel,

Apologies for the delay; I was reassigned to other tasks at work.

Could you please provide more details on the required changes for this feature? Specifically, I'm confused because when I change timeout to `Optional[int]` in the `AskSpec` class and pass `timeout=None` to `AskUserMessage`, it waits far longer than 60 seconds. I can still interact with it and submit a message even after an hour.

Is there something I'm missing? Any clarification would be greatly appreciated so I can make the necessary adjustments.

Thank you!

PR #943, opened 2024-04-27
use the label for menu item

kevinwmerritt:
This ensures menu items display the item label rather than value.

**ex:**
```python
@cl.on_chat_start
async def on_chat_start():
    await cl.ChatSettings(
        [
            Select(
                id="Model",
                label="OpenAI - Model",
                items={
                    "GPT-3.5-Turbo": "gpt-3.5-turbo",
                    "GPT-4": "gpt-4-turbo"
                },
                initial_value="gpt-4-turbo",
            )
        ]
    ).send()
```

**bug:** the value is displayed instead of the label for each menu option
image

**fix:**
Screenshot 2024-04-29 at 8 39 00 AM

### DIFF
--- a/frontend/src/components/atoms/inputs/selects/MenuItem.tsx
+++ b/frontend/src/components/atoms/inputs/selects/MenuItem.tsx
@@ -43,7 +43,7 @@ const MenuItem = ({
   >
     <Stack direction="row" alignItems="center" spacing={1}>
       {item.icon ? item.icon : null}
-      <span>{item.value}</span>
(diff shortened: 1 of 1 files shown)


tpatel:
This looks like a good change, could you share some context about what this change is fixing?

kevinwmerritt:
@tpatel I updated the PR description with more detail.

tpatel (approved):
Thanks for your contribution and the updated PR context ✨

PR #944, opened 2024-04-27
do not prevent thread revalidation

kevinwmerritt:
Ensures returning to a previously created chat will display all the messages.


tpatel:
Thanks for the issue + PR 🙏 

Is there a reason why you've removed `revalidateOnFocus: false` ? It looks like the problem was already fixed when you've removed `revalidateIfStale: false`.

kevinwmerritt:
@tpatel there was not a specific reason. I put back `revalidateOnFocus: false`. Thanks!

tpatel (approved):
And thanks for the update 👍

PR #956, opened 2024-05-01
Fix Haystack tool step text

mosheber:
Fixing a bug where the tool_result text is not displayed in the UI.
* replaced the tool_result update by calling stream_token
* replaced the id of the tool step with the parent id of the step above it, instead of the id itself.


mosheber:
@willydouhard is there something I should add to this PR?

PR #958, opened 2024-05-02
feat(storage): add gcs as a storage client

shuvamg007:
2 non-breaking changes:

- Adds GCS as a storage option
- S3 storage accepts credentials through parameters (useful in some scenarios like secrets manager)

### DIFF
--- a/backend/chainlit/data/storage_clients.py
+++ b/backend/chainlit/data/storage_clients.py
@@ -39,16 +39,32 @@ class AzureStorageClient(BaseStorageClient):
 class S3StorageClient(BaseStorageClient):
     """
     Class to enable Amazon S3 storage provider
+
+    params:
+        bucket: Name of the S3 bucket
+        access_key: AWS access key
+        secret_key: AWS secret key
+        session_token: AWS session token
     """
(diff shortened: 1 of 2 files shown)


backend/chainlit/data/storage_clients.py
```diff
@@ -39,20 +40,68 @@ async def upload_file(self, object_key: str, data: Union[bytes, str], mime: str
...
     Class to enable Amazon S3 storage provider
+
+    params:
+        bucket: Name of the S3 bucket
+        access_key: AWS access key
+        secret_key: AWS secret key
+        session_token: AWS session token
     """
-    def __init__(self, bucket: str):
+    def __init__(self, bucket: str, access_key: Optional[str] = None, secret_key: Optional[str] = None, session_token: Optional[str] = None):
```
mayaankvad, line 50: A quick trick here is to accept the s3 client itself as an optional arg, and if it's null make a default one. This allows true customization for the end user. For example, they may want to use the Stubber for unittests or pass in a botocore.config.Config to customize the behavior.

willydouhard:
This has been done in the `2.0.0` release!

PR #981, opened 2024-05-09
Fix SQLAlchemy bug and change add logging option

hayescode:
Fixes bug that shows input even when show_input = False because the boolean is actually a string
```python
# before
input=(
    step_feedback.get("step_input", "")
    if step_feedback["step_showinput"]
    else ""
),
# after
input=(
    step_feedback.get("step_input", "")
    if step_feedback["step_showinput"] == "true" 
    else None
),
```

Added Logging option to init `show_logger`
- Since I'm using chainlit's logger it's difficult to selectively silence so this optionally shows info/debug logs during development.

### DIFF
--- a/backend/chainlit/data/sql_alchemy.py
+++ b/backend/chainlit/data/sql_alchemy.py
@@ -39,9 +39,11 @@ class SQLAlchemyDataLayer(BaseDataLayer):
         ssl_require: bool = False,
         storage_provider: Optional[BaseStorageClient] = None,
         user_thread_limit: Optional[int] = 1000,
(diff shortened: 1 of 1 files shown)


ckrapu-nv:
@hayescode, has there been any documentation on the database schema required by the SQLAlchemy data layer? I'd like to use it and am wondering if that info has to be read from the code or if it's been recorded somewhere.

hayescode:
@ckrapu-nv #836 has it in the PR description. Since then threadId has been added to feedbacks, and it seems ~50% of steps fields are going away soon.

PR #982, opened 2024-05-09
Fix: defined default var for langchain cache path to enable custom path (issue 987)

necjamahim:
When defining `lc_cache_path`  and enabling cache in config:
``` toml
# Enable third parties caching (e.g LangChain cache)
cache = true
lc_cache_path = './.custom-cache-path.db'
```

TypeError is thrown:
``` bash
TypeError: chainlit.config.ProjectSettings() got multiple values for keyword argument 'lc_cache_path'
```

### DIFF
--- a/backend/chainlit/cache.py
+++ b/backend/chainlit/cache.py
@@ -19,7 +19,7 @@ def init_lc_cache():
         if config.project.lc_cache_path is not None:
             set_llm_cache(SQLiteCache(database_path=config.project.lc_cache_path))
 
-            if not os.path.exists(config.project.lc_cache_path):
+            if os.path.exists(config.project.lc_cache_path):
                 logger.info(
                     f"LangChain cache created at: {config.project.lc_cache_path}"
                 )
--- a/backend/chainlit/config.py
+++ b/backend/chainlit/config.py
@@ -54,6 +54,9 @@ session_timeout = 3600
 # Enable third parties caching (e.g LangChain cache)
 cache = false
 
+# Set langchain cache path (default is .langchain.db inside config dir)
+# lc_cache_path = './.my-cache-path.db'
+
 # Authorized origins
 allow_origins = ["*"]
 
@@ -268,9 +271,9 @@ class CodeSettings:
 
     author_rename: Optional[Callable[[str], str]] = None
     on_settings_update: Optional[Callable[[Dict[str, Any]], Any]] = None
-    set_chat_profiles: Optional[Callable[[Optional["User"]], List["ChatProfile"]]] = (
-        None
-    )
+    set_chat_profiles: Optional[
+        Callable[[Optional["User"]], List["ChatProfile"]]
+    ] = None
 
 
 @dataclass()
@@ -280,7 +283,8 @@ class ProjectSettings(DataClassJsonMixin):
     # List of environment variables to be provided by each user to use the app. If empty, no environment variables will be asked to the user.
     user_env: Optional[List[str]] = None
     # Path to the local langchain cache database
-    lc_cache_path: Optional[str] = None
+    # default to "{config_dir}/.langchain.db"
+    lc_cache_path: Optional[str] = os.path.join(config_dir, ".langchain.db")
     # Path to the local chat db
     # Duration (in seconds) during which the session is saved when the connection is lost
     session_timeout: int = 3600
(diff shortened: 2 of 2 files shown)


backend/chainlit/config.py
```diff
@@ -420,10 +424,7 @@ def load_settings():
                 "Your config file is outdated. Please delete it and restart the app to regenerate it."
             )
 
-        lc_cache_path = os.path.join(config_dir, ".langchain.db")
-
         project_settings = ProjectSettings(
```
tpatel, line 427: I believe you're missing something to change the `project_config["lc_cache_path"]` value, like:

```suggestion
        project_config["lc_cache_path"] = project_config.get("lc_cache_path", lc_cache_path)
        project_settings = ProjectSettings(
```

backend/chainlit/cache.py
```diff
@@ -19,7 +19,7 @@ def init_lc_cache():
         if config.project.lc_cache_path is not None:
             set_llm_cache(SQLiteCache(database_path=config.project.lc_cache_path))
 
-            if not os.path.exists(config.project.lc_cache_path):
+            if os.path.exists(config.project.lc_cache_path):
```
tpatel, line 22: I'm not sure to understand why you've removed this?

tpatel (changes requested):
Thanks for your code, I've added two comments.

backend/chainlit/config.py
necjamahim: the way i understand it and tested it locally, possible custom path will be picked up by `toml_dict` module [here](https://github.com/Chainlit/chainlit/blob/c153257fde517711d8d387de0f9add1ca925d810/backend/chainlit/config.py#L417C9-L417C54): 
``` python
project_config = toml_dict.get("project", {})
```
or will default to default value as proposed by my PR [here](https://github.com/Chainlit/chainlit/pull/982/files/c153257fde517711d8d387de0f9add1ca925d810#diff-e76edd7e411b9e044dbc1f307015efc5d8e83f7b254af919786191f86dfc7c74R287)
```python
# default to "{config_dir}/.langchain.db"
lc_cache_path: Optional[str] = os.path.join(config_dir, ".langchain.db")
 ```

File: backend/chainlit/cache.py (hunk 2 above)
Comment (necjamahim) [reply backend/chainlit/cache.py:22]: it is because previous line `st_llm_cache` generates sqlite db file at path
```python
set_llm_cache(SQLiteCache(database_path=config.project.lc_cache_path))
```
consequently in next line we want to inform cache was generated by testing if path exists and not the opposite. I hope I was clear - was trying not to use too many "if", "not" in the explaination

backend/chainlit/config.py
```diff
@@ -268,9 +271,9 @@ class CodeSettings:
 
     author_rename: Optional[Callable[[str], str]] = None
     on_settings_update: Optional[Callable[[Dict[str, Any]], Any]] = None
-    set_chat_profiles: Optional[Callable[[Optional["User"]], List["ChatProfile"]]] = (
-        None
-    )
+    set_chat_profiles: Optional[
+        Callable[[Optional["User"]], List["ChatProfile"]]
+    ] = None
```
dokterbob, line 276: How is this relevant to the issue at hand?

backend/chainlit/config.py
```diff
@@ -54,6 +54,9 @@
 # Enable third parties caching (e.g LangChain cache)
 cache = false
 
+# Set langchain cache path (default is .langchain.db inside config dir)
+# lc_cache_path = './.my-cache-path.db'
```
dokterbob, line 58: If this is a default config option, it seems better to uncomment it, in line with other default config options.

backend/chainlit/config.py
dokterbob: I think @necjamahim is right on this one. Could you please have another look @tpatel?

dokterbob (changes requested):
Thanks for your contribution @necjamahim!

I see where you're coming from; current config doesn't allow overriding the langchain cache configured through `lc_cache_path`, which seems hardcoded to `{config_dir}/.langchain.db`.

So this patch keeps the previously hardcoded value as a default and makes it configurable. 

However, I don't see how your changes to `set_chat_profiles` come in. Could you please remove this from your patch and resolve any merge conflicts?

As soon as that's done I'll do a final manual test and I think we're good to go!

PR #999, opened 2024-05-15
Added threadId to Feedback

mayaankvad:
Added threadId to Feedback. This will allow for getting the associated threadId easily. Plan to use this in adding document database data layers to chainlit like Dynamo. Discussed with @willydouhard

### DIFF
--- a/backend/chainlit/discord/app.py
+++ b/backend/chainlit/discord/app.py
@@ -3,13 +3,12 @@ import mimetypes
 import re
 import uuid
 from io import BytesIO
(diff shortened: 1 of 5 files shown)


willydouhard:
I think this is great! This does not update the data layers but this is probably out of scope. I assume you are using a custom data layer yourself?

mayaankvad:
Yes I wanted to keep this PR small. The literal and sql data layers also don't need to use the threadId property so there was no reason to update them.

Issue #964, opened 2024-05-05
Feature: Make HTTP Headers Available on Start or UserSession


christoferson:
**Is your feature request related to a problem? Please describe.**
I am hosting the app in front of a load balancer that does the Authentication and the results are sent as HTTP Headers to the Application. I would like to access the username from the Headers that are passed in. However I cannot find a callback to access the values.

**Describe the solution you'd like**
Would be great if we can have the HTTP headers made available in the @cl.on_chat_start or make it available in the UserSession object.

**Describe alternatives you've considered**
Create Custom Javascript to fetch the information via websockets. Not all hosting environments support WS connection. Also, having it part of the built in callbacks and API is cleaner.

wfjt:
You can do this in `socketio.pu`, the `@connect` callback receives the `asgi` `environ`, didn't test but you could put this at the end of `@connect` and try it out, just after the ws session is initialized:

```python
    asgi_scope  = environ.get("asgi.scope", {})
    headers: List[Tuple[bytes, bytes]] = asgi_scope.get("headers", {})
    headers_dict: Dict[str, str] = {}
    headers = [headers_dict.update({k.decode("utf-8"): v.decode("utf-8")})
               for k, v in headers if k.decode("utf-8") != "authorization"]

    setattr("ws_session", "headers", headers)
    # from chainlit.session import ws_sessions_id, ws_sessions_sid
    ws_sessions_id[session_id] = ws_session
    ws_sessions_sid[sid] = ws_session
    if session_id in user_sessions:
        user_sessions[session_id]["headers"] = headers_dict.copy()
```

Didn't check the session handling logic, not sure if these would get overwritten but mostly they seem to check if their session id is found in one of those three and if so the value is used.

It's not clear to me how the session ID flows from the ws endpoint to context.session, maybe one of those emitter events triggers this. No time to comb through it all.

dokterbob:
Closing this in favour of later issues describing this.

Issue #393, opened 2023-09-13
Is it possible to access the full request object, in particular the headers?


Midnighter:
When I include a request parameter in my message handler

```py
import chainlit as cl
from fastapi import Request

@cl.on_message
def handle_message(message: str, request: Request):
    print(request)
```

It turns out that the request object is just a string with presumably the session id. Is it at all possible to access the full object or at least the headers of the request?

willydouhard:
This is not possible since on_message is a websocket message, not an http request. Do you want to access the headers for authentication purposes?

Midnighter:
> This is not possible since on_message is a websocket message, not an http request.

I was afraid that's the case.

We have [Posit Connect](https://docs.posit.co/rsc/) running to handle authentication. However, I was still hoping to grab some kind of user identifier from the headers. It was just a try, I'm not sure there even would be useful information there.

willydouhard:
we could expose the handshake headers. However we are going to release today or tomorrow a completely overhauled authentication system. It will include a `header_auth` for companies running chainlit through a reverse proxy. Would love to get your feedback before considering exposing the headers :)

Midnighter:
Happy to test something. Just let me know which branch or release candidate you would like me to test.

willydouhard:
Here is the [announcement](https://discord.com/channels/1088038867602526210/1096367751481802812/1151915856977203230) with all the details

Midnighter:
I'll give it a whirl. Thanks.

Midnighter:
Okay, I've tried the new versions and successfully accessed the headers using the `@cl.header_auth_callback` handler.

My main question now is, can I somehow access the current user in a `cl.on_message` handler? You must somehow know the correct user since you are exposing the specific `cl.user_session` object. I would very much like to get a user identifier.

willydouhard:
Yes, this is not documented enough but you can see an example in the[ password test](https://github.com/Chainlit/chainlit/blob/main/cypress/e2e/password_auth/main.py)

Midnighter:
Perfect, thank you! 😃

Siddhijain16:
> Okay, I've tried the new versions and successfully accessed the headers using the `@cl.header_auth_callback` handler.
> 
> My main question now is, can I somehow access the current user in a `cl.on_message` handler? You must somehow know the correct user since you are exposing the specific `cl.user_session` object. I would very much like to get a user identifier.

Hi @Midnighter 
[...]

PR #1002, opened 2024-05-18
Add availability for X-Forwarded-For HTTP header

acordier16:
- Add X-Forwarded-For HTTP header availability in the Chainlit user session

The goal is to be able to gather the client IP address from the Chainlit user session, in order for example to apply rate limiting on client IP adresses using limits.

Note: currently, the headers are not available within the session (see https://github.com/Chainlit/chainlit/issues/964 and https://github.com/Chainlit/chainlit/issues/393).

Usage:

`
client_ip = cl.user_session.get("http_forwarded_for")
`


dokterbob:
Seems like a must-have feature for many production applications. But, then again, would this not be superseded by generic access to the full request in `on_chat_start()` as proposed in #1213?

PR #1005, opened 2024-05-19
Added threadId to Element

mayaankvad:
Adding this as prep for a custom data layer. This change will allow more flexible custom data layers, especially document ones like Dynamo.

Currently you can get the threadId in create element:
```python
async def create_element(self, element: "Element"):
        thread_id = element.thread_id)
```

This change adds the thread_id param to delete element:
```python
async def delete_element(self, element_id: str, thread_id: Optional[str]):
```

All current data layers have been modified to accept the new param. The [Custom Data Layer API Docs](https://docs.chainlit.io/api-reference/data-persistence/custom-data-layer) will need to be updated. Should not be a breaking change but should be a mention in a migration guide.

### DIFF
--- a/backend/chainlit/data/__init__.py
+++ b/backend/chainlit/data/__init__.py
@@ -94,7 +94,7 @@ class BaseDataLayer:
         pass
 
     @queue_until_user_message()
-    async def delete_element(self, element_id: str):
+    async def delete_element(self, element_id: str, thread_id: Optional[str] = None):
         pass
 
     @queue_until_user_message()
@@ -340,7 +340,7 @@ class ChainlitDataLayer(BaseDataLayer):
         return self.attachment_to_element_dict(attachment)
 
     @queue_until_user_message()
-    async def delete_element(self, element_id: str):
+    async def delete_element(self, element_id: str, thread_id: Optional[str] = None):
         await self.client.api.delete_attachment(id=element_id)
 
     @queue_until_user_message()
--- a/backend/chainlit/data/__init__.py
+++ b/backend/chainlit/data/__init__.py
@@ -94,7 +94,7 @@ class BaseDataLayer:
         pass
 
     @queue_until_user_message()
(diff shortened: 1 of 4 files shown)


cypress/e2e/data_layer/main.py
```diff
@@ -114,7 +114,7 @@ async def get_thread_author(self, thread_id: str):
         return "admin"
 
     async def list_threads(
-        self, pagination: cl_data.Pagination, filter: cl_data.ThreadFilter
+        self, pagination: cl_data.Pagination, filters: cl_data.ThreadFilter
```
mayaankvad, line 117: Fixed error:

```
Parameter 'filters' has been renamed to 'filter' in overriding 'TestDataLayer.list_threads' method PylintW0237:arguments-renamed

Method "list_threads" overrides class "BaseDataLayer" in an incompatible manner
  Parameter 3 name mismatch: base parameter is named "filters", override parameter is named "filter" Pylance reportIncompatibleMethodOverride
__init__.py(118, 15): Overridden method
```

backend/chainlit/data/sql_alchemy.py
mayaankvad, line 1: Accidentally auto formatted this to change more lines than intended. But the changes are all just putting the logger statement on a new line. Leaving this minor style fix in, happy to revert if needed.

backend/chainlit/data/__init__.py
```diff
@@ -94,7 +94,7 @@ async def get_element(
         pass
 
     @queue_until_user_message()
-    async def delete_element(self, element_id: str):
+    async def delete_element(self, thread_id: str, element_id: str):
```
willydouhard, line 97: this is a breaking change, i would put it after element_id as optional?

backend/chainlit/data/sql_alchemy.py
mayaankvad, line 1: git hook `pnpm lint-staged` has made changes everywhere here to add the logger on a new line.

PR #1012, opened 2024-05-22
added Gitlab OAuth provider

matchaniat:
Added Gitlab OAuth provider (tested for self-hosted instance)


willydouhard:
Super cool, I think only the [icon](https://github.com/Chainlit/chainlit/blob/main/frontend/src/components/molecules/auth/ProviderButton.tsx) is missing.

matchaniat:
> Super cool, I think only the [icon](https://github.com/Chainlit/chainlit/blob/main/frontend/src/components/molecules/auth/ProviderButton.tsx) is missing.

Done !

image

tpatel (approved):

PR #1017, opened 2024-05-23
Keeping the given content and the selected action in AskActionMessage function

NaguiHW:
I was developing a kind of admin manager view for a user with admin role and this admin can upload more data and delete data from the database and I was using the `AskActionMessage` function and I found that is a bit confusing only show the default message: `**Selected:** {res["label"]` when we are asking more than 1 action to the user.

This is why I created this pr to solve this issue.

This is an example:
Screenshot 2024-05-23 002612
If you see the 2 first answers, you don't know the context, but in the third one you can see the question and the action selected.

I just added an optional parameter `keep_content` that is a boolean and is `False` by default.

Of course this is useful only if we are going to ask more than 1 action. Like I said when I was developing this type of view, like an admin panel, I have to use at least 6 `AskActionMessage`.

I know that we have `@cl.action_callback("action_button")`, but is not the same, because the `input query` is not disabled and is not that clear how we can delete all other action buttons if we have more than 1 action.


dokterbob:
@NaguiHW Thanks for this great contrib! Sorry it took us quite a while to get back to you!

Does your contrib has any implications for the documentation? If so, could you please propose any relevant changes to the [docs](https://github.com/Chainlit/docs)

Lastly, I want to test this! Any chance you could supply me a minimal test project and basic testing instructions so I can observe the before/after difference so I can manually reproduce the issue and observe your proposed solution? ❤️

willydouhard:

PR #1019, opened 2024-05-24
Added Norwegian translation

buzzCraft:
I saw that you don't necessary want to host translations in the repo, but adding this in case you find a way to host it.

A suggestion would be to pin a maintainer for each language, and give them a heads-up whenever changes are needed.


tpatel:
Thanks for your input, I understand it would benefit users to have accurate Chainlit translations, however we haven't chosen the right way to store them and keep them updated.

Regarding your commit, it looks like the file is empty there.

I'll close this PR for now while we figure out a solution.

buzzCraft:
Oh, you are right.. Wonder what happened there 😅 
Feel free to ping me when/if you want to start building translation sets

PR #1026, opened 2024-05-26
Adding the ability to specify a Root Path (Sub-Directory) 

jayteaftw:
Chainlit cannot currently specify a root path (sub-directory) when serving an application.
This makes it difficult for us who do not want to run on a subdomain or the main domain itself.

Drafted Implemented Logic
1. Takes a root_path as a new Chainlit run option (Similar to host and port, and defaults to an empty string if not set.
2. In Python, sets the FastAPI root_path
3. Passes the root_path from Python to React through the DOM.
4. In React, sets the ```createBrowserRouter``` basename option as the root_path

I believe the Python steps are sound; however, I am unsure about the React portion of the logic as I am not too familiar with JS/TS & React. 

Although it passes the test cases locally, it needs to be looked at more closely as I believe the test cases cannot test for the implemented changes. (I.E. it will only test when the root_path is equal to an empty string therefore nothing has changed)


willydouhard:
This looks super interesting, will try that this week. Thank you for your contribution!

jayteaftw:
> This looks super interesting, will try that this week. Thank you for your contribution!

Sounds good! I tested the normal chatbot functionality and it seems to be working; however, when I tested the login feature it redirected me to ```/login``` not to ```root_path/login```. Locally, I seemed to have fixed that problem; however when I try to push with the latest code I get the error 
```
chainlit/data/__init__.py:233: error: "AsyncLiteralAPI" has no attribute "get_my_project_id"  [attr-defined]
Found 1 error in 1 file (checked 47 source files)
 ELIFECYCLE  Command failed with exit code 1.
husky - pre-commit hook exited with code 1 (error)
```

Furthermore, I am not quite sure what code in the React portion I need to change.
For example, to get the login functionality to work, I had to update the ```window.location.pathname``` and ```window.location.href``` to use the path_root. Would this also apply to each instance under ```src/pages/``` that use the ```navigate ()``` function? For example ```navigate('/')``` would become ```navigate(getRouterBasename()+'/')```?

kellerkind84:
Hi guys, thanks for working on this... for me, this is basically essential to be able to host this anywhere. Hosting two chainlit apps on the same domain seems to be impossible, and this would fix it. So, the fast you can get this to work, the better. ;)

willydouhard:
Thank you for your contribution 🥳 . I had to make a couple of changes but it was a great foundation. I ended up pulling your repo in the release branch https://github.com/Chainlit/chainlit/pull/1064. Your contribution will be clearly visible. Let me know if that works for you!

jayteaftw:
> Thank you for your contribution 🥳 . I had to make a couple of changes but it was a great foundation. I ended up pulling your repo in the release branch #1064. Your contribution will be clearly visible. Let me know if that works for you!

Sounds great to me! I really want this feature to be added 😄 so however it gets in there works for me

dokterbob:
AFAIK, this is a released feature now, right? 

Can we close this one?

dokterbob:
Adding back-reference for future proposal #1220

dokterbob:
@jayteaftw What's the status of this? Isn't this in main already?

If not, how is it different from what's in main? 

If it is, can we close it? 🙏🏼

PR #1027, opened 2024-05-26
Update sql_alchemy.py - bugfix for delete_step

ralphkink:
Fixed bug in sql_alchemy.py:delete_step: Wrong/non-existing primary key name for steps_query (was "forId" instead of "id", "forId" does not exist in table "steps") - probably a copy/paste bug from the previous line of code.


willydouhard (approved):

ralphkink:
Hi @willydouhard - this is a very simple and straight forward fix for a bug that breaks my code as I cannot delete steps/threads in the custom data layer.
Any reason why this has not been merged?
Thanks!

willydouhard (approved):

PR #1028, opened 2024-05-26
Release/1.1.300

willydouhard:


minhduc0711:
Why was `default_expand_messages` removed? Is there anyway I can have this auto expand behavior in 1.1.300?

Issue #1031, opened 2024-05-28
Allow user session timeout to be configurable


wholesomewilson:
**Is your feature request related to a problem? Please describe.**
Currently, the user session timeout is hardcoded to 15 days. https://github.com/Chainlit/chainlit/blob/126970fa8b33fe5fa8707e0fb54f00069522ca7f/backend/chainlit/auth.py#L55

In some use cases, there are requirements for longer or shorter timeout.

**Describe the solution you'd like**
We propose to shift the user session timeout to config.toml using the variable name `user_session_timeout`

**Describe alternatives you've considered**
We have considered using environment variables. However, since session_timeout is implemented in config.toml, we decided to follow this convention.

PR #1032, opened 2024-05-28
feat: Add configurable user_session_timeout

wholesomewilson:
Addresses #1031

This adds `user_session_timeout` variable to config.toml to make user session timeout configurable.

### DIFF
--- a/backend/chainlit/config.py
+++ b/backend/chainlit/config.py
@@ -62,6 +62,9 @@ user_env = []
 # Duration (in seconds) during which the session is saved when the connection is lost
 session_timeout = 3600
 
+# Duration (in seconds) of the user session expiry
+user_session_timeout = 1296000  # 15 days
+
 # Enable third parties caching (e.g LangChain cache)
 cache = false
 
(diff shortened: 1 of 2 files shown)


Physium:
Any updates if this can be merged?

backend/chainlit/config.py
```diff
@@ -51,6 +51,9 @@
 # Duration (in seconds) during which the session is saved when the connection is lost
 session_timeout = 3600
 
+# Duration (in seconds) of the user session expiry
+user_session_timeout = 1296000
```
dokterbob, line 55: I would suggest adding that it's 15 days to the comment here.

backend/chainlit/config.py
```diff
@@ -297,6 +300,8 @@ class ProjectSettings(DataClassJsonMixin):
     # Path to the local chat db
     # Duration (in seconds) during which the session is saved when the connection is lost
     session_timeout: int = 3600
+    # Duration (in seconds) of the user session expiry
+    user_session_timeout: int = 60 * 60 * 24 * 15
```
dokterbob, line 304: I'd suggest adding comment on 15 days here.

dokterbob (changes requested):
The feature is well explained and clearly scoped and code looks clean -- but outdated. Could you please rebase and add requested comments?

If tests pass I think we should be good to go.

backend/chainlit/config.py
aniruddha-adhikary: We just added a comment there to suggest `15 days`.

backend/chainlit/config.py
aniruddha-adhikary: Comment added as well!

aniruddha-adhikary:
Hi @dokterbob, we have made the changes requested.

Issue #983, opened 2024-05-09
HIGH SECURITY VIOLATION in cypress/support/utils.ts


hayescode:
**Describe the bug**
My workplace uses Apiiro for security compliance and [this section of code](https://github.com/Chainlit/chainlit/blob/main/cypress/support/utils.ts#L26-L33) gives a high severity security violation.

Error message: `Improper Neutralization of Special Elements used in an OS Command ('OS Command Injection')`

References: https://cheatsheetseries.owasp.org/cheatsheets/Nodejs_Security_Cheat_Sheet.html#do-not-use-dangerous-functions

![image](https://github.com/Chainlit/chainlit/assets/35790761/fb34a2f5-d231-471e-ab51-6bf54c36f50d)

willydouhard:
Thank you for the report. This code is only used to orchestrate the CI. It never lands in any Chainlit application.

That being said, it would be nice to clear the warning!

dokterbob:
Looking at the code it, I don't think there's a trivial work around.

Nor an underlying security issue! Note: "[...] if the input is user controllable."

In the CI context _there is no user_, it's simply a wrapper around `execSync()` to ensure that tests are called with proper `pnpm` context. To my knowledge, there's no way to do this without forking, pnpm doesn't have a supported/documented JS API. Nor is this necessary.

Rather, best to tell the security scanner that this is a false positive.

dokterbob:
Closing this as it's a false positive (see comment before).

PR #1036, opened 2024-05-28
Fix security vulnerability in cypress/support/utils.ts

hayescode:
Fixes issue #983


cypress/support/utils.ts
```diff
@@ -24,10 +24,13 @@ export async function runTests(matchName) {
 }
 
 export function runCommand(command: string, cwd = ROOT) {
-  return execSync(command, {
+  const [cmd, ...args] = command.split(' ');
+  const fullCommand = [cmd, ...args].join(' ');
```
willydouhard, line 28: in the end the command being ran is the same right? Your security scanner happens to prefer this syntax?

cypress/support/utils.ts
```diff
@@ -24,10 +24,13 @@ export async function runTests(matchName) {
...
+  return execSync(fullCommand, {
     encoding: 'utf-8',
     cwd,
     env: process.env,
-    stdio: 'inherit'
+    stdio: 'inherit',
+    shell: process.platform === 'win32' ? 'cmd.exe' : '/bin/sh',
```
willydouhard, line 34: can you explain this addition?

cypress/support/utils.ts
hayescode: @willydouhard i based this on gpt4 feedback but you're right, this doesn't seem to remove the vunerability. The scanner doesn't run every day and the last check (where I manually added this) still has it.

Any other ideas?

dokterbob:
Closing for now. Rationale: https://github.com/Chainlit/chainlit/issues/983#issuecomment-2285848329

Issue #691, opened 2024-01-24
Login issue: Invalid request for /login/callback


jdb78:
This is a strange one and only seems to happen if the app is deployed on Azure App Service. At login, sometimes (not always!), I get a `Invalid HTTP request received.` with status: `400 Bad request`. If you refresh long enough, you eventually get logged in.

Here is a log with chainlit `1.0.200` where I annonymized some details.

You can see that it all passes the oauth callback (which does some additional pulling of data from graphapi) but then seems to error

```
2024-01-24T10:16:07.043042428Z 2024-01-24 10:16:07 - load_ssl_context verify=True cert=None trust_env=True http2=False
2024-01-24T10:16:07.043568126Z 2024-01-24 10:16:07 - load_verify_locations cafile='/opt/pysetup/.venv/lib/python3.11/site-packages/certifi/cacert.pem'
2024-01-24T10:16:07.049984803Z 2024-01-24 10:16:07 - connect_tcp.started host='login.microsoftonline.com' port=443 local_address=None timeout=5.0 socket_options=None
2024-01-24T10:16:07.064571449Z 2024-01-24 10:16:07 - connect_tcp.complete return_value=<httpcore._backends.anyio.AnyIOStream object at 0x7eea91f7ac10>
2024-01-24T10:16:07.064660998Z 2024-01-24 10:16:07 - start_tls.started ssl_context=<ssl.SSLContext object at 0x7eea91ddfc80> server_hostname='login.microsoftonline.com' timeout=5.0
2024-01-24T10:16:07.088082821Z 2024-01-24 10:16:07 - start_tls.complete return_value=<httpcore._backends.anyio.AnyIOStream object at 0x7eea91e03810>
2024-01-24T10:16:07.088632742Z 2024-01-24 10:16:07 - send_request_headers.started request=<Request [b'POST']>
2024-01-24T10:16:07.089092005Z 2024-01-24 10:16:07 - send_request_headers.complete
2024-01-24T10:16:07.089124546Z 2024-01-24 10:16:07 - send_request_body.started request=<Request [b'POST']>
2024-01-24T10:16:07.089195880Z 2024-01-24 10:16:07 - send_request_body.complete
2024-01-24T10:16:07.089214054Z 2024-01-24 10:16:07 - receive_response_headers.started request=<Request [b'POST']>
2024-01-24T10:16:07.211936736Z 2024-01-24 10:16:07 - receive_response_headers.complete return_value=...
[...]

PR #1044, opened 2024-05-31
Add UVICORN_TIMEOUT_KEEP_ALIVE to fix Invalid HTTP request received.

hayescode:
Me and my users have been experiencing a `Invalid HTTP request received` periodically. After some debugging i narrowed it down to uvicorn and found [this suggested fix ](https://github.com/encode/uvicorn/issues/441#issuecomment-535652028) from someone also using Async/FastAPI. I only get this error when deployed to Azure App Service, like many of the commenters.

This PR gives users the option via an environment variable `UVICORN_TIMEOUT_KEEP_ALIVE` to adjust to 0 like recommended. The default is 5 which is the current default value used.

### DIFF
--- a/backend/chainlit/cli/__init__.py
+++ b/backend/chainlit/cli/__init__.py
@@ -41,6 +41,9 @@ def run_chainlit(target: str):
     ws_per_message_deflate_env = os.environ.get(
         "UVICORN_WS_PER_MESSAGE_DEFLATE", "true"
     )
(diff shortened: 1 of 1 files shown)


hayescode:
Closed because @jdb78 seems to have figured this out 4 months ago. Thanks @jdb78 !

https://github.com/Chainlit/chainlit/issues/691

PR #1051, opened 2024-06-03
Prevent the 'Message' components from unexpectedly re-rendering while receiving tokens

0xez:
While receiving tokens from the server, the  'updateMessageContentById' function will create new objects for all the messages having 'steps' member, which will modify the references values of the messages and cause all the 'Message' components based on those messages to rerender. If we have a long conversations, like 100 rounds, this may cause serious performance problem.

This patch tried to fix it by checking if the messages really changed and only creating new objects for changed ones.

Meanwhile, given that only one message need to be updated and for most of the cases the last one, this patch also make some performance optimization by iterating from the last element and break the loop when finding matched one.


willydouhard:
I think we fixed this in another PR!

PR #1056, opened 2024-06-04
WIP: Auth based on signed token taken from the URL

nileshtrivedi:
This adds a new auth mechanism which takes a signed/untamperable token from the URL. This supports the implementations where Chainlit app is embedded in an iframe in a container app where seamless auth experience is desirable but it is not possible to send a custom HTTP header. The container app and user's chainlit app would share a secret. Container app would sign a token that contains some kind of user-identifier, and the chainlit script would resolve it to identify the user in a callback function.

[Here is the relevant thread](https://discord.com/channels/1088038867602526210/1126876648839598130/1247459822971125802) in Discord.

This is not a complete PR as it is missing the frontend and documentation changes. Also, I'm not very familiar with Chainlit codebase so might have missed something important. Please don't merge this just yet.


sandangel:
Hi @nileshtrivedi , I also have the same use case. I wonder how can I contribute?

sandangel:
Hi @tpatel , can you share chainlit team opinions on this?

sandangel:
Hmm, I'm thinking of using header auth combining with custom script loading. So in custom js script, we check if it's inside an iframe, then we send a request with header received from the parent window (using postMessage API). that header auth callback then will validate the token and create a user. Not sure how the frontend will know that the user is authenticated in that case.

nileshtrivedi:
@sandangel I think it will not require custom script loading. The first request (the `src` used for the iframe) itself can have the signed token, allowing chainlit server to authenticate the user and set the session cookie.

I do want to hear core team's opinion before proceeding with this approach. Do note that recent Chainlit release allows [mounting it as a FastAPI sub-app](https://docs.chainlit.io/integrations/fastapi) which is my preferred approach and thus makes this PR less critical.

sandangel:
Hi @nileshtrivedi thanks for sharing that. In that case, how will you handle the user authenticated state on frontend? I assume that you will use the default chainlit frontend right?

nileshtrivedi:
I am assuming that once the first page is served with a valid session cookie, frontend will show the right auth state.

sandangel:
Oh, so the frontend is getting the authentication status via the cookie? From what I read in the frontend code, it will first get auth/config, then based on the auth method you configured, it will send the request (auth/header) or showing login page (password/oauth). 

There is also a CHAINLIT_CUSTOM_AUTH env var, but not sure how can we utilize it. In this case, which auth method do you plan to use? Are you going to create a custom endpoint, and then use that endpoint for `src` of iframe? But we still need to somehow send the generated access token back to frontend and have the @chainlit/react-client set that access token when making API requests, right?

nileshtrivedi:
> rom what I read in the frontend code, it will first get auth/config, then based on the auth method you configured, it will send the request (auth/header) 

@sandangel Interesting! Yeah, this could be trickier than I thought.
And FWIW, I might need this feature after all, because NextJS has some annoying bugs when sending custom request headers via proxy.

dokterbob:
This is a great feature idea!

It might also help towards our use case of being able to create 'share links' allowing new users to go straight into a conversation.

However, token auth is tricky and having chainlit users roll their own might expose their users to more risk than necessary. Particularly, generating and consuming signed tokens is non-trivial. 

Perhaps we want a (default) implementation, supporting your exact business case, rather than a hook?

It's very clearly scoped; one app generates a token server-side based on a secret and a timestamp, chainlit validates the token and timestamp.

nileshtrivedi:
@dokterbob ChainLit already [supports token auth](https://docs.chainlit.io/authentication/header) - but as a request header, not from URL. But JWT tokens tend to be a bit too long for including in the URL.

The use-case is to integrate Chainlit in an `<iframe>` in a host app and have the user be seamlessly authenticated.

Completing this turned out to be not trivial. ChainLit uses LocalStorage, not cookies, for storing the session token. This means that it is not possible to handle auth on the first request itself and and log the user in. Instead, the auth API call is made after the React frontend is loaded. That code is not preserving the URL so even if a token is provided in the <iframe> URL, it is lost by the time login API gets called.

Internally, we have moved away from ChainLit because of this and other limitations. Discarding and closing this.

sandangel:
hi @nileshtrivedi , may I know what is the new tool you are switching to?

nileshtrivedi:
We ended up building our own React component for the chat UI which talks to the agent on the server (with database persistence, background jobs etc) over a JSON API. Separating the UI solves all the issues around embedding, authentication, long-running agent tasks etc.

dokterbob:
> @dokterbob ChainLit already [recommends token auth](https://docs.chainlit.io/authentication/header) - but as a request header, not from URL. JWT tokens tend to be a bit too long for including in the URL so for URL, custom token auth handling becomes necessary.
> 
> The use-case is to integrate Chainlit in an `<iframe>` in a host app and have the user be seamlessly authenticated.
> 
> Completing this turned out to be not trivial. ChainLit uses LocalStorage, not cookies, for storing the session token. This means that it is not possible to handle auth on the first request itself and and log the user in. Instead, the auth API call is made after the React frontend is loaded. That code is not preserving the URL so even if a token is provided in the <iframe> URL, it is lost by the time login API gets called.
> 
> Internally, we have moved away from ChainLit because of this and other limitations. Discarding and closing this.

The problem with token based auth like this (in the URL) is it easily makes an app susceptible to replay attacks. If I remember well, the OAuth2 spec has a use case like this as well, and takes care to mention some of these caveats.

It's a pity you 'moved away', we're really looking to bundle efforts to bring Chainlit on track towards production-grade FOSS. But I understand, it will take some 😅 and 💪🏼!

nileshtrivedi:
@dokterbob The goal is to be able to use ChainLit in existing apps. One way, this could be achieved is if ChainLit provides its UI not as a webpage, but as a **component** (React or Custom Element). This would give the host application sufficient hooks to customize behaviour like auth strategy. 

There are projects like https://deepchat.dev building in that direction. But it's a thought to consider.

PR #1057, opened 2024-06-04
Added env `CHAINLIT_DEBUG` on run_chainlit(target), default to False.

davizucon:
Added env `CHAINLIT_DEBUG`(default False) when run_chainlit to enable to set log debug mode.

### DIFF
--- a/backend/chainlit/cli/__init__.py
+++ b/backend/chainlit/cli/__init__.py
@@ -37,6 +37,7 @@ def cli():
 def run_chainlit(target: str):
     host = os.environ.get("CHAINLIT_HOST", DEFAULT_HOST)
     port = int(os.environ.get("CHAINLIT_PORT", DEFAULT_PORT))
+    debug = bool(os.environ.get("CHAINLIT_DEBUG", False))
 
     ssl_certfile = os.environ.get("CHAINLIT_SSL_CERT", None)
     ssl_keyfile = os.environ.get("CHAINLIT_SSL_KEY", None)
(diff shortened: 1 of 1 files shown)


willydouhard:
Isn't it already done by the CLI run command? Unless you are not starting Chainlit with the CLI?

kevinwmerritt:
FWIW I do a combination of setting the env and setting config directly:

Probably would be nice use one approach or the other consistently.

```
from chainlit.config import config
from chainlit.cli import run_chainlit

...

# Specify a different port to run the server on
os.environ["CHAINLIT_PORT"] = str(settings.get("port"))

# Will prevent auto opening the app in the browser
config.run.headless = True

if settings.get("cluster") == "dev":
    # Reload the app when the module changes
    config.run.watch = True
    # Set the log level to debug
    config.run.debug = True
    
run_chainlit("my_chainlit_app.py")
```

davizucon:
Hey , thanks for review !

When debugging through the IDE `run_chainlit`, _some?_ environment variables are not read, hence this PR.

However, when analyzing the code further, as suggested by @kevinwmerritt , it would really be better if execution via both CLI and `run_chainlit` maintained the same behavior.

Maybe something to look at here: [config.py#L484](https://github.com/Chainlit/chainlit/blob/29eda96ccf8eb4eccd390a4e5d132421733a2a67/backend/chainlit/config.py#L484.)

I will close this PR for a better evaluation.

Thanks !

PR #1137, opened 2024-07-16
Compatibility of data layer with sqlite

a-romero:
Updated to support databases that natively return JSONB objects as strings, such as sqlite.


dokterbob:
Solved in https://github.com/Chainlit/chainlit/pull/1319

PR #1155, opened 2024-07-23
fix: favicon path not correct when root path is set

### LINKED ISSUE #1133: When starting with root-path, favicon cannot be loaded
**Describe the bug**
When the app is launched with root-path, the favicon cannot be loaded. When the app is launched without root-path, the favicon can be displayed normally.

**To Reproduce**
Add a favicon image under /public. Use `chainlit run app4.py --port 8081` and the favicon is displayed normally. Use `chainlit run app4.py --port 8081 --root-path /chat` and the favicon will not be displayed.

**Expected behavior**
favicon is displayed.

**Screenshots**
When the app is started with the root-path, I noticed that when the web page starts, it made a request to http://localhost:8081/chat/chat/favicon to get the favicon. I think it should be http://localhost:8081/chat/favicon
PixPin_2024-07-15_02-11-04

- OS: win10
- Browser [chrome, edge]
- Version [e.g. 1.1.306]

sandangel:

Fixed #1133


sandangel:
https://github.com/Chainlit/chainlit/pull/1155/files#diff-9a73e67e163edc3fae6839bf2bcc8c6ab650624eff6668502388a7a86b7ece78L309

at this line, we already replace all the href and src, so won't need to change root path for favicon.

sandangel:
@tpatel can you help me have a look at this PR?

PR #1160, opened 2024-07-24
chore: update version for fastapi, starlette and uvicorn

jiahao1553:
Update versions for fastapi, starlette and uvicorn so that integration with r2r package can be done. Currently, there are dependency conflicts:
```bash
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
chainlit 1.1.306 requires fastapi<0.111.0,>=0.110.1, but you have fastapi 0.109.2 which is incompatible.
chainlit 1.1.306 requires starlette<0.38.0,>=0.37.2, but you have starlette 0.36.3 which is incompatible.
chainlit 1.1.306 requires uvicorn<0.26.0,>=0.25.0, but you have uvicorn 0.27.1 which is incompatible.
```


dokterbob:
Thanks for the contrib!

However, we cannot automatically assume that the latest version of dependencies will automatically work as there are [breaking changes](https://fastapi.tiangolo.com/release-notes/#breaking-changes). 

Hence, it's recommended to [pin the fastapi version](https://fastapi.tiangolo.com/deployment/versions/#pin-your-fastapi-version). 

For reliable use as part of other FastAPI-requiring packages, the only way is [testing](https://fastapi.tiangolo.com/tutorial/testing/) of our backend -- but we're really not there, yet. (If we were, we could use tox to run tests for each version of FastAPI so we can provide proper support.)

Would it work for you to fork the other package instead and upgrading deps there rather than _downgrading_ FastAPI here (and potentially introducing security issues).

We might also be open to upgrading fastapi and other core deps to the latest version (but then we'll pin it there).

A start of testing infra would also be ❤️❤️❤️ 

For now, I'm closing this but I'm happy to hear your thoughts on my suggestions.

PR #1162, opened 2024-07-24
Add window.postMessage support.

mauanga:
## Description

This update enhances the functionality of the Chainlit application when it is embedded within another web application via an iframe. By enabling message exchange between the parent and child windows, the application can now support extended features. Some examples of the use cases would be; allowing users to select datasets from their dashboard, supply saved prompts, and more.

## Changes

1. Added an event listener listening for window `message` events in `AppWrapper.tsx`.
2. Added a callback to emit the `window_message` event in `useChatInteract.ts`
3. Added a new decorator `on_window_message` to `__init__.py`
4. Added the callback for the decorator in `config.py`
5. Added a socket io event `window_message` in `socket.py`


dokterbob:
Interesting feature!

 Code looks clean, on quick review.
 
@mauanga Is there any chance you could add a test for this feature in https://github.com/Chainlit/chainlit/tree/main/cypress/e2e?

mauanga:
@dokterbob Thanks for taking the time to review the PR.

I’ve made a small update to include support for 2-way messaging, which now allows messages to also be sent from the backend to a listening host window.

I'll start working on the cypress e2e test and will add it to the PR soon.

jfperusse-bhvr:
@mauanga Just tested this PR and it's great! I would suggest initializing the context inside `window_message(sid, data)` so we can use cl when receiving the message. As a test, with this change, I was able to remove actions from the user interface after receiving the message in my iframe (context initialization is based on the other decorators):

```py
@sio.on("window_message")
async def window_message(sid, data):
    """Handle a message send by the host window."""
    session = WebsocketSession.require(sid)
    context = init_ws_context(session)

    await context.emitter.task_start()

    if config.code.on_window_message:
        try:
            await config.code.on_window_message(data)
        except asyncio.CancelledError:
            pass
        finally:
            await context.emitter.task_end()
```

backend/chainlit/emitter.py
jfperusse-bhvr, line 1: Missing from BaseChainlitEmitter:

```py
    async def send_window_message(self, data: Any):
        """Stub method to send custom data to the host window."""
        pass
```

Issue #1097, opened 2024-06-24
Allow changing thread/conversation names


trinhkiet0105:
I try to change thread Name on the chat UI but know how to

The thread name is default to take the first User message as the thread name. I want to customize thread name after the first assistant response (like chatGPT)

AidanShipperley:
I'd also love to know if there is an actual way to do this properly. 

I was able to find a way to manually do this, however it doesn't update the UI until the user refreshes the page. Regardless, it may still be useful to you or the maintainers:

## Make a Function to Create Chat Title
```python
async def generate_conv_summary(conv_messages):
    
    client = AsyncOpenAI()
    conv = ""
    for message in conv_messages:
        if message["role"] == "user":
            conv += f"User: {message['content']}\n"
        elif message["role"] == "assistant":
            conv += f"Assistant: {message['content']}\n"

    messages = [
        {"role": "system",
         "content": "Act as a conversation summarization engine designed to create a short title for a conversation between a user and an AI assistant. Given a conversation, you will generate a title for that conversation in 3 (three) words or less. The title should be a concise summary of the conversation, capturing the main topic or theme of the conversation."},
        {"role": "system", "content": conv}
    ]

    stream = await client.chat.completions.create(
            messages=messages, stream=True, model="gpt-4o"
    )

    return stream
```

## Add Functionality to Main Function
```python
# import chainlit.data as cl_data
@cl.on_message
async def main(message: cl.Message):
    
    message_history = cl.user_session.get("message_history")
    # Custom code to identify if this is the first message in a thread
    first_msg = False
    if len([msg for msg in message_history if msg.get('role') == 'user']) == 1:
        first_msg = True
        
    
    # Generate and give msg to user as usual
    # ...
    # message_history.append({"role": "assistant", "content": generated_msg.content})
    ###
    
    # Once assistant message is generated, generate chat title and update the thread name
    if first_msg:
        chat_title = ""
        conv_summary_streamer = await generate_conv_summary(message_history)
        async for part in conv_summary_streamer:
            if token := part.choices[0].delta.content or "":
                chat_title += token

        chat_title = chat_title.title().replace(".", "")
        await cl_data._data_layer.update_thread(message.thread_id, name=chat_title)
```
---
Hopefully this helps, I would love to know of a way to update the thread name and instantly have it render onto the UI.

trinhkiet0105:
> I'd also love to know if there is an actual way to do this properly.
> 
[...]

PR #1174, opened 2024-07-29
Eliminate auto capitalization of thread names

gm3197:
Removing auto-capitalization of thread names because it is not necessary and forces this design decision on developers.

### DIFF
--- a/frontend/src/components/organisms/sidebar/threadHistory/ThreadList.tsx
+++ b/frontend/src/components/organisms/sidebar/threadHistory/ThreadList.tsx
@@ -204,7 +204,7 @@ const ThreadList = ({
                             textOverflow: 'ellipsis'
                           }}
                         >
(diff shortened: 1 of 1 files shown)


dokterbob:
Hi @gm3197, thanks for your patch!

As it stands, applying it would unexpectedly change the UX for pre-existing users, while we do not (yet) have a clear-cut/well-documented way of directly controlling thread names. If there were, developers can just implement capitalisation on their own and we could document this as a breaking change, giving an example of how to do this.

Is there _any_ chance you'd be able to have a look at this ticket? https://github.com/Chainlit/chainlit/issues/1097

With a clear and sensible API to be able to control chat titles (in a way where it's not glitching, e.g. changing erratically), merging your patch would be a no-brainer.

Curious to hear what @willydouhard images here.

willydouhard:
this has been released in the `2.0.0` version!

PR #1182, opened 2024-08-02
fix callback regression

adhil-tinvio:


willydouhard (approved):

willydouhard:
Nice catch! Thank you for fixing it 👍

PR #1196, opened 2024-08-08
Jinja2 index template

dokterbob:
Jinja2 (rather than custom template) in index view. Would allow future overrides of headers, including arbitrary JS and meta tags (analytics, ads, Twitter cards, whatever) which we'll submit as a second patch.

Right now, the backend reads a file from the frontend dist folder and the `index.html` from the frontend is not valid HTML anymore.

Many creds to @dhruvsyos who did the heavy lifting on this one.

### DIFF
--- a/backend/chainlit/config.py
+++ b/backend/chainlit/config.py
@@ -124,6 +124,9 @@ cot = "full"
 # Specify a custom meta image url.
 # custom_meta_image_url = "https://chainlit-cloud.s3.eu-west-3.amazonaws.com/logo/chainlit_banner.png"
 
+# Specify a custom meta url.
+# custom_url = ""
+
 # Specify a custom build directory for the frontend.
 # This can be used to customize the frontend code.
 # Be careful: If this is a relative path, it should not start with a slash.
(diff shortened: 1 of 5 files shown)


synergyai-os:
Exciting! Can't wait for this to be approved and add tracking into my chainlit app with amplitude

dokterbob:
@willydouhard Curious to hear your opinion on the question from above, as it's a non-trivial design decision:

Right now, the backend reads a file from the frontend folder and the index.html from the frontend is not valid HTML anymore.

Question to the maintainer: should we keep a default index.html in the frontend (as valid HTML), then have a Jinja2 template in the backend which allows overrides?

dokterbob:
Please don't review yet, I'd still like to add unit tests and think really well about where and how to place this index.html -- the current strategy seems rather inelegant.

ahmadmayahi:
Any updates on this PR?

dokterbob:
> Any updates on this PR?

I'm working on test coverage for unit test views and considering to properly implement this following #1268.
In addition, I want to make it possible for people to install chainlit as `--editable` through #1224.

PR #1192, opened 2024-08-07
Bump Python requirement to 3.9. Solves #1189.

dokterbob:
See https://github.com/Chainlit/chainlit/issues/1189

PR #1217, opened 2024-08-14
Black isort as dev deps in pyproject.toml

### LINKED ISSUE #1194: Required Black and isort code formatters not in pyproject.toml
**Describe the bug**
The pre-commit hook calls `black`, but it's not mentioned as a (dev) dependency in pyproject.toml. They are explicitly mentioned in the contribution instructions.

Adding this to dev dependencies (rather than main deps) would facilitate running them on CI in addition to making dev easier and more consistent (together with #1190).

**Observations**
There's currently a `tests` and a `mypy` group but not a `dev` group.
There is no group containing these code formatters.

What is the rationale behind this, if there is any? Is there a use case when mypy is required but tests aren't, and vice versa? Are there use cases where black and/or isort are not required?

**Suggestion**
A common practice is to have a single `dev` group with all dev-related requirements, used both for local dev and remote CI. This is  conceptually easier.

dokterbob:
Add dev deps group and add isort and black there. Closes #1194 

Forked from #1192 as to allow for smooth merge/prevent merge conflicts.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -25,7 +25,7 @@ I've copy/pasted the whole document there, without the previous two headings.
 
 ### Requirements
 
-1. Python >= `3.8`
+1. Python >= `3.9`
 2. Poetry ([See how to install](https://python-poetry.org/docs/#installation))
 3. NodeJS >= `16` ([See how to install](https://nodejs.org/en/download))
 4. Pnpm ([See how to install](https://pnpm.io/installation))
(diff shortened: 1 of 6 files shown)


willydouhard (approved):

dokterbob:
@willydouhard If I merge this as it stands, I am _forced_ to squash even though 'black and isort' (this issue) has _nothing_ to do with the Python 3.9 version bump.

Please either:
1. Enable normal merge or;
2. Approve #1192 so I can merge that first.

As to avoid history to become messy, making tracing of root causes for bugs harder.

PR #1218, opened 2024-08-14
fix: edit_message in correct place in config.toml

### LINKED ISSUE #1204: `edit_message` in wrong place in `config.toml`
**Describe the bug**
If you run chainlit with an empty `config.toml` the `edit_message` key is placed below `[features.audio]`, however it only works when it is inside the `[features]` block.

**To Reproduce**

1. Run chainlit without a `config.toml` to generate the file
2. The generated `config.toml` has the line `edit_message = true`, change it to `edit_message = false`
3. Run chainlit again
4. Messages are still editable

**Expected behavior**
Messages should not be editable when `edit_message = false`.

**Additional context**
Moving `edit_message = false` up to the `[features]` block disables editing messages as expected.

EWouters:
Fixes #1204.

Also adds a comment for the setting:

> \# Allow users to edit their own messages

### DIFF
--- a/backend/chainlit/config.py
+++ b/backend/chainlit/config.py
@@ -70,6 +70,9 @@ latex = false
 # Automatically tag threads with the current chat profile (if a chat profile is used)
 auto_tag_thread = true
 
+# Allow users to edit their own messages
+edit_message = true
+
 # Authorize users to spontaneously upload files with messages
 [features.spontaneous_file_upload]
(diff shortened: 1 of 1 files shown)


dokterbob (approved):
Looks good to me! Replicated the config rendering issue and your cause analysis seems correct -- thanks for clear explanation of the problem!

PR #1220, opened 2024-08-15
fix: mount chainlit app to correct path

sandangel:

### DIFF
--- a/backend/chainlit/utils.py
+++ b/backend/chainlit/utils.py
@@ -130,4 +130,4 @@ def mount_chainlit(app: FastAPI, target: str, path="/chainlit"):
 
     ensure_jwt_secret()
 
-    app.mount("/", chainlit_app)
+    app.mount(path, chainlit_app)


willydouhard (approved):

willydouhard:
Actually I don't think this is required. The chainlit app is already [prefixing](https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/server.py#L195) the router with the root path

dokterbob:
@sandangel Thanks for the contrib San, could you describe the specific issue you're addressing?

I think @willydouhard has a good point; the the app already listens on the mount point through the root path configuration option: https://github.com/Chainlit/chainlit/blob/0d043cbdce1fd577e57d4526e8eae50700b9b157/backend/chainlit/server.py#L178

sandangel:
@dokterbob I have an existing fastapi app and it mount another sub app on a different path. After mounting this chainlit app my subapp that is mounting on another path is not working anymore. This will fix that issue. I tested it creating my own mount_chainlit function and it's working.

sandangel:
Okay, turned out that if I mount to `app.mount(path, chainlit_app)` then the rover api path will become: `/root_path/root_path/auth/config` instead of `/root_path/auth/config`. I wonder if we have a better way to handle this.

dokterbob:
> Okay, turned out that if I mount to `app.mount(path, chainlit_app)` then the chainlit api path will become: `/root_path/root_path/auth/config` instead of `/root_path/auth/config`. I wonder if we have a better way to handle this.

Yeah that's the issue I anticipated having looked at the code.

I have found, however, that mounting chainlit _after_ defining my app endpoints avoids the problem you (and I!) ran into.

I feel that being able to `app.mount()` chainlit (so mounting it as a FastAPI/Starlette app within another FastAPI/Startlette) would be preferable as it would be a much better developer experience but perhaps there are reasons why it has been done in this way (maybe because of the way websockets work?). 

@willydouhard Are there any reasons why you chose to implement `mount_chainlit()` over exporting a Starlette `app` for people to `app.mount()` to a subpath?

dokterbob:
I'm closing this in favour of my issue where we can discuss steps towards implementing a more structural solution.

As for now, keep in mind that that `mount_chainlit()` should be called _after_ any other endpoints are set up.

sandangel:
Thanks, I will go with the workaround for now and wait for the proper fix.

PR #1227, opened 2024-08-17
Remove value equals empty input.tsx

mosheber:
Allow the chat string to be empty. This enables a scenario where a user wishes to only submit attachments, for example.


dokterbob:
Seems to yield build errors, could you please address them @mosheber ?

dokterbob:
Superseded by ##1261.

Issue #1213, opened 2024-08-14
Make Request object available in `on_chat_start()`


dokterbob:
**Is your feature request related to a problem?**
This would address, in one go:
https://github.com/Chainlit/chainlit/issues/144
https://github.com/Chainlit/chainlit/issues/964
https://github.com/Chainlit/chainlit/pull/1002
https://github.com/Chainlit/chainlit/issues/1103
https://github.com/Chainlit/chainlit/issues/1140

It might also work towards solving:
https://github.com/Chainlit/chainlit/pull/1056

* It should allow devs to put whatever request data they need in `user_session`.
* It allows using query parameters, cookies and arbitrary HTTP headers. 
* We might be able to deprecate things like `languages` and other headers copied in verbatim.

**Describe the solution you'd like**
Accept a `request: starlette.requests.Request` parameter to `on_chat_start()`.

**Describe alternatives you've considered**
See issues/PR above. The proposed solution is more generic, requires less maintenance (e.g. no specific support for various use cases required).

Open question: are there any reasons (security or otherwise) not to do this?

dokterbob:
@willydouhard Would love your ideas on this one, should be really easy to implement.

Are there any security or other implications?

cta2106:
Much needed feature imo. Only thing stopping me from deploying Chainlit for a number of clients at the enterprise level!

comediansd:
I need this very urgend to send user links for special training bots

nethi:
@willydouhard  would this PR  https://github.com/Chainlit/chainlit/pull/1260 help ? For HTTP hooks, I am using ContextVar like this in server.py

```python
# Create a context variable to store the request
request_context_var: ContextVar[Request] = ContextVar("request_context_var")

def get_current_request() -> Request:
    return request_context_var.get()

class RequestContextMiddleware(BaseHTTPMiddleware):
    async def dispatch(self, request: Request, call_next):
        # Set the request in the context variable
        request_context_var.set(request)
        # Process the request and get the response
        response = await call_next(request)
        return response
    
# Add request contextvar middleware to the FastAPI application
app.add_middleware(RequestContextMiddleware)
```

dokterbob:
> @willydouhard would this PR #1260 help ? For HTTP hooks, I am using ContextVar like this in server.py
> 
> ```python
> # Create a context variable to store the request
> request_context_var: ContextVar[Request] = ContextVar("request_context_var")
> 
> def get_current_request() -> Request:
[...]

PR #1234, opened 2024-08-19
Support browser query parameters on chat start

qtangs:


dokterbob:
@qtangs 👀 

Please have a look at #1213 -- we're looking to see if we can have the entire (initiating) request object available, it might be a more sustainable approach. 

A PR towards that would be greatly appreciated!

Issue #1219, opened 2024-08-15
Logout Button Causing Issue w/ Header Auth


marty-sullivan:
**Describe the bug**
It appears like header-based login might break after clicking the logout button. I will admit, I'm doing something a little different than what is designed, as is described here:

https://github.com/Chainlit/chainlit/issues/144#issuecomment-2227005294

Given that though, I'm seeing some very odd behavior as described below. It doesn't seem like it should matter that I'm using something from the referrer header vs some other header.

**To Reproduce**

The logs show the following...

1. User authenticates using the header method I use, linked above
2. User clicks on logout button
3. User tries to log in again by being directed to `/login?token=...`
4. User is redirected to `/auth/header` and `header_auth_callback` succeeds (returns valid `User`)
5. User is redirected to chat
6. User is immediately redirected back to `/login` and then back to `/auth/header`
7. `header_auth_callback` fails (returns `None` because there is no token query param)
8. User is redirected back to `/auth/header` a second time
9. `header_auth_callback` fails (returns `None` because)

**Expected behavior**
I would expect that the user is successfully logged in on the first redirect to `/auth/header` but it seems like after the user manually logs out, it somehow makes it so that the header login flow is failing to set the User for the session.

I'm certainly willing to accept that what I'm doing isn't supported, however, this behavior is not something I've run into and only seems to happen after the user manually logs out. It's almost like a cookie or session storage item is no longer able to be set by the header login flow.

I have tried clearing browser cookies and data and all of that, it doesn't seem to help once the user clicks that button. Very mysterious.

marty-sullivan:
Some additional tests show that this problem does not seem to happen with v1.1.306

dokterbob:
Is this perhaps related to #1207?

marty-sullivan:
@dokterbob I don't think so, I was just using the basic header auth, not oauth

willydouhard:
Is this still the case in `2.0.0`?

pratipal-dev:
I am still facing the same issue, I am on 2.5.5. I am using [this](https://github.com/Chainlit/chainlit/issues/144#issuecomment-2227005294) method to pass the access_token into an iframe. Even after returning a valid user it keeps on calling header_auth_callback and second time it doesn't have those query string parameters so it fails. This is prevent me from using chainlit in production.

Any help would be appreciated.

PR #1239, opened 2024-08-19
Support browser query parameters on chat start

qtangs:
Here's my attempt to address this much-discussed issue: [**Make browser query parameters available in @on_chat_start**](https://github.com/Chainlit/chainlit/issues/144)

Also fix this bug: [**Logout Button Causing Issue w/ Header Auth**](https://github.com/Chainlit/chainlit/issues/1219)

Here are the changes:
- [x] Move authentication check from top level AppWrapper to Page (this removes errornous redirection to login, as reported by https://github.com/Chainlit/chainlit/issues/1219)
- [x] Add logic to call header auth in Page if user is not yet authenticated
- [x] Update both Login and Page to pass query parameters when redirecting between each other (unauthenticated Page -> Login -> Page after authentication)
- [x] Update header_auth and password_auth tests to add the scenario where a query parameter is read and used on chat start

I have tested on my app: https://2in.ai

### DIFF
--- a/backend/poetry.lock
+++ b/backend/poetry.lock
@@ -13,98 +13,98 @@ files = [
 
 [[package]]
 name = "aiohappyeyeballs"
-version = "2.3.5"
+version = "2.3.7"
 description = "Happy Eyeballs for asyncio"
 optional = false
 python-versions = ">=3.8"
 files = [
-    {file = "aiohappyeyeballs-2.3.5-py3-none-any.whl", hash = "sha256:4d6dea59215537dbc746e93e779caea8178c866856a721c9c660d7a5a7b8be03"},
-    {file = "aiohappyeyeballs-2.3.5.tar.gz", hash = "sha256:6fa48b9f1317254f122a07a131a86b71ca6946ca989ce6326fff54a99a920105"},
+    {file = "aiohappyeyeballs-2.3.7-py3-none-any.whl", hash = "sha256:337ce4dc0e99eb697c3c5a77d6cb3c52925824d9a67ac0dea7c55b8a2d60b222"},
+    {file = "aiohappyeyeballs-2.3.7.tar.gz", hash = "sha256:e794cd29ba6a14078092984e43688212a19081de3a73b6796c2fdeb3706dd6ce"},
 ]
 
 [[package]]
 name = "aiohttp"
-version = "3.10.3"
+version = "3.10.4"
 description = "Async http client/server framework (asyncio)"
 optional = false
 python-versions = ">=3.8"
 files = [
-    {file = "aiohttp-3.10.3-cp310-cp310-macosx_10_9_universal2.whl", hash = "sha256:cc36cbdedf6f259371dbbbcaae5bb0e95b879bc501668ab6306af867577eb5db"},
-    {file = "aiohttp-3.10.3-cp310-cp310-macosx_10_9_x86_64.whl", hash = "sha256:85466b5a695c2a7db13eb2c200af552d13e6a9313d7fa92e4ffe04a2c0ea74c1"},
-    {file = "aiohttp-3.10.3-cp310-cp310-macosx_11_0_arm64.whl", hash = "sha256:71bb1d97bfe7e6726267cea169fdf5df7658831bb68ec02c9c6b9f3511e108bb"},
-    {file = "aiohttp-3.10.3-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:baec1eb274f78b2de54471fc4c69ecbea4275965eab4b556ef7a7698dee18bf2"},
-    {file = "aiohttp-3.10.3-cp310-cp310-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl", hash = "sha256:13031e7ec1188274bad243255c328cc3019e36a5a907978501256000d57a7201"},
-    {file = "aiohttp-3.10.3-cp310-cp310-manylinux_2_17_s390x.manylinux2014_s390x.whl", hash = "sha256:2bbc55a964b8eecb341e492ae91c3bd0848324d313e1e71a27e3d96e6ee7e8e8"},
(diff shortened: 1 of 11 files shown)


dokterbob:
Hi @qtangs, thanks a bunch for this PR!

I fully agree that having access to query parameters is a *hot* topic which many users, myself included, are aching for. In fact, I'd like to have this in chainlit in weeks rather than months.

However, I am not sure if your proposed patch is the way to go.

First of all, as I mentioned in #1213 , providing a full `Request` object (either based on the original index request, or based on the websocket connection) yields a similar amount of code but enables many more use cases. Instead, if we accept your current proposal, we would break developer experience once we switch to supporting the full Request.

Secondly, having read your patch, you seem to be using the HTTP referrer header rather than the actual URL. Most notably, the referrer header is optional and not validated server-side, whereas, the request URL is validated by the server.

They represent different things and can lead to unexpected behaviour and possibly, security issues! Specifically in your case, your patch seems to contain a lot of code which related specifically to authentication rather than mere access to the query header.

Any referrer based authentication is a classic foot gun for security and not the way to go. Perhaps a resource like this might provide background on it: https://web.dev/articles/referrer-best-practices

I also see you included fixes for one pre-existing issue and some additional issues as part of this PR. That's great!

As it seems we still have some architectural decisions to take on the query params, perhaps it would be easier for us to have them as part of separate pull requests.

Any chance you're able to pry the parts not related to query params out and fire (preferably) multiple PR's, one for each separate issue, with enough information for us to replicate the issue while we review your fixes? 

I know it's a lot to ask for, but reviewing all these issues at the same time makes it harder to do a thorough review and especially with auth stuff that can lead to dire consequences. Thanks for understanding!

qtangs:
Hey @dokterbob, thanks for the super quick response to this. Really appreciate your taking time to look into these important issues.

Let me try to respond as best as I can.

I think you brought some very good points and I really like the Request object idea, in fact it'd be very useful to have it in many other handlers besides `on_chat_start`.
And I admit I'm trying to address a few things together in this PR:

1. Pass the query parameters around, between the paths `/`, `/login` and also `/env`, as users are redirected upon failing/passing authentication or env.
I think for this one, it doesn't have conflict with https://github.com/Chainlit/chainlit/issues/1213 materializes, as the former is about passing data between different requests while your idea is about a single request, unless you want the Request object to contain data from multiple linked browser requests (all data from `/`, `/login`, `/env` are added to 1 Request object?).

2. Fix the bug https://github.com/Chainlit/chainlit/issues/1219
Agree that this may benefit from a separate PR.

3. Use `http_referer` for on_chat_start
This is optional, dependent on the user (dev) of chainlit, it's not enforced by chainlit so while the security concern is there, it should be handled by the user. Once https://github.com/Chainlit/chainlit/issues/1213 is completed, anyone can easily update `on_chat_start` to use it.

Happy to discuss further.

dokterbob:
Hi @qtangs,

This is what I ❤️ about open source. :)

> 1. Pass the query parameters around, between the paths `/`, `/login` and also `/env`, as users are redirected upon failing/passing authentication or env.
>    I think for this one, it doesn't have conflict with [Logout Button Causing Issue w/ Header Auth #1219](https://github.com/Chainlit/chainlit/issues/1219), as the former is about passing data between different requests while your idea is about a single request, unless you want the Request object to contain data from multiple linked browser requests (all data from `/`, `/login`, `/env` are added to 1 Request object?).


Because the way chainlit is developed, as a websocket-based SPA app, the concept of a request is not so clearly defined. Rather, the websocket session (in addition to the data layer's session) is what's used to keep state. This is probably why a `Request` object wasn't originally made available.

However, as we're finding now, having access to headers, query parameters etc. is essential to allow dev's like ourselves to _go wild_ in terms of amazing features we want to implement.

We don't always have a well-defined `Request` but we do have a well-defined `user_session`. In addition, `on_chat_start()` _does_ always correspond to a HTTP request (namely, the one opening the websocket). So, passing a Request once to `on_chat_start()` would be valid, after which dev's can literally do whatever they want with the request and it's data, including but not limited to adding the entire thing to the `user_session`.

The auth issue is a bit of a trickier beast and, if I'm entirely honest, given the amount of issues related to auth and the _insane_ variety of auth options, my current stance (which we still need to align on amongst maintainers!) is that we should delegate _all_ auth to third-party libraries so we can focus on a silky smooth LLM-chat experience.

> 2. Fix the bug [Logout Button Causing Issue w/ Header Auth #1219](https://github.com/Chainlit/chainlit/issues/1219)
>    Agree that this may benefit from a separate PR.
If you would do this, this seems to be the most obvious way to include this chunk of your work at this point. Please, provide 
as clear as possible testing instructions, including, ideally, a test project with deps and minimal code showing both the issue and it's disappearance or, if you're _super_ kind and slightly crazy, a regression test in our e2e test suite.

> 3. Use `http_referer` for on_chat_start
>    This is optional, dependent on the user (dev) of chainlit, it's not enforced by chainlit so while the security concern is there, it should be handled by the user. Once [Make Request object available in `on_chat_start()` #1213](https://github.com/Chainlit/chainlit/issues/1213) is completed, anyone can easily update `on_chat_start` to use it.

But your patch _does_ add it to the tests. Tests are often used as a form of (authoritative, known to be working) example for other devs and (potentially) insecure examples are really something we should take care to stay clear of.

As for now, I will be closing the current PR in favour of a PR with an (optional) `request: Request` parameter in `on_chat_start()` as well as a separate PR for 2. I do hate to be 'rejecting' your work as I do appreciate great contribs.

To avoid that, in the future, I suggest using an issue to propose features to allow others to comment on it and to hold space to form a clearly crystallised concept before firing the PR -- that will provide a straight path to getting your code shipped out. (I hope to add this info to the contribution instructions once the process is a bit more crystallised itself. :p)

Thanks for understanding and please keep the good contrib coming!

qtangs:
@dokterbob I understand your decision. Will see what else can be done. For the moment, at least the code will be here as a reference for a quick workaround while waiting for the proper solution once Request object is made available. Rooting for this: :)
> to allow dev's like ourselves to go wild in terms of amazing features we want to implement.

Issue #1223, opened 2024-08-16
Export ASGI `app` allowing for `app.mount()` instead of `mount_chainlit()`


dokterbob:
**Is your feature request related to a problem? **
In order to include chainlit in a 3rd party FastAPI/ASGI/WSGI app, it's currently necessary to call `mount_chainlit()`.

As this is not a common way to expose apps for mounting in other apps, it leads to confusion amongst developers. See #1166 and #1220.

Specifically, the current implementation requires that `mount_chainlit()` is called after any other endpoints are defined. As chainlit would be mounted on a subpath, this is not intuitive.

The current implementation also makes it impossible to mount chainlit multiple times (e.g. with different configuration, on different paths). Note that the current suggestion will not directly enable this, but it seems a logical step in the direction of supporting multiple chainlits.

**Describe the solution you'd like**
Having a chainlit `app` which can either be imported from a module or dynamically generated with a factory function (`get_app()`), as an independent and self-contained ASGI-app.

This would include using the request path to derive the root path (rather than passing it hardcoded from `mount_chainlit()` as is currently done). This also seems a cleaner approach.

stephenrs:
While I don't have enough direct experience with asgi to comment on specifics, I'll give anything that makes chainlit less of a bully a thumbs up :) 

It seems the underlying original philosophical notion is that CL is going to be the only thing running on a server...it wants to do auth...it wants to terminate SSL...and if I'm reading this and the related issues correctly, it wants to control routing. So, "Yes" to anything that makes CL play more nicely with others.

In my case, since I didn't see any mention of Flask in the docs (and didn't get any tips in discord), I resorted to loading CL in an iframe from a Flask template. Ugly (but isolated).

tituslhy:
I'd like to upvote this issue please. After mounting my chainlit app on FastAPI I'm also not too sure how to invoke it from a FastAPI sub-application endpoint perspective (https://fastapi.tiangolo.com/advanced/sub-applications/#check-the-automatic-api-docs). But I have verified that running 0.0.0.0:port_number opens the standard Chainlit frontend.

stephenrs:
I really hope some serious thought is being given to breaking CL free from the pristine localhost prison that it is currently in. Hard for folks to adopt something that is only comfortable in one deployment scenario.

[...]

PR #1260, opened 2024-08-25
Serve multiple agents from a single chainlit server, access request path/query parameters

nethi:
This PR adds support to serve multiple agents  from a single chainlit server. A good example is a no-code agent server that takes agent_id from URL (request path or query parameters), fetches details of the agent from another server and setups chainlit agent accordingly.

In order to do this, I needed to do the following.
1) Ability to access agent_id from request path/query parameters
Several chainlit hooks donot have any way to get request path or query parameter details from several of the http hooks. I have added a ContextVar setup and a utility function to access fastapi request in chainlit http hooks

With this, we can now mount chainlit on path like /agents/{agent_id} and access path in hooks

For http based cl hooks
                                    
```python
        from chainlit.utils import get_current_http_request

        req = get_current_http_request()
        agent_id = extract_agent_id_from_path(req.url.path, prefix_to_match="/agents")
```

For ws based cl hooks

```python
@cl.on_chat_start
async def on_chat_start():
        http_path = cl.user_session.get("http_path")
        http_query = cl.user_session.get("http_query")
```

Besides being able to serve multiple agents from single server, accessing URL path and query parameters is also a frequent ask from several chainlit users in the forums.

2) Customize Socket IO path

while we can set http endpoint path as /agents/{agent_id},  we cannot use this approach for Socket.io as socket.io asgi server doesn't support path parameters or wild card prefixes.  In the above example, socket.io path will become /agents/{agent_id}/ws/socket.io,  which won't work. But socket.io server allows to add any extra path parameters such as /ws/socket.io/foobar  

With this ability to customize socket.io path, it can instead be configured as, for example: /agents/ws/socket.io.  With this, a custom frontend can connect to /agents/ws/socket.io/agent1.  websocket hooks then can use logic like this to extract "agent1" from path parameter.

```python
 cl.mount_chainlit(app=app, target=target, 
                                    path="/agents/{agent_id}",
                                    sio_path="/agents/ws")

@cl.on_chat_start
async def on_chat_start():
        http_path = cl.user_session.get("http_path")
        http_query = cl.user_session.get("http_query")

        agent_id = extract_agent_id_from_path(http_path, prefix_to_match="/agents/ws/socket.io")
        
        #fetch agent_id details from an external server
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,9 +4,9 @@ on:
   workflow_call:
   workflow_dispatch:
   pull_request:
-    branches: [main, dev]
+    branches: [main, dev, 'release/**']
   push:
-    branches: [main, dev]
+    branches: [main, dev, 'release/**']
 
 permissions: read-all
 
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -34,6 +34,9 @@ jobs:
         name: Install Python, poetry and Python dependencies
         with:
           poetry-working-directory: ${{ env.BACKEND_DIR }}
+      - name: Build Python distribution
+        run: poetry self add poetry-plugin-ignore-build-script && poetry build --ignore-build-script
+        working-directory: ${{ env.BACKEND_DIR }}
       - name: Publish package distributions to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
--- a/backend/README.md
+++ b/backend/README.md
@@ -18,7 +18,7 @@ Chainlit is an open-source async Python framework which allows developers to bui
 
 Full documentation is available [here](https://docs.chainlit.io). You can ask Chainlit related questions to [Chainlit Help](https://help.chainlit.io/), an app built using Chainlit!
 
-> [!NOTE]  
+> [!NOTE]
 > Check out [Literal AI](https://literalai.com), our product to monitor and evaluate LLM applications! It works with any Python or TypeScript applications and [seamlessly](https://docs.chainlit.io/data-persistence/overview) with Chainlit by adding a `LITERAL_API_KEY` in your project.
 >
 > Chainlit is developed and maintained by the Literal AI team, which is currently focused on expanding the capabilities of Literal AI. While we continue to support and maintain Chainlit, we are also committed to enabling the community to contribute, particularly in areas like integrations and data layers.
@@ -43,7 +43,7 @@ If this opens the `hello app` in your browser, you're all set!
 The latest in-development version can be installed straight from GitHub with:
 
 ```sh
-pip install git+https://github.com/Chainlit/chainlit.git@dokterbob/build_frontend_on_poetry_build#subdirectory=backend/
+pip install git+https://github.com/Chainlit/chainlit.git#subdirectory=backend/
 ```
 
 (Requires Node and pnpm installed on the system.)
--- a/package.json
+++ b/package.json
@@ -25,8 +25,7 @@
     "formatUi": "cd frontend && pnpm run format",
     "lintPython": "cd backend && poetry run mypy chainlit/ tests/",
     "formatPython": "black `git ls-files | grep '.py$'` && isort --profile=black .",
-    "buildUi": "cd libs/react-client && pnpm run build && cd ../copilot && pnpm run build && cd ../../frontend && pnpm run build",
-    "build": "pnpm run buildUi && (mkdir -p backend/chainlit/frontend && cp -R frontend/dist backend/chainlit/frontend) && (mkdir -p backend/chainlit/copilot && cp -R libs/copilot/dist backend/chainlit/copilot) && (cd backend && poetry build)"
+    "buildUi": "cd libs/react-client && pnpm run build && cd ../copilot && pnpm run build && cd ../../frontend && pnpm run build"
   },
   "pnpm": {
     "overrides": {


dokterbob:
Thanks for the contrib @nethi! 

As I see it, there's _two_ major _and_ desirable features in this one. But they're a big overhaul, so it's essential we think and align well on it.

In cases of larger overhauls, I usually prefer to hash out the features really well in issues, to make sure we can accept as much of your work as possible. In addition, to facilitate review, feedback and overview, I strongly prefer to have separate PR's for separate features. 

In this case, your work overlaps with the following two issues;
1. #1213 (as you rightly mention)
2. #1223 

After a quick read through your PR and code, in light of the more general suggestions above, I'm observing that:
1. You're choosing to privilege the `path` and `query` above other request parameters. 
2. You're _adding_ rather than removing complexity from the `mount_chainlit()` logic.

### 1. Copying `Request` properties
With regards to 1, the problem is that we are opening ourselves up to adding support (increasing our API and maintenance requirements) for a literal unlimited amount of manually configured parameters mapping to properties of `Request`. We already have quite a few and it seems a slippery slope to keep adding more of them rather than just providing access to the full `Request` object as it's received _either_ upon loading the frontend (index view) _or_ on starting the Websocket sessions (`on_chat_start()`).

Unless I'm missing something (still somewhat new to the Chainlit codebase), I would need some convincing why we shouldn't actually remove/deprecate mirrored properties of `Request` and allow access to the entire `Request` object.

### 2. Simplifying app/mount logic
I fully feel your pain on not being able to have multiple Chainlit's mounted and it's really a feature we _need_. In fact, it's something I personally need for my own use cases. 

But, again, there seems to be some fishy stuff going on that smells like workaround to me _in the original implementation_. In `mount_chainlit()`, we _currently_ seem to be using environment variables to pass information from one end of the application to another. While, if I understand this well, this information is only ever passed in this direction (though it's also referenced somewhere else). Hence, _we're implicitly using the environment to manage global state_.

My proposal in #1223 tries to make Chainlit a normal ASGI app, which can be mounted as any ASGI app in Starlette or FastAPI and in which SocketIO is mounted on a subpath. This would get rid of silly global state (in favour of normal access to request path) and not only allow multiple agents but allow developers to truly go bananas in terms of use cases (dynamically created share links, whatever).

Again, I might be wrong on this and the reason I have not yet started work on this is that I first want to explore all implications of this, so we can arrive at a more stable API, a smaller code base _and_ a better developer experience.

As such, seeing what you did to make this work in the current PR, I'd _love_ to hear your opinion on my proposals. In the process, I am still convincible that my approach is not the right way to go -- but I'll need some convincing before choosing to go that way. ;)

nethi:
> Thanks for the contrib @nethi!
> 
> As I see it, there's _two_ major _and_ desirable features in this one. But they're a big overhaul, so it's essential we think and align well on it.
> 
> In cases of larger overhauls, I usually prefer to hash out the features really well in issues, to make sure we can accept as much of your work as possible. In addition, to facilitate review, feedback and overview, I strongly prefer to have separate PR's for separate features.
> 
> In this case, your work overlaps with the following two issues;
> 
> 1. [Make Request object available in `on_chat_start()` #1213](https://github.com/Chainlit/chainlit/issues/1213) (as you rightly mention)
> 2. [Export ASGI `app` allowing for `app.mount()` instead of `mount_chainlit()` #1223](https://github.com/Chainlit/chainlit/issues/1223)
> 
> After a quick read through your PR and code, in light of the more general suggestions above, I'm observing that:
> 
> 1. You're choosing to privilege the `path` and `query` above other request parameters.
> 2. You're _adding_ rather than removing complexity from the `mount_chainlit()` logic.
> 
> ### 1. Copying `Request` properties
> With regards to 1, the problem is that we are opening ourselves up to adding support (increasing our API and maintenance requirements) for a literal unlimited amount of manually configured parameters mapping to properties of `Request`. We already have quite a few and it seems a slippery slope to keep adding more of them rather than just providing access to the full `Request` object as it's received _either_ upon loading the frontend (index view) _or_ on starting the Websocket sessions (`on_chat_start()`).
> 
> Unless I'm missing something (still somewhat new to the Chainlit codebase), I would need some convincing why we shouldn't actually remove/deprecate mirrored properties of `Request` and allow access to the entire `Request` object.
> 
> ### 2. Simplifying app/mount logic
> I fully feel your pain on not being able to have multiple Chainlit's mounted and it's really a feature we _need_. In fact, it's something I personally need for my own use cases.
> 
> But, again, there seems to be some fishy stuff going on that smells like workaround to me _in the original implementation_. In `mount_chainlit()`, we _currently_ seem to be using environment variables to pass information from one end of the application to another. While, if I understand this well, this information is only ever passed in this direction (though it's also referenced somewhere else). Hence, _we're implicitly using the environment to manage global state_.
> 
> My proposal in #1223 tries to make Chainlit a normal ASGI app, which can be mounted as any ASGI app in Starlette or FastAPI and in which SocketIO is mounted on a subpath. This would get rid of silly global state (in favour of normal access to request path) and not only allow multiple agents but allow developers to truly go bananas in terms of use cases (dynamically created share links, whatever).
> 
> Again, I might be wrong on this and the reason I have not yet started work on this is that I first want to explore all implications of this, so we can arrive at a more stable API, a smaller code base _and_ a better developer experience.
> 
> As such, seeing what you did to make this work in the current PR, I'd _love_ to hear your opinion on my proposals. In the process, I am still convincible that my approach is not the right way to go -- but I'll need some convincing before choosing to go that way. ;)

Thank you for your detailed comments/observations. I am inclined to pass Request object, but two concerns:
 a) Given websocket session is handled by socketio, starlette request mayn't be readily available. Is there another way to do this ?
 b) Question of API contract change - I diI dn't want to introduce a major contract change and followed along the approach used in the current code base.
 
Also, I have been using chainlit only for the last 8 weeks or so, so I mayn't fully understand some of the design choices and rationale. I can rework the PR based on consensus.

nethi:
@dokterbob  It looks like there is still no solution to the problem of retrieving websocket request, query (and cookie headers if possible) parameters.  I know there were several discussions on this topic but not sure if any other alternatives are in works ?

If there is no alternate solution in works, are you open to considering a PR update where it only contains adding PATH, Query and Cookie headers to the user session object ?

PR #1261, opened 2024-08-26
Allow the input chat string to be empty when have attachments.

EcoleKeine:
This enables a scenario where a user wishes to only submit attachments
for example, a document or text translator app


dokterbob (approved):
Code looks good to me, let's see whether tests are passing. Thanks for supplying the use case.

dokterbob:
Wow. Why did this merge without tests passing!???

Issue #1240, opened 2024-08-19
Replace custom oauth by well-supported library


dokterbob:
**Is your feature request related to a problem?**
The current auth implementation yields a [lot of issues](https://github.com/Chainlit/chainlit/labels/auth). But auth is not by far the USP for chainlit and arguably, it will never be.

In addition, rolling your own auth yields a plethora of security risks. Security is hard.
In short, unless you're a security expert, if you roll your own authn/authz, sooner later, you _will_ fail.

**Describe the solution you'd like**
But we don't have to fail! There's great and well supported libraries for handling (o)auth. 

For one, FastAPI has built-in oauth authorization (as a server): https://fastapi.tiangolo.com/tutorial/security/
In addition, there's fastapi-oauth2 which seems to provide world-class support of a plethora of oauth providers (as a client): https://github.com/pysnippet/fastapi-oauth2

The combination should allow us to ensure:
1. Better UX (e.g. no quirky redirects, working logout, 
2. Significantly better security (consider that we might be handling confidential conversations!).
3. Better provider support (have a look: https://github.com/python-social-auth/social-core/tree/master/social_core/backends).
4. Less code and so less maintenance.

**Plan of action**
We'd have to lay out a clear roadmap, this would deprecate a lot of code and break a lot of UX. Arguably, this is the biggest challenge about such a change.

At this point, the issue merely serves as a 'test balloon'. Do maintainers/devs and community members want/need this? Is it feasible? Something I glanced over?

stephenrs:
> But auth is not by far the USP for chainlit and arguably, it will never be.
> In addition, rolling your own auth yields a plethora of security risks. Security is hard.
> In short, unless you're a security expert, if you roll your own authn/authz, sooner later, you will fail.

@dokterbob Just found this after writing my dissertation on #1265. I couldn't agree more.

stephenrs:
I thought it might be helpful to give you a better sense of my particular use case…CL will be fulfilling 2 roles in the domain of customer support:

1. The full app version with chat history is already deployed internally as a trial/experiment to assist human customer support agents in crafting robust responses to customers - and they/we love it. 
2. A Copilot version will be first deployed as an in-app assistant for existing users, then if all goes well, it will be released to our public website for anyone to use.

Under the hood, 2 separate instances of CL will be running on separate ports because I’ve split the CL app target into separate modules that have slightly different behaviors but share a common core. So, for example, they both have blocks that look something like this:
```

@cl.on_message
[...]

PR #1280, opened 2024-08-30
Add option for custom auth

patrykkotlowski-dsstream:
# Custom authentication

### What's changed:
- New options were added to the configuration
- There is an option to create custom oauth provider
- There is an option for custom JWT token check.


dokterbob:
Thanks for the contrib @patrykkotlowski-dsstream! 

Do you see any way of using this work to arrive at a more generic pluggable auth? We're really looking forward to move support of auth frameworks more towards the community, so we'd rather remove than add new methods.

But if yours could be refactored towards having more pluggable auth methods (e.g. all auth methods are classes) that can be easily swapped, which can be returned from a _single_ call back hook, that might be really nice.

In addition, before we merge this in, we'd really like to see both E2E and unit tests. Otherwise, it's really hard for us to ensure that it won't break time and again while building on other stuff.

What do you think?

patrykkotlowski-dsstream:
Hi, thank you for the feedback!

I would be more than happy to add both E2E and unit tests to ensure the stability of the implementation. Also, I’m definitely interested in refactoring the auth methods in the future to make them more pluggable, as suggested. However, if it’s okay with you, I’d prefer to handle that refactor in a separate PR, since I’m currently focused on adding the CustomAuth provider as quickly as possible.

Let me know your thoughts!

danpe:
Would be great if the custom JWT token check will be able to use something like https://github.com/TCatshoek/fastapi-nextauth-jwt

dokterbob:
> https://github.com/TCatshoek/fastapi-nextauth-jwt

Where I want to be, honestly, is we don't implement auth ourselves but rather rely on a well-supported libraries (client and server-side) in order to reduce attack and maintenance surface.

@patrykkotlowski-dsstream Happy to merge once test coverage lands (at least E2E, unit test is nice to have at this point). Definitely, the refactor should not be part of this request!

lwieczorek-dss:
> > https://github.com/TCatshoek/fastapi-nextauth-jwt
> 
> Where I want to be, honestly, is we don't implement auth ourselves but rather rely on a well-supported libraries (client and server-side) in order to reduce attack and maintenance surface.
> 
> @patrykkotlowski-dsstream Happy to merge once test coverage lands (at least E2E, unit test is nice to have at this point). Definitely, the refactor should not be part of this request!

Hi folks, we've covered the code with unit tests but we are not sure about the E2E as OAuth for default supported providers is not covered with E2E tests.

backend/tests/test_callbacks.py
```diff
@@ -107,6 +107,65 @@ async def auth_func(
...
+    ):
+
+        @custom_authenticate_user
+        async def auth_func(
+            provider_id: str,
+            token: str,
+            raw_user_data: dict,
+            default_app_user: User,
+            id_token: str | None = None,
+        ) -> User | None:
```
dokterbob, line 129: What's the reason to add this callback over using `@oauth_callback`?

What does the value of `id_token` represent, in OAuth terms -- and why is it not tested?

backend/chainlit/callbacks.py
```diff
@@ -87,6 +91,38 @@ async def oauth_callback(provider_id: str, token: str, raw_user_data: Dict[str,
...
+    if len(get_configured_oauth_providers()) == 0:
+        raise ValueError(
+            "You must set the environment variable for at least one oauth provider to use oauth authentication."
+        )
+
+    config.code.custom_authenticate_user = wrap_user_function(func)
+    return func
+
+
+def custom_oauth_provider(func: Callable[[], OAuthProvider]) -> None:
```
dokterbob, line 115: I am not sure whether `custom_` is the right name here. As we'll be moving to having less and less oauth providers built in, I feel it's more likely that _most_ if not _all_ oauth providers will be pluggable. Additionally, I want to move towards using an externally maintained OAuth client framework instead (see #1240). 

So perhaps, this could become the _default_ way of setting up oauth providers -- in that case something like `@register_oauth_provider` or merely `@oauth_provider` might be a more appropriate name.

backend/chainlit/data/acl.py
```diff
@@ -8,7 +8,7 @@ async def is_thread_author(username: str, thread_id: str):
         raise HTTPException(status_code=400, detail="Data layer not initialized")
 
     thread_author = await data_layer.get_thread_author(thread_id)
-    
+
```
dokterbob, line 11: Please make sure this is removed from the PR, it is conducive to merge conflicts.

dokterbob:
> > > https://github.com/TCatshoek/fastapi-nextauth-jwt
> > 
> > 
> > Where I want to be, honestly, is we don't implement auth ourselves but rather rely on a well-supported libraries (client and server-side) in order to reduce attack and maintenance surface.
> > @patrykkotlowski-dsstream Happy to merge once test coverage lands (at least E2E, unit test is nice to have at this point). Definitely, the refactor should not be part of this request!
> 
> Hi folks, we've covered the code with unit tests but we are not sure about the E2E as OAuth for default supported providers is not covered with E2E tests.

That's exactly one of the problems. ;)
Auth is definitely something that requires testing!

dokterbob (commented):
Thanks so much for your endeavours! This is actually a really great feature!
However...

Given that I aim to move towards using an external OAuth library, I wonder a bit what to do with this PR -- as it stands it would get more users to rely on `OAuthProvider` which is exactly what I aim to remove from the codebase!

Then again, perhaps we can evolve this PR into using a 3rd party library already? Alternatively, we can leave this one open and I'll attempt to adopt it to the 'cleaner' OAuth approach -- although it could take ~6 weeks for me to do that.

Happy to hear what y'all think. Whether perhaps you have the bandwidth to:
1. Help evaluate/choose a _well supported_, _test-covered_ and _actively maintained_ OAuth client lib (e.g. implementing `OAuthProvider` type of functionality).
2. Integrate it with this work, possibly refactoring the way OAuth is configured (perhaps explicit instead of implicit?).

dokterbob:
> @dokterbob @patrykkotlowski-dsstream This is a good solution for one of the Auth problem I have as well. Is there a chance of this getting merged soon ?
> 
> Here is the problem I am looking at solving:
> 
> 1. There is a non-chainlit Auth token in my system where the user is already authenticated
> 2. I would pass this as Auth token to Chainlit
> 3. Implement custom_authenticate_user callback to validate the token and create chainlit User object.
> 
> I suppose this would work for both chainlit REST APis and websocket connections.

It's currently blocked by requested changes in the review here. (Only) if @patrykkotlowski-dsstream is not available to address them, you're welcome to fork their branch to address the issues, resolve merge conflicts and do a 2nd PR.

nethi:
> > @dokterbob @patrykkotlowski-dsstream This is a good solution for one of the Auth problem I have as well. Is there a chance of this getting merged soon ?
> > Here is the problem I am looking at solving:
> > 
> > 1. There is a non-chainlit Auth token in my system where the user is already authenticated
> > 2. I would pass this as Auth token to Chainlit
> > 3. Implement custom_authenticate_user callback to validate the token and create chainlit User object.
> > 
> > I suppose this would work for both chainlit REST APis and websocket connections.
> 
> It's currently blocked by requested changes in the review here. (Only) if @patrykkotlowski-dsstream is not available to address them, you're welcome to fork their branch to address the issues, resolve merge conflicts and do a 2nd PR.

@patrykkotlowski-dsstream would you be looking at making requested changes ?

dokterbob:
@nethi Just a little head's up that I'm in the last passes of making cookie-based auth the default #1521.

From now on, the user will authenticate with the backend, the backend sets a HTTP-only cookie and from thereon all requests are authenticated with cookies. This is required as we're serving static files, potentially scripts, potentially user or LLM-generated, and we want to make sure those won't have access to our auth token.

The actual login authentication mechanism does not change as this point (but that's on the roadmap! :p).

To have a 3rd party site authenticate against chainlit, the best approach is to:
1. Have the 3rd party side generate a short-lived signed auth token.
2. Have the client do a POST call to some chainlit auth endpoint, validating the token and setting the auth cookie.

(Ideally, used and unexpired short-lived tokens should be kept track of to prevent replay attacks, but that requires shared state/persistence which we currently do not have. Other than that, this is best practise.)

I believe a similar mechanism is already employed for the [copilot](https://docs.chainlit.io/deploy/copilot). In the coming few days, I will have to get down into the nitty gritty of things in order to finish cookie based auth.

lwieczorek-dss:
> Chainlit already supports custom auth, maybe not documented well enough. The custom frontend cookbook example uses it https://github.com/Chainlit/cookbook/blob/main/custom-frontend/backend/app.py

It's not what we are looking for tbh. I like the cookie-based approach but we are still missing of a way of additional token validation on our side. All the chainlit endpoints depends on `get_current_user` method where user is authenticated but we cannot secure those endpoints with our additional token validation due to lack of customization of this method.

PR #1285, opened 2024-09-01
[fix] `LlamaIndexCallbackHandler` to support displaying function calls as tools in CoTs

tslmy:
In Llama Index, tool-calling events are emitted with the event type of `FUNCTION_CALL`. All tools in Llama Index, when used, emits these events, so they should definitely be supported in Chainlit.

**Example usage** is in [this repo](https://github.com/StarsRail/Cocai). This is a simple LLM agent that can roll a dice for you. The dice-rolling capability is implemented as simple `random.randint` call ([code](https://github.com/StarsRail/Cocai/blob/b94d26c38aae1542c8ceea316af3822aed2c3a0e/tools.py#L26)) wrapped with `FunctionTool` ([code](https://github.com/StarsRail/Cocai/blob/b94d26c38aae1542c8ceea316af3822aed2c3a0e/main.py#L125-L127)).

| Before | After | 
| --- | --- |
|  image (notice that the the function-calling step is missing) | image |

### DIFF
--- a/backend/chainlit/llama_index/callbacks.py
+++ b/backend/chainlit/llama_index/callbacks.py
@@ -8,6 +8,7 @@ from literalai.helper import utc_now
 from llama_index.core.callbacks import TokenCountingHandler
 from llama_index.core.callbacks.schema import CBEventType, EventPayload
 from llama_index.core.llms import ChatMessage, ChatResponse, CompletionResponse
+from llama_index.core.tools.types import ToolMetadata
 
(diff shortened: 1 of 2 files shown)


dokterbob (approved):
LGTM.

dokterbob:
@tslmy Thanks again for the contribt!

I agree that fixing this issue is out of scope but it would be great if you could use your current awareness of it to amend the issue I created: https://github.com/Chainlit/chainlit/issues/1310

PR #1267, opened 2024-08-27
Relax dependencies

EWouters:
This includes many updates for Python dependencies. Many packages were restrained to out of date versions. This is the most up to date set of dependencies I've been able to come up with. Once no packages depend on `pydantic=<2` anymore (`farm-haystack`) a lot of packages can be relaxed to current versions.

My notes to reproduce:
```bash
# Restraints
poetry add httpx literalai dataclasses-json fastapi starlette uvicorn python-socketio aiofiles syncer asyncer nest-asyncio click tomli pydantic="^1.10" python-dotenv uptrace watchfiles filetype lazify packaging="^23.2" python-multipart pyjwt numpy="^1.26"
poetry add --group=tests pytest pytest-asyncio pytest-cov openai langchain="^0.2.5" llama-index="^0.10.68" tenacity="^8.4" transformers="^4.39" matplotlib farm-haystack plotly slack_bolt discord botbuilder-core
poetry add --group=dev black isort
poetry add --group=mypy mypy types-requests="^2.31" types-aiofiles mypy-boto3-dynamodb
poetry add --group=custom-data asyncpg SQLAlchemy boto3 azure-identity azure-storage-file-datalake

# Restraints + Relaxations
poetry add httpx="^0.27" literalai dataclasses-json fastapi="^0.112" starlette="^0.38" uvicorn="^0.30" python-socketio="^5.11" aiofiles="^24.1" syncer asyncer nest-asyncio click tomli pydantic="^1.10" python-dotenv uptrace="^1.26" watchfiles="^0.23" filetype lazify packaging="^23.2" python-multipart pyjwt="^2.9" numpy="^1.26"
poetry add --group=tests pytest pytest-asyncio="^0.24" pytest-cov openai="^1.42" langchain="^0.2.5" llama-index="^0.10.68" tenacity="^8.4" transformers="^4.39" matplotlib="^3.9" farm-haystack plotly="^5.23" slack_bolt="^1.20" discord botbuilder-core="^4.16"
poetry add --group=dev black="^24.8" isort="^5.13"
poetry add --group=mypy mypy types-requests="^2.31" types-aiofiles="^24.1" mypy-boto3-dynamodb="^1.35"
poetry add --group=custom-data asyncpg SQLAlchemy boto3="^1.35" azure-identity="^1.17" azure-storage-file-datalake="^12.16"

```

Issue #1266, opened 2024-08-27
Prepare for release 1.1.403


dokterbob:
ETA: ~4/9

* [ ] Update changelog
* [ ] Update docs based on changelog
* [ ] Manual testing by @willydouhard

EWouters:
> Hey! Do you think you guys will have time to add relaxed requirements for dataclasses_json in this release?
> 
> Using LangChain with unstructured currently gives the following error:
> 
> The conflict is caused by: chainlit 1.1.304 depends on dataclasses_json<0.6.0 and >=0.5.7 unstructured-client 0.24.1 depends on dataclasses-json>=0.6.4

Addressed in #1267

dokterbob:
> > Hey! Do you think you guys will have time to add relaxed requirements for dataclasses_json in this release?
> > Using LangChain with unstructured currently gives the following error:
> > The conflict is caused by: chainlit 1.1.304 depends on dataclasses_json<0.6.0 and >=0.5.7 unstructured-client 0.24.1 depends on dataclasses-json>=0.6.4
> 
> Addressed in #1267

We'd be open to have some _explicit_ removal of dep versions, particularly if this comes with unit tests for the interaction surface with those specific deps (so we can be sure that chainlit doesn't break if we remove `<` versions, now or in the future).

dokterbob:
We're probably gonna bump the minor version due to added features.

Release PR/changelog is here: https://github.com/Chainlit/chainlit/pull/1289

Release ETA: later today.

Issue #1224, opened 2024-08-16
Build frontend as part of backend build


dokterbob:
**Is your feature request related to a problem?**
An increasing amount of users are using `mount_chainlit()` to deploy from within another FastAPI app.

Some of them are running into issues with dependencies (#731 #950) and many of them (including the undersigned) want to make improvements or changes and test them locally.

As it stands, the Python package can be installed from a subdirectory (e.g. `poetry add git+https://github.com/myorg/mypackage_with_subdirs.git@main#subdirectory=subdir`) but in this case it seems that the frontend is not distributed alongside the package as it would be for PyPI.

**Describe the solution you'd like**
Ideally, I would suggest that the Python install automatically builds the frontend on install. Alternatively, we should at least have instructions on how to proceed in such a case.

#### Build on install
It's not evidently clear how to call arbitrary code (like `pnpm run buildUi` to build the frontend) during the Python package installation process, especially not when installing as editable/from source.

Example: https://sam.hooke.me/note/2023/08/poetry-build-py-example/

#### Build on first launch
One approach might be to build the frontend on first application launch.

We'd check for existence of the frontend dir and if not, we build.

#### Package frontend separately
Another could be to package the frontend as a separate (binary/blob) type of dependency, so that it can be separately installed.

An example of this: https://pypi.org/project/vite-project/ (source: https://github.com/itayB/vite-project)

**Resources/references**
https://discuss.python.org/t/running-custom-code-on-package-install/16160/2

dokterbob:
Based on my research, we should be able to do this with a builds script, as follows (untested):

`pyproject.toml`:
```toml
[build-system]
requires = ["poetry-core", "setuptools"]

[tool.poetry.build]
script = "scripts/build_frontend.py"
generate-setup-file = true
```
`build_frontend.py`:
```
import subprocess
import sys

def build_frontend():
    try:
        subprocess.run(["npm", "install"], cwd="src/your_package/frontend", check=True)
        subprocess.run(["npm", "run", "build"], cwd="src/your_package/frontend", check=True)
    except subprocess.CalledProcessError:
        print("Error: Failed to build frontend. Make sure Node.js and npm are installed.")
        sys.exit(1)
    except FileNotFoundError:
        print("Error: Node.js or npm not found. Please install Node.js to build the frontend.")
        sys.exit(1)

if __name__ == "__main__":
    build_frontend()
```

Reference: https://github.com/python-poetry/poetry/issues/2740#issuecomment-1668981210

PR #1291, opened 2024-09-03
Bump dataclasses to latest version.

dokterbob:
As we're relying on dataclasses pretty much everywhere, and we don't nearly have enough unit test coverage, I would like to ask the community to assist in manual testing before we ship this in the latest release (in addition to contributions to increase unittest coverage).

As discussed [here](https://github.com/Chainlit/chainlit/pull/1267#issuecomment-2315006242), we prefer to be careful when upgrading dependencies and would prefer to do it one by one. Otherwise, if something goes wrong, it will be extremely challenging to figure out _where_ it went wrong. (#1267)

Depending on the amount of testing/feedback we can get in, we hope to ship this with the coming release. We're also actively working towards allowing (editable) installs straight from GIT so users won't be as dependent on our release cycle (and we can get more feedback on features/PR's/fixes). (#1224)

Specifically looking in the direction of @jpolvto, @aidanshipperley and @EWouters.

Ref:
* https://github.com/Chainlit/chainlit/issues/1266

### DIFF
--- a/backend/poetry.lock
+++ b/backend/poetry.lock
@@ -1027,22 +1027,18 @@ tests = ["pytest", "pytest-cov", "pytest-xdist"]
 
 [[package]]
 name = "dataclasses-json"
-version = "0.5.9"
-description = "Easily serialize dataclasses to and from JSON"
+version = "0.6.7"
+description = "Easily serialize dataclasses to and from JSON."
 optional = false
-python-versions = ">=3.6"
+python-versions = "<4.0,>=3.7"
 files = [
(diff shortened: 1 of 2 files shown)


dokterbob:
I'm opening this up for review but I might add some test coverage later.

willydouhard (approved):

dokterbob:
To be merged after:
* https://github.com/Chainlit/chainlit/pull/1288
* https://github.com/Chainlit/chainlit/pull/1292

PR #1311, opened 2024-09-04
Removed redundant key/value pair from dictionary creator

duckboy81:
Just removing an extraneous key/value declaration set during the creation of a dictionary.

`createdAt` is set on line 79 and line 85

https://github.com/Chainlit/chainlit/blob/122369a64ebdc3d34f6f88cbe24ec612d209c20b/backend/chainlit/message.py#L74-L93


dokterbob (approved):
Looks good to me! Thanks!

PR #1316, opened 2024-09-05
Fixed using multiple auth methods clashing and automatically kicking out users.

Jaluus:
# Fixed Password Auth

## What was the problem?

When using password auth and header based auth at the same time the frontend called the header auth regardless, even if a user was already authenticated kicking the user out.

## What changed

Moved the check if a user is authenticated in front of the header auth call.
If a user is authenticated by the password auth, the header based auth or an existing JWT the app will now not purge the JWT and thus kick out any user.


willydouhard:
I think I agree with this

willydouhard:

PR #1318, opened 2024-09-06
move hotkey after return statement

jdb78:
Hotkey for microphone recording even gets triggered if audio input is disabled. this fixes the issue


dokterbob:
Thanks for the contrib @jdb78! Code looks sensible and good to me.

Would be great to have a test for it but I imagine audio in a headless browser might be challenging. Nonetheless, if you're up for a challenge, I could leave the PR open or, to move things faster, be open to a second PR with tests for the audio feature.

In the mean time, I would appreciate feedback from other reporters to move this merge forward -  our bandwidth for manual testing is limited. If I get one OK, I'm confident to approve and merge. @stephenrs @puppetm4st3r

xandao-dev:
As a workaround I have completely disabled the 'p' hotkey, because whenever I press 'p' in the copilot it triggers the mic.


window.cl_shadowRootElement.removeEventListener('keydown', blockPKeyEvent);
    function blockPKeyEvent(event) {
      if (event.key === 'P' || event.key === 'p') {
        event.stopPropagation();
      }
    }

willydouhard:
I don't think you can return before a hook

willydouhard:

PR #1325, opened 2024-09-09
Refine feedback UI and improve type handling

willydouhard:


backend/chainlit/langchain/callbacks.py
```diff
@@ -587,12 +587,17 @@ def _on_run_update(self, run: Run) -> None:
         outputs = run.outputs or {}
         output_keys = list(outputs.keys())
         output = outputs
+        
+        print("output_keys", output_keys)
+        print("outputs", outputs)
```
dokterbob, line 592: (!)

backend/chainlit/langchain/callbacks.py
```diff
@@ -587,12 +587,17 @@ def _on_run_update(self, run: Run) -> None:
...
         if output_keys:
             output = outputs.get(output_keys[0], outputs)
+            
+        print("output", output)
```
dokterbob, line 596: (!)

dokterbob (changes requested):
Please:
* Remove debug `print()` statements.
* Irrelevant code in FeedbackButtons and (possibly) on line 600 (or add a comment how it's relevant).

And, most importantly:
* Add a comment _why_ env should be loaded at the top, so we won't get this problem again!

In time, I do hope we can refactor the code to prevent this issue from occurring in the first place. :)

willydouhard:
Nice catch, I removed the print statements. The reason the .env should be loaded at the top is that the data layer is checking for the `LITERAL_API_KEY` to decide if data persistance should be enabled or not.

backend/chainlit/langchain/callbacks.py
```diff
@@ -587,12 +587,13 @@ def _on_run_update(self, run: Run) -> None:
...
         output = outputs
+
         if output_keys:
             output = outputs.get(output_keys[0], outputs)
-
+            
         if current_step:
             current_step.output = (
-                output[0] if isinstance(output, Sequence) and len(output) else output
+                output[0] if isinstance(output, Sequence) and not isinstance(output, str) and len(output) else output
```
willydouhard, line 596: Since a string is a `Sequence`, it would only take the first character of the string. This is improving the type check to make sure the output is correctly logged.

frontend/src/components/molecules/messages/Message.tsx
```diff
@@ -186,9 +186,12 @@ const Message = memo(
                       {actions?.length ? (
                         <MessageActions message={message} actions={actions} />
                       ) : null}
-                      {scorableRun && isScorable ? (
-                        <MessageButtons message={message} run={scorableRun} />
-                      ) : null}
+                      <MessageButtons
```
willydouhard, line 189: Previously, feedback buttons were displayed if the run ended with an assistant message. Feedback buttons should be displayed at the end of the run regardless of the type of the last step

frontend/src/components/molecules/messages/Messages.tsx
```diff
@@ -79,11 +79,19 @@ const Messages = memo(
             // Score the current run
             const _scorableRun = m.type === 'run' ? m : scorableRun;
             // The message is scorable if it is the last assistant message of the run
-            const isScorable =
+
+            const isRunLastAssistantMessage =
```
willydouhard, line 83: see comment above

frontend/src/components/molecules/messages/components/FeedbackButtons.tsx
```diff
@@ -39,6 +44,10 @@ const FeedbackButtons = ({ message }: Props) => {
   const [feedback, setFeedback] = useState(message.feedback?.value);
   const [comment, setComment] = useState(message.feedback?.comment);
 
+  if (!config.config?.dataPersistence) {
```
willydouhard, line 47: feedback buttons should not be displayed if no data layer is configured.

dokterbob (approved):

PR #1326, opened 2024-09-09
Fastapi parameter validation

dokterbob:
Validate path hierarchy and path parameters for:
* `get_avatar`
* `project_settings
* `project_translations`

Includes test coverage for all relevant methods.

### DIFF
--- a/backend/tests/test_server.py
+++ b/backend/tests/test_server.py
@@ -1,9 +1,10 @@
 import os
-from unittest.mock import Mock, create_autospec
+from pathlib import Path
+from unittest.mock import Mock, create_autospec, mock_open
 
 import pytest
 from chainlit.auth import get_current_user
-from chainlit.config import APP_ROOT, ChainlitConfig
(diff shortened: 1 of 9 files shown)


willydouhard (approved):
lgtm

backend/tests/test_server.py
```diff
@@ -0,0 +1,230 @@
...
+def test_get_avatar_default(test_client: TestClient, monkeypatch: pytest.MonkeyPatch):
+    """Test with default avatar."""
+    response = test_client.get("/avatars/default")
+    assert response.status_code == 200
+    assert response.headers["content-type"].startswith("image/")
+
+
+def test_get_avatar_custom(test_client: TestClient, monkeypatch: pytest.MonkeyPatch):
+    """Test with custom avatar."""
+    custom_avatar_path = os.path.join(
```
LucasMagnum, line 161: @dokterbob this is assuming that the avatar will be called with correctly which is not the case for the chat. If the assistant name contains space, it will call the API passing that name from the UI and it will return a 400.

PR #1332, opened 2024-09-11
fix: negative feedback class incorrect

sandangel:


sandangel:
cc @willydouhard

dokterbob:
@sandangel Thanks for the PR! 

Could you perhaps explain a bit about the issue (e.g. with a screenshot) and give steps how to replicate and possibly fix it?

sandangel:
The feedback value is either 0 or 1.  It used to be -1, 0 or 1. We forgot to change the value here so the negative-feedback-on class was not applied correctly so we could not use it to style the button the way we want.

willydouhard (approved):

willydouhard:
Nice catch!

Issue #1330, opened 2024-09-10
Error Message "Could not reach the server" on mounting to FastAPI


sajsax10:
**Describe the bug**
When trying to run Chainlit mounted on FastAPI , the Chainlit app seems to be failing on the path "/chainlit" with the error message:

> Could Not Reach the Server

Accompanied by this in the terminal: 

> Translated markdown file for en-US not found. Defaulting to chainlit.md.

The Chainlit app runs perfectly fine when run separately.

**To Reproduce**
Error observed for Chainlit v1.1.404, normal behavior for v1.1.400

Mounting chainlit on FastAPI app
`mount_chainlit(app=app, target="chatbot/chatbot.py", path="/chainlit")`

Run FastAPI app:

` uvicorn main:app --host 0.0.0.0 --reload`

**Expected behavior**
App should be served as expected unless I'm doing something wrong

**Desktop (please complete the following information):**

- OS: Windows10
- Browser: Chrome
- Version 128.0.6613.120

stephenrs:
@sajsax10 Since you are specifying a path, this is likely related to #1313, #1262, and #1317

As far as I'm aware the last release that didn't have this critical bug was 1.1.402, but unfortunately the Chainlit company doesn't seem to be taking it very seriously :(

dokterbob:
@stephenrs We are. Please refrain from attacking us, it's unpleasant and unproductive. We appreciate your support but please be aware that we are doing are very best with the bandwidth we have.

@sajsax10 @stephenrs Could you please try and validate whether #1337 resolves your issue?

stephenrs:
@dokterbob I'm not attacking you, I'm pointing out my observations, and I'm glad to see you're finally showing some indications that you're paying attention to critical issues like this one.

I would recommend that you get your emotions under control, and stop whining and being so defensive. None of this is personal. Focus on the work and smile about it :)

stephenrs:
@dokterbob Just a quick follow-up. You seem to spend a considerable amount of time thinking about and talking about how you feel. So, I think it would be helpful for you to devote some time to thinking about how people in this community feel as a result of your actions, inactions, or lack of communication.

People who encountered this particular problem were completely blocked on their entire project if they didn't know they could downgrade to 1.1.402. This is likely an actual crisis for some people. So, I'm thinking about them and how they feel and doing my best to make sure they are informed about where your priorities appear to be so they don't get stuck in a crisis situation - because that feels terrible.

[...]

Issue #1317, opened 2024-09-06
Could not reach the server in 1.1.404


Tug:
**Describe the bug**
The client is displaying the following error when upgrading to 1.1.404 from 1.1.402:

> Could not reach the server

Inspecting the network we see that socket.io connections returns plain HTML:

<img width="1143" alt="Screenshot 2024-09-06 at 12 52 23" src="https://github.com/user-attachments/assets/de031ccd-daea-429b-9ca8-4442598e9651">

The url requested seems correct:

`http://localhost:8000/chat/ws/socket.io/?EIO=4&transport=polling&t=P76zd9L`

**To Reproduce**

- Clone this repo: https://github.com/Tug/chainlist-minimal-test-case
- Make sure you have poetry installed
- `make`
- Load http://localhost:8000/chat/

**Expected behavior**
The socket.io client should be able to establish a connection with the server

**Screenshots**

**Desktop (please complete the following information):**

- OS: macOS
- Browser: chrome
- Version 126

dokterbob:
@Tug Thanks for your report.

Could you please try and validate whether #1337 resolves your issue?

Tug:
@dokterbob Just tested, it does fix the issue 👍  thx

dokterbob:
@Tug We're shipping a fixup release today.

dokterbob:
Appreciate you provide such a clear test case. Thanks!

Issue #1313, opened 2024-09-05
Chainlit socket io path is set wrongly when ROOT_PATH is used


qtangs:
**Describe the bug**
Chainlit socket io path seems wrong.

**To Reproduce**

1. Run chainlit with environment variable: CHAINLIT_ROOT_PATH=chainlit
2. Open app using the custom path: http://localhost:8000/chainlit
3. App is not able to establish websocket connection

**Expected behavior**
App should work normally.

**Suggested solution**
I have tested changing these lines
- https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/server.py#L178
  fr: `sio_mount_location = f"{ROOT_PATH}/ws" if ROOT_PATH else "ws"`
  to: `sio_mount_location = f"{ROOT_PATH}/ws" if ROOT_PATH else "/ws"`

- https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/server.py#L185
  fr: `app.mount(f"/{sio_mount_location}", asgi_app)`
  to: `app.mount(sio_mount_location, asgi_app)`

**Screenshots**
Nil

**Desktop (please complete the following information):**

- OS: MacOS
- Browser: Chrome
- Version: 128.0.6613.120

**Smartphone (please complete the following information):**
N/A

**Additional context**
Nil

nethi:
I had to do similar workaround. Also I was surprised to see this breaking change in a previous release but no mention of it.

Tug:
> I have tested changing these lines
> 
> https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/server.py#L178
> fr: sio_mount_location = f"{ROOT_PATH}/ws" if ROOT_PATH else "ws"
> to: sio_mount_location = f"{ROOT_PATH}/ws" if ROOT_PATH else "/ws"
> 
> https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/server.py#L185
> fr: app.mount(f"/{sio_mount_location}", asgi_app)
> to: app.mount(sio_mount_location, asgi_app)

Indeed that fixes the problem. https://github.com/Chainlit/chainlit/pull/1203 seems to be the culprit. @qtangs will you open a PR ? I can do it also.

dokterbob:
Thanks for the report @qtangs!

Are you perhaps able to help to verify whether #1337 resolves your issue?

dokterbob:
Tested to work with `CHAINLIT_ROOT_PATH=chainlit` for #1337. 

[...]

PR #1337, opened 2024-09-12
fix: websocket connection when submounting chainlit

willydouhard:
Running this [example](https://docs.chainlit.io/integrations/fastapi) currently fails (websocket connection fails). This is an attempt to better submount chainlit and fix this issue.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -23,6 +23,14 @@ jobs:
       - name: Install fastapi ${{ matrix.fastapi-version }}
         run: poetry add fastapi@^${{ matrix.fastapi-version}}
         working-directory: ${{ env.BACKEND_DIR }}
+      - uses: ./.github/actions/pnpm-node-install
+        name: Install Node, pnpm and dependencies.
+        with:
+          node-version: 22.7.0
+          pnpm-version: 9.7.0
+          pnpm-install-args: --no-frozen-lockfile
+      - name: Build UI
+        run: pnpm run buildUi
       - name: Run Pytest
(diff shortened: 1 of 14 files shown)


sandangel:
Will this approach have issue with `router` prefix? So sending request to `/app-root` will now become `/app-root/app-root`?

dokterbob:
> Will this approach have issue with `router` prefix? So sending request to `/app-root` would become `/app-root/app-root`?

Unlikely but we'd love some feedback from users on this, as there's such a large variety of use cases.

willydouhard:
The router is [no longer prefixed](https://github.com/Chainlit/chainlit/pull/1337/files#diff-9a73e67e163edc3fae6839bf2bcc8c6ab650624eff6668502388a7a86b7ece78L195) so it should be fine imo

dokterbob:
Good step towards #1223!

willydouhard:
I tested this PR with different scenarios:
1. Submounting the application to a parent Fast API app (including with oauth redirections)
2. Starting Chainlit with `--root-path`
3. Starting Chainlit normally

dokterbob:
### Testing in progresss

- [ ] Mounted in FastAPI app
- [ ] Example from docs (https://github.com/Chainlit/docs/issues/172)
- [ ] Cookbook example (https://github.com/Chainlit/cookbook/issues/137)
- [ ] https://github.com/Chainlit/chainlit/issues/1313
- [ ] https://github.com/Chainlit/chainlit/issues/1317
- [ ] https://github.com/Chainlit/chainlit/issues/1330

dokterbob (approved):
Let's gooo!

PR #1342, opened 2024-09-16
Paul assistant

willydouhard:


backend/chainlit/__init__.py
```diff
@@ -402,6 +427,9 @@ def acall(self):
     "TaskStatus",
     "Video",
     "ChatSettings",
+    "AssistantSettings",
+    # assistant
```
willydouhard, line 431: remove comment

backend/chainlit/assistant.py
```diff
@@ -0,0 +1,16 @@
+from typing import List, Dict
```
willydouhard, line 1: if we have multiple assistant file it can be in a dedicated folder/module

backend/chainlit/server.py
```diff
@@ -194,11 +196,11 @@ def get_build_dir(local_target: str, packaged_target: str):
 
 router = APIRouter(prefix=ROOT_PATH)
 
-app.mount(
-    f"{ROOT_PATH}/public",
-    StaticFiles(directory="public", check_dir=False),
-    name="public",
-)
+# app.mount(
```
willydouhard, line 199: Why is this commented?

backend/chainlit/server.py
```diff
@@ -931,6 +931,34 @@ async def get_avatar(avatar_id: str):
...
+# post avatar/{avatar_id} (only for authenticated users)
+@router.post("/avatars/{avatar_id}")
+async def upload_avatar(
+    avatar_id: str,
+    file: UploadFile,
+    current_user: Annotated[
+        Union[None, User, PersistedUser], Depends(get_current_user)
+    ],
+):
+    # if not current_user:
```
willydouhard, line 943: why is this commented?

frontend/src/components/organisms/header.tsx
```diff
@@ -40,6 +41,7 @@ const Header = memo(() => {
         }}
       >
         <ChatProfiles />
+        <SelectedAssistantDisplay />
```
willydouhard, line 44: can chat profiles co exist with assistants?

frontend/src/components/organisms/sidebar/index.tsx
```diff
@@ -23,18 +25,18 @@ const DRAWER_WIDTH = 260;
...
   const { config } = useConfig();
+  const { assistantSettingsInputs } = useChatData();
   const enableHistory = !!user.accessToken && !!config?.dataPersistence;
 
   useEffect(() => {
     if (isMobile) {
       setChatHistoryOpen(false);
     } else {
-      setChatHistoryOpen(enableHistory);
+      setChatHistoryOpen(true);
```
willydouhard, line 37: we should use enableHistory and include assistants in it

frontend/src/components/organisms/header.tsx
ModEnter: they should not

PR #1362, opened 2024-09-21
Fix auto login

ModEnter:
Fix the automatic enter of the only oauth providers, which makes it unable to logout

### DIFF
--- a/backend/chainlit/oauth_providers.py
+++ b/backend/chainlit/oauth_providers.py
@@ -96,7 +96,7 @@ class GoogleOAuthProvider(OAuthProvider):
             "scope": "https://www.googleapis.com/auth/userinfo.profile https://www.googleapis.com/auth/userinfo.email",
             "response_type": "code",
             "access_type": "offline",
-            "prompt": "consent",
+            "prompt": "login",
         }
 
     async def get_token(self, code: str, url: str):
@@ -164,7 +164,7 @@ class AzureADOAuthProvider(OAuthProvider):
             "response_type": "code",
             "scope": "https://graph.microsoft.com/User.Read",
             "response_mode": "query",
-            "prompt": "consent",
+            "prompt": "login",
         }
 
     async def get_token(self, code: str, url: str):
@@ -249,7 +249,7 @@ class AzureADHybridOAuthProvider(OAuthProvider):
             "scope": "https://graph.microsoft.com/User.Read https://graph.microsoft.com/openid",
             "response_mode": "form_post",
             "nonce": nonce,
-            "prompt": "consent",
+            "prompt": "login",
         }
 
     async def get_token(self, code: str, url: str):
@@ -329,7 +329,7 @@ class OktaOAuthProvider(OAuthProvider):
             "response_type": "code",
             "scope": "openid profile email",
             "response_mode": "query",
-            "prompt": "consent",
+            "prompt": "login",
         }
 
     def get_authorization_server_path(self):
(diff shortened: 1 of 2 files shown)


ModEnter:
@dokterbob I did a new PR from the good branch this time 😉

dokterbob (commented):
I just reviewed the functionality (in our own app, with a single oauth provider). 

What I hoped this would do, was to remove the intermittent login step before the redirect, which our users have found confusing (as there's no user interaction required).

What instead happened, is that now there is a single button for the user to click; there is a choice to be made but nothing to choose. Remember those websites back in the 90's  with the 'Click here to ender the website'-button?

Of course, users need to be able to log out and I see the pain there. Our own users also have this issue.

What's happening AFAIK, is that when a user logs out from chainlit, the oauth authentication is immediately initiated again. _If a user was not logged out from the oauth provider_, this means that they are automatically logged in again. What the user's expectation is instead, is that when they're logging out, they are presented a prompt to log in again. If there's a single OAuth provider, they should end up seeing that provider's oauth login screen again (not a page with a single button).

So IMHO, what's necessary is that logging a user out in the frontend also logs them out on the oauth backend. The proper way might be to add an optional `async def logout()` method to `OAuthProvider` which for each OAuth provider uses the available credentials to have the provider revoke their credentials.

Alternatively, if 'properly' implementing OAuth (logout) (not a trivial matter!), perhaps we could make this behaviour configurable, setting prior behaviour as the default (e.g. `single_oauth_auto_redirect = True`, allowing users to specifically disable it?

dokterbob:
I could not help myself from digging a little deeper and it seems that OAuth2 has a `prompt=consent` option to the auth request. This way, a user is not logged out from their OAuth provider (we don't want users to log out from GH if they use GH to login to Chainlit!), but it will require users to confirm their login (and allow them to change to a different user). 

To implement this, we could add `prompt=consent` on all outgoing auth requests -- which AFAIK (we need to test this) only happens when a user as logged out, requiring chainlit to send the request in the first place.

That seems like a much simpler and elegant solution!

ModEnter:
Hello, this method is interesting, I'm going to dig that.

ModEnter:
@dokterbob I restored the previous behaviour and added prompt=consent to the authorization params for each one of the 9 chainlit external providers, as refered [here](https://openid.net/specs/openid-connect-core-1_0.html)

dokterbob:
> here

Great! Could you please indicate to what extent you manually tested it? It's essential for this we do extensive manual testing!

ModEnter:
Sure, I only tested it on github oauth, but since it is defined in the OpenID connect core 1.0 spec, it should be sure that it work with the other one. If you want to test them, it would be great, I personnaly do not have the possibility to do so...

dokterbob (changes requested):
@ModEnter I've just tested your branch with our own Descope integration and have unable to verify the solution. The `prompt=consent` parameter is correctly passed, yet it does not seem to be respected by Descope's `oauth2`'s authorize endpoint.

Descope's docs (for my own reference): https://docs.descope.com/getting-started/oidc-endpoints#guide-to-using-oidc-endpoints. For obvious reasons, I'm not going to paste a full breakdown of URL parameters, but it does seem the consent prompt request is passed along the login flow, so it might be a bug on their side. It's interesting to note that the login request is entirely absent from their login page.

After changing it to `prompt=login`, I got a login page.

Similarly, I've found that `prompt=select_account` does not have any effect. 

Hence, for now, I'd suggest that _at least_ for Descope, we use `prompt=login`. Alternatively, as you've verified GH to function, we could use `prompt=login` everywhere except for GH.

ModEnter:
hello @dokterbob, for now we are just going to follow your temporary solution, `prompt=login` everywhere except for GH. Can you commit that or do you want me to do so ? Dan wants me to close this issue ASAP

dokterbob (approved):

Issue #1355, opened 2024-09-18
Mixed dependencies in the Storage Clients module for the Data Layer


hadarsharon:
**Describe the bug**
When importing a certain storage client for usage in the data layer (e.g. `S3StorageClient`), one might encounter import errors due to mixed dependencies in the module, such as: **Azure** packages missing.

**To Reproduce**

1. Install Chainlit with relevant packages for using the `S3StorageClient` for the data layer (i.e. `boto3`)
2. Make sure you have no Azure packages installed in your Python runtime env. 
3. Try importing the client with `from chainlit.data.storage_clients import S3StorageClient`
4. You should encounter import errors because the `storage_clients` module has top-level imports for Azure:
```python
import boto3  # type: ignore
from azure.storage.filedatalake import (
    ContentSettings,
    DataLakeFileClient,
    DataLakeServiceClient,
    FileSystemClient,
)
```
**(Note the top-level import of both boto3 and azure.storage.filedatalake within the module, even though we only choose one client out of the two)**
6. Same issue would apply if you had Azure dependencies installed, but not S3 (`boto3`) 

**Expected behavior**
The storage client should be importable and ready for instantiation based on the type you require, without having to install all other storage clients' dependencies. 

**Desktop (please complete the following information):**

- OS: iOS
- Version: 1.2.0

---

I can imagine that splitting the `storage_clients` module to separate sub-modules based on the specific client type (e.g. `S3`, `Azure`, `GCS` etc.) would resolve this issue, as each would import **only** what that specific client needs.

ndricca:
A simple workaround is to just create a new script for only S3StorageClient and import it from there into your main app.py file.
I could do a PR to handle this maybe in the next days.

hadarsharon:
> > I can imagine that splitting the storage_clients module to separate sub-modules based on the specific client type (e.g. S3, Azure, GCS etc.) would resolve this issue, as each would import only what that specific client needs.
> 
> I have done this locally (creating a new script for only S3StorageClient and import it from there into your main app.py file) and it works like a charm. I could do a PR to handle this maybe in the next days.

Yes, I have done so too and it works fine, but I imagine some people would not appreciate having to do that, or not realize that this is a solution. 

Bunching all the imports together in the same module for all the storage clients just seems like a simple oversight to me. I'm not sure why the data layers were properly distributed to modules (e.g. dynamodb in its own module) but the storage clients weren't.

dokterbob:
[...]

PR #1363, opened 2024-09-21
solving issue #1355 - Mixed dependencies in the Storage Clients module for the Data Layer 

ndricca:
PR to handle issue #1355 

I have added also a pytest for S3StorageClient and moto to mock aws connection. I am not so experienced with Azure so I did not add a test on it.

### DIFF
--- a/backend/tests/data/test_sql_alchemy.py
+++ b/backend/tests/data/test_sql_alchemy.py
@@ -6,7 +6,7 @@ from sqlalchemy import text
 from sqlalchemy.ext.asyncio import create_async_engine
 
 from chainlit import User
-from chainlit.data.base import BaseStorageClient
+from chainlit.data.storage_clients.base import BaseStorageClient
 from chainlit.data.sql_alchemy import SQLAlchemyDataLayer
 from chainlit.element import Text
 
--- a/backend/chainlit/data/base.py
+++ b/backend/chainlit/data/base.py
@@ -105,17 +105,3 @@ class BaseDataLayer(ABC):
     @abstractmethod
     async def build_debug_url(self) -> str:
         pass
-
-
-class BaseStorageClient(ABC):
-    """Base class for non-text data persistence like Azure Data Lake, S3, Google Storage, etc."""
-
-    @abstractmethod
-    async def upload_file(
-        self,
-        object_key: str,
-        data: Union[bytes, str],
-        mime: str = "application/octet-stream",
-        overwrite: bool = True,
-    ) -> Dict[str, Any]:
-        pass
--- a/backend/chainlit/data/dynamodb.py
+++ b/backend/chainlit/data/dynamodb.py
@@ -12,7 +12,8 @@ import aiohttp
 import boto3  # type: ignore
 from boto3.dynamodb.types import TypeDeserializer, TypeSerializer
 from chainlit.context import context
-from chainlit.data.base import BaseDataLayer, BaseStorageClient
+from chainlit.data.base import BaseDataLayer
+from chainlit.data.storage_clients.base import BaseStorageClient
 from chainlit.data.utils import queue_until_user_message
 from chainlit.element import ElementDict
 from chainlit.logger import logger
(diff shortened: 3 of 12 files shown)


hadarsharon:
Thanks for implementing this @ndricca 🥇

ndricca:
i noticed that @dokterbob said "I imagine `storage_clients` module in the data module with a `base.py` and files for each implementation" while I was still using the  `base.py` from parent folder, so i splitted also this script moving the base method into `storage_clients` submodule

dokterbob:
Can't wait to review this (prolly around Wednesday in the coming week), unless @willydouhard beats me to it!

gsornsen:
Oh, nice! Looking forward to trying this out and dropping the need for a supervisor to start up moto-server and create a local s3 bucket in my current implementation.

dokterbob:
> @dokterbob am i suppose to push also the poetry.lock file? I see fails refer to this. I ran all poetry tests in local and they worked

We use the lockfile to make sure we're all dev'ing in exactly the same environment, creating more reproducible results.

ndricca:
> > @dokterbob am i suppose to push also the poetry.lock file? I see fails refer to this. I ran all poetry tests in local and they worked
> 
> We use the lockfile to make sure we're all dev'ing in exactly the same environment, creating more reproducible results.

I have merged incoming commits from main and pushed poetry.lock, now it should work

backend/tests/data/test_s3_storage_client.py
dokterbob, line 1: Please, follow the same import path pattern in test layout. So `/data/storage_clients/test_s3.py`.

backend/tests/data/test_sql_alchemy.py
dokterbob, line 1: It doesn't seem changes to this file are relevant to this PR. Things like these can easily create merge conflicts! 

Please `git checkout main -- backend/tests/data/test_sql_alchemy.py` this one. :)

ndricca: i did this, than changed other things and pushed but i see that this file is still here 🤔

ndricca: no, can you please tell me what do you want me to do? a git revert to exclude it? please tell me the proper commands, i don't want to make noise in the repo.

moreover, i have also updated `cypress/e2e/custom_data_layer/sql_alchemy.py` since it uses the `AzureStorageClient,` it that ok or not?

ndricca: no, can you please tell me what do you want me to do? a git revert to exclude it? please tell me the proper commands, i don't want to make noise in the repo.

moreover, i have also updated `cypress/e2e/custom_data_layer/sql_alchemy.py` since it uses the `AzureStorageClient`, is that ok or not?

ndricca:
@dokterbob have you had any chance to read my last comments? Can you tell me what exactly do you want me to do about git commits and how to proceed so that we can close this issue?

backend/tests/data/test_s3_storage_client.py
dokterbob: `git mv backend/tests/data/test_s3_storage_client.py backend/tests/data/test_s3.py` then rolling a commit will do it!

backend/tests/data/test_sql_alchemy.py
dokterbob: Forgive my confusion, the `git mv` comment was meant for the other one above (where I spelled it out).

Here, you can reset the file to the original state with `git checkout origin/main -- backend/tests/data/test_sql_alchemy.py`.

dokterbob (changes requested):
Sorry for the delay. Please see my specific instructions and resolve merge conflicts.

The conflict in `poetry.lock`, you can resolve by (removing other conflicts in the file and) removing the last `content-hash` line altogether. Then, run:
`poetry lock --no-update`

That will regenerate the content-hash and you can `git add poetry.lock`.

Let me know if you get stuck, happy to support! :)

dokterbob (approved):

PR #1367, opened 2024-09-22
Python fixes 2

EWouters:
Second pass of Python fixes with the [Trunk](https://trunk.io/) linters.

Not sure if the exception handling in aac3d501a655489b92cea96d7a19b10767cf8d26 will lead to the same behavior.

There are some warnings left with respect to exception handling, but I will leave them for now.

### DIFF
--- a/backend/chainlit/callbacks.py
+++ b/backend/chainlit/callbacks.py
@@ -16,7 +16,7 @@ from starlette.datastructures import Headers
 
 @trace
 def password_auth_callback(
-    func: Callable[[str, str], Awaitable[Optional[User]]]
+    func: Callable[[str, str], Awaitable[Optional[User]]],
 ) -> Callable:
     """
     Framework agnostic decorator to authenticate the user.
@@ -38,7 +38,7 @@ def password_auth_callback(
 
 @trace
 def header_auth_callback(
-    func: Callable[[Headers], Awaitable[Optional[User]]]
+    func: Callable[[Headers], Awaitable[Optional[User]]],
 ) -> Callable:
     """
     Framework agnostic decorator to authenticate the user via a header
@@ -177,7 +177,7 @@ def set_chat_profiles(
 
 @trace
 def set_starters(
-    func: Callable[[Optional["User"]], Awaitable[List["Starter"]]]
+    func: Callable[[Optional["User"]], Awaitable[List["Starter"]]],
 ) -> Callable:
     """
     Programmatic declaration of the available starter (can depend on the User from the session if authentication is setup).
@@ -245,7 +245,7 @@ def on_audio_end(func: Callable) -> Callable:
 
 @trace
 def author_rename(
-    func: Callable[[str], Awaitable[str]]
+    func: Callable[[str], Awaitable[str]],
 ) -> Callable[[str], Awaitable[str]]:
     """
     Useful to rename the author of message to display more friendly author names in the UI.
--- a/backend/chainlit/data/sql_alchemy.py
+++ b/backend/chainlit/data/sql_alchemy.py
@@ -7,7 +7,6 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
 import aiofiles
 import aiohttp
-
 from chainlit.data.base import BaseDataLayer, BaseStorageClient
 from chainlit.data.utils import queue_until_user_message
 from chainlit.element import ElementDict
--- a/backend/chainlit/discord/__init__.py
+++ b/backend/chainlit/discord/__init__.py
@@ -2,5 +2,5 @@ import importlib.util
 
 if importlib.util.find_spec("discord") is None:
     raise ValueError(
-        "The discord package is required to integrate Chainlit with a Slack app. Run `pip install discord --upgrade`"
+        "The discord package is required to integrate Chainlit with a Discord app. Run `pip install discord --upgrade`"
     )
--- a/backend/chainlit/discord/__init__.py
+++ b/backend/chainlit/discord/__init__.py
@@ -1,6 +1,6 @@
-try:
-    import discord
-except ModuleNotFoundError:
+import importlib.util
+
+if importlib.util.find_spec("discord") is None:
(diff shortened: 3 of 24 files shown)


backend/chainlit/callbacks.py
```diff
@@ -16,7 +16,7 @@
 
 @trace
 def password_auth_callback(
-    func: Callable[[str, str], Awaitable[Optional[User]]]
+    func: Callable[[str, str], Awaitable[Optional[User]]],
```
dokterbob, line 19: Why the `,` here?

backend/chainlit/data/sql_alchemy.py
```diff
@@ -7,7 +7,6 @@
 
 import aiofiles
 import aiohttp
-
```
dokterbob, line 10: Why remove this whitespace? I feel it's often indicative to separate:
1. Python stdlib imports
2. Third-party deps
3. Internal deps

backend/chainlit/discord/__init__.py
```diff
@@ -1,6 +1,6 @@
-try:
-    import discord
-except ModuleNotFoundError:
+import importlib.util
+
+if importlib.util.find_spec("discord") is None:
     raise ValueError(
         "The discord package is required to integrate Chainlit with a Slack app. Run `pip install discord --upgrade`"
```
dokterbob, line 5: Unrelated, but did someone forget to change Slack to Discord here? @willydouhard

backend/chainlit/step.py
```diff
@@ -140,7 +140,7 @@ def sync_wrapper(*args, **kwargs):
                     try:
                         if result and not step.output:
                             step.output = result
-                    except:
+                    except Exception as e:
                         step.is_error = True
                         step.output = str(e)
```
dokterbob, line 145: Wow, was this a broken exception handler here?

backend/tests/data/conftest.py
```diff
@@ -1,7 +1,6 @@
-import pytest
-
 from unittest.mock import AsyncMock
 
+import pytest
```
dokterbob, line 3: Why was this changed?

backend/tests/data/test_sql_alchemy.py
```diff
@@ -1,17 +1,13 @@
...
-
-from chainlit.data.base import BaseDataLayer, BaseStorageClient
+from chainlit import User
+from chainlit.data.base import BaseStorageClient
 from chainlit.data.sql_alchemy import SQLAlchemyDataLayer
 from chainlit.element import Text
-from chainlit import User
-from chainlit.user import PersistedUser
+from sqlalchemy import text
+from sqlalchemy.ext.asyncio import create_async_engine
```
dokterbob, line 10: See earlier comments on preferred import order. If we're not following that, I'd like to hear your suggestion.
It's a good time to start getting consistent on this (more important than 'my own way'). :)

dokterbob (commented):
There was a lot of changes to import order.

Without those, apart from my few comments/questions, I think I'm happy to merge.

With those, I'd like to first agree on how we sanitise imports from now on, making whatever we stick to a policy so I won't have to deal with reviewing import statement reorders again. ;)

My personal preference is:
1. Stdlib imports (as modules, so `import datatime` over `from datetime import datetime`.
2. 3rd party deps
3. Internal deps.

Beyond that, I guess alphabetical sorting is nice. But none of that is essential. Consistency is. :)

backend/chainlit/callbacks.py
EWouters: That is the result of running `ruff format`.

backend/chainlit/step.py
EWouters: Indeed, this was introduced in [4e0baab](https://github.com/Chainlit/chainlit/commit/4e0baabe3fd48abdd2f9998395b02c40e1f3d9e4#diff-4677e8359efad4a24ef6e70a44fd94c6549e46d017c3a0a9692bfd7d890bb75eR115).

backend/chainlit/data/sql_alchemy.py
EWouters: That is the result of running `ruff format`.

backend/tests/data/test_sql_alchemy.py
EWouters: That is the result of running the Trunk formatter, which uses ruff, isort and black to organize the imports. I'm trying to move everything to be consistent. It is not my own way.

backend/tests/data/conftest.py
EWouters: That is the result of running the Trunk formatter.

EWouters:
Thanks for your reply. For all the whitespace and import changes I used formatters to make the changes. None of these are my own preference. Moving to consistency is the goal.

You can pick which formatter you like, for instance:
- `ruff check --select I --fix`
- `isort .`
- `black .`

Trunk uses a combination of these. With the current changes all files pass the trunk linters. If you rather choose one of the above, I can change the imports to the order of that formatter.

backend/chainlit/discord/__init__.py
EWouters: Yes, unrelated, but fixed in c9aaa4b9d247a4435056495ae1d86ea6f565f180.

dokterbob:
Thanks for the feedback! 

I think what I missed, specifically, is having Python stdlib imports first, then other imports and finally internal/project imports.

AFAIK `ruff check --select I --fix` does this (but please check, it might be that we need to update settings in `pyproject.toml`.

Looking forward to fixups in that as well as resolution of the merge conflict in the SQLAlchemy test.

Once again, for the future; (several) smaller PR's (e.g. only doing one of these things) are easier to review and thus faster to merge (and less merge conflicts, as well). :)

dokterbob:
@EWouters Didn't want to pass you by but also wanted to 'clean house' in terms of PR's, so I rebased your PR minus formatting (prefer to first have clear standards, enforced on a CI-level before heading that way).

PR #1368, opened 2024-09-22
filter feedback 0 for nevigate feedback

zhiwehu:
if filter the nevigate (value is 0), it'll make feedback value to None then it'll filter all the threads.


dokterbob:
Thanks for the contrib! @qtangs is currently working on a full refactor of the SQLAlchemy stuff, I already asked him to take this into account. It would be good to have unittest for this as well.

dokterbob:
@dosu Is there perhaps a more elegant and readable way to implement the same functionality?

dokterbob (changes requested):
TLDR; Please, try and add a regression test.

We now have [unit tests for SQLAlchemy, based on SQLite](https://github.com/Chainlit/chainlit/blob/main/backend/tests/data/test_sql_alchemy.py). 

As such, would you be able to write a regression test, forked off main, which will fail without your patch, such that if you cherry-pick it into your PR, the tests pass? That way we can be 100% sure what the reproduction case is, that your patch will fix it _and that it won't come back in the future_?

If you do not have experience with unit tests, I recommend Claude to help writing them!

Thanks!

willydouhard:

Issue #1350, opened 2024-09-17
Include a pandas dataframe element


tituslhy:
**Is your feature request related to a problem? Please describe.**
Chainlit offers many elements that can be tagged to a message. I would like to suggest loading a pandas dataframe element because the response of LLMs can be made more neat with a table.

**Describe the solution you'd like**
Inclusion of a pandas dataframe element.

**Describe alternatives you've considered**
Streamlit has this feature https://docs.streamlit.io/develop/api-reference/data/st.dataframe

It would be great if chainlit has it too.

hadarsharon:
Unfortunately you are not the first to have requested this feature. 
For now I can only suggest using something like this (assuming your dataframe is stored in a variable `df`):

```python
await cl.Message(
    content="Data Table",
    elements=[cl.Text(content=df.to_markdown(index=False), display="inline")],
).send()
```

I also believe native support for DataFrames should be added, just like in Streamlit, because Markdown tables don't look half as good and aren't interactive. 
I've also tried something similar with `df.to_html` and `language="html"` in the cl.Message element, but:

1. It requires rendering unsafe HTML which is a potential security concern 
2. Still doesn't look that nice, not like Streamlit at least

desertproject:
I'm currently working on a draft pull request to implement this feature. The implementation will utilize the [MUI X Data Grid](https://mui.com/x/react-data-grid/) component. I'll update this issue with more details and a link to the PR once it's available for feedback.

tituslhy:
Looks great!

tituslhy:
Can I check if users can interact with the data frame like in streamlit - if they click the column name does it sort the data frame?

desertproject:
> Can I check if users can interact with the data frame like in streamlit - if they click the column name does it sort the data frame?

Yes, users can interact with the data frame similarly to how they would in Streamlit, including clicking the column name to sort it. You can explore additional interactive features and functionalities in this library [here](https://mui.com/x/react-data-grid/#mit-license-free-forever).

sundar-sarvam:
When will a tag for this release be made?

PR #1373, opened 2024-09-24
Add support for displaying pandas DataFrame as a table

desertproject:
This is an attempt to address issue #1350, and it may serve as a basis for future improvements and better implementations.

To use this functionality, follow these steps:

1. Convert the DataFrame to JSON: Use the to_json() method with the "split" orientation to convert the DataFrame into a JSON string, formatting the data for display.
2. Create a DataFrame Element: Pass the JSON string as the content parameter when creating the DataFrame element.
3. Append and Send: Append the DataFrame element to a message and send it for display.
    
```python
import chainlit as cl
import pandas as pd


@cl.on_chat_start
async def start():
    iris = pd.read_csv(
        "https://raw.githubusercontent.com/mwaskom/seaborn-data/master/iris.csv"
    )

    json_dataframe = iris.to_json(orient="split")

    elements = [
        cl.Dataframe(content=json_dataframe, display="inline", name="Dataframe")
    ]

    await cl.Message(content="This message has a Dataframe", elements=elements).send()

```

### DIFF
--- a/backend/chainlit/__init__.py
+++ b/backend/chainlit/__init__.py
@@ -23,6 +23,7 @@ from chainlit.context import context
 from chainlit.element import (
     Audio,
     Component,
+    Dataframe,
     File,
     Image,
     Pdf,
--- a/backend/chainlit/element.py
+++ b/backend/chainlit/element.py
@@ -31,7 +31,16 @@ mime_types = {
 }
 
 ElementType = Literal[
-    "image", "text", "pdf", "tasklist", "audio", "video", "file", "plotly", "component"
+    "image",
+    "text",
+    "pdf",
+    "tasklist",
+    "audio",
+    "video",
+    "file",
+    "plotly",
+    "dataframe",
+    "component",
 ]
 ElementDisplay = Literal["inline", "side", "page"]
 ElementSize = Literal["small", "medium", "large"]
@@ -355,6 +364,13 @@ class Plotly(Element):
         self.mime = "application/json"
 
         super().__post_init__()
+
+
+@dataclass
+class Dataframe(Element):
+    type: ClassVar[ElementType] = "dataframe"
+
+    size: ElementSize = "large"
 
 
 @dataclass
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -20,6 +20,7 @@
     "@mui/icons-material": "^5.14.9",
     "@mui/lab": "^5.0.0-alpha.122",
     "@mui/material": "^5.14.10",
+    "@mui/x-data-grid": "^6.20.4",
     "formik": "^2.4.3",
     "highlight.js": "^11.9.0",
     "i18next": "^23.7.16",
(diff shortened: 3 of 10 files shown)


hadarsharon:
@desertproject Nice one :)

One question though: Instead of expecting (forcing) the user to serialize the dataframe to JSON in a specific orientation, why don't we just accept the DataFrame element and call `to_json` with `orient="split"` ourselves behind the scenes, for example in the `__post_init__` method?

This way people wouldn't have to serialize/deserialize their dataframes throughout and could just pass a DataFrame at any point without having to worry about the underlying implementation.

dokterbob:
@desertproject Cool feature! Could you please create an E2E test for it, so we can support it towards the future? I'd also love some screenshots/screengrab of how it works is supposed to work and/or a cookbook entry. 🙏🏼 🥺

desertproject:
> Instead of expecting (forcing) the user to serialize the dataframe to JSON in a specific orientation, why don't we just accept the DataFrame element and call `to_json` with `orient="split"` ourselves behind the scenes, for example in the `__post_init__` method?

@hadarsharon That's a great idea! I had thought of something similar but wasn’t sure exactly where it would fit. Using the `__post_init__` method for this makes perfect sense. Thank you for the suggestion!

desertproject:
> Could you please create an E2E test for it, so we can support it towards the future? I'd also love a cookbook entry. 🙏🏼 🥺

@dokterbob I don’t have any experience with E2E testing or Cypress, but I’d be happy to give it a try! Could you provide some guidance on the specific tests you’d like to see? I can also take care of the cookbook entry without any problem.

dokterbob:
> > Could you please create an E2E test for it, so we can support it towards the future? I'd also love a cookbook entry. 🙏🏼 🥺
> 
> @dokterbob I don’t have any experience with E2E testing or Cypress, but I’d be happy to give it a try! Could you provide some guidance on the specific tests you’d like to see? I can also take care of the cookbook entry without any problem.

Basically, you'd create a test suite similar to the ones already there: https://github.com/Chainlit/chainlit/tree/main/cypress/e2e

Create a test project setup to use your feature, then test everything you'd normally test for it. E.g. return a dataframe, ensure it's properly rendered. If there's interaction possible, simulate it. 

I highly recommend to use something like Claude if you haven't written tests before (but of course, check the work for common slipups!).

Similarly unit tests for Python code would be great as well. Note how the unit tests are laid out to match the Python modules.

dokterbob (approved):
Definitely looks good to me.

I'd like to ask @willydouhard to take a quick look at it as well, other than that we're good to merge!

sundar-sarvam:
When will a new tag for this change be released? Eager to try it out:)

dokterbob:
@sundar-sarvam Should be in today's 2.0.dev1 pre-release.

@desertproject We'd love a PR also for the docs https://github.com/Chainlit/docs!

desertproject:
I'll take care of it!

wac4s:
hi everyone, it seems like i cant use the cl.Dataframe, 

  File "E:\project\.venv1\Lib\site-packages\chainlit\utils.py", line 70, in __getattr__
    module_path = registry[name]
                  ~~~~~~~~^^^^^^
KeyError: 'Dataframe'


can anyone help?

PR #1374, opened 2024-09-25
fix: do not block entire program with sleep

sandangel:

with time.sleep(), the entire program is stopped, which yield very bad performance for server handling concurrent requests.

Introducing new `async init` method for creating Step and Message with `asyncio.sleep` so that the async loop will give back the control to the program while waiting for sleep.

Let me know what is your thought and I help to replace existing usage with the new one.


dokterbob:
> Signed-off-by: San Nguyen [vinhsannguyen91@gmail.com](mailto:vinhsannguyen91@gmail.com)
> 
> with time.sleep(), the entire program is stopped, which yield very bad performance for server handling concurrent requests.
> 
> Introducing new `async init` method for creating Step and Message with `asyncio.sleep` so that the async loop will give back the control to the program while waiting for sleep.
> 
> Let me know what is your thought and I help to replace existing usage with the new one.

Why is this sleep statement there in the first place? @willydouhard

backend/chainlit/message.py
```diff
@@ -206,6 +206,37 @@ class Message(MessageBase):
...
+            elements,
+            type,
+            metadata,
+            tags,
+            id,
+            parent_id,
+            created_at,
+        )
+        return self
+
```
dokterbob, line 239: I don't see where this `init()` is called, here or anywhere. Could you please explain why this is here?

dokterbob: Also note the type error raised by mypy.

sandangel: @dokterbob , as mentioned, if sleep is needed then we will need to initiate all Message or Step with `init`, which will be lots of changes. May I confirm on your side that sleep is absolutely needed before making a big change?

dokterbob: I don't know why the sleep statement is here. Perhaps @willydouhard does?

I am asking about init() because you add this code to this PR and it does not seem to be called anywhere. I'd like to avoid dead code as much as possible.

sandangel:
Hi, is there any update so  we can just remove the sleep statement?

sandangel:
hi @willydouhard , can we have your comment?

willydouhard:
This is a hacky way of making sure the start date of a step is < than the end date. Not possible to use async version everywhere as of know since we use it in the constructor of the Message class (which can't be async). I'd like to see this go away. Will think about it!

willydouhard:
For now I merged this https://github.com/Chainlit/chainlit/pull/1634

PR #1379, opened 2024-09-27
feat: chainlit option to require websocket

sandangel:

When socket.io polling try upgrading to websocket transport, and it fails, it does not retry. This PR fix that by reinitiate the connection.


dokterbob (commented):
@sandangel Thanks for the contrib! Could you please provide a test app or some code in which I can replicate the problem as well as the fix?

In addition, the only place where I could find any docs mentioning `upgradeError` is here: https://www.npmjs.com/package/engine.io-client

Whilst the library we're directly using (socket.io) has the following to say on connection state recoveries: https://socket.io/docs/v4/connection-state-recovery

Could you, perhaps in comments, refer to where or how you derived this method (in addition to helping us replicate and test this)? 

This helps us to make sure we're not unintentionally creating problems while resolving them, or implement solutions which inadvertently do not fix the underlying issue.

sandangel:
@dokterbob , it is in the source code. https://github.com/socketio/socket.io/blob/d0fc72042068e7eaef448941add617f05e1ec236/packages/engine.io-client/lib/socket.ts#L1048

I think we will also need to disconnect and then reconnect to make sure we don't leak any socket.

sandangel:
You can use any current chainlit app and use transports=["polling"] for socket.io.AsyncServer. you will see request upgrade error and then frontend keep http polling. If the request is successful then you don't see any further http polling

sandangel:
In the future we may also consider use websocket only to skip the sticky session

dokterbob:
> @dokterbob , it is in the source code. https://github.com/socketio/socket.io/blob/d0fc72042068e7eaef448941add617f05e1ec236/packages/engine.io-client/lib/socket.ts#L1048
> 
> I think we will also need to disconnect and then reconnect to make sure we don't leak any socket.

It seems unwise to me to rely on undocumented functionality. It will be hard for us to support that, moving forward. If it's the only solution, at least we should have test coverage for it.

What's the reason we cannot use the documented approach suggested here?
https://socket.io/docs/v4/connection-state-recovery

sandangel:
Hi, I think this page better explain how it works: https://socket.io/docs/v4/how-it-works/#upgrade-mechanism

The reason we can not use https://socket.io/docs/v4/connection-state-recovery is that the connection is not failing, it just fails to upgrade, so we won't get any event except the upgradeError event.

In this case, it will keep using http polling transport.

dokterbob:
> https://socket.io/docs/v4/how-it-works/#upgrade-mechanism

So what your patch is doing (please bare with me) is retrying to upgrade to websocket instead of keeping polling if the websocket upgrade fails?

Is this not an issue other users of this library have (fixed)?

sandangel:
> > https://socket.io/docs/v4/how-it-works/#upgrade-mechanism
> 
> So what your patch is doing (please bare with me) is retrying to upgrade to websocket instead of keeping polling if the websocket upgrade fails?
> 
> Is this not an issue other users of this library have (fixed)?

This is correct. The only way to retry upgrade is to reconnect. This will not be an issue for users if they are using persistent sticky session. In case they are using soft session affinity in case of `istio`, then sometimes upgrade request will fail. Actually I can introduce a parameter to connect function to enable this behavior instead of having it "on" for all users.

sandangel:
@dokterbob  I updated the PR to have an option to use websocket. This should not break any existing app.

dokterbob:
> @dokterbob I updated the PR to have an option to use websocket. This should not break any existing app.

Thanks for taking the time explaining this.

I've looked at your changes and have the following questions:

1. I see you added a `useWebSocket` parameter, but I don't see how this is used or where this is referred. If we're adding a feature, it should be clear both in documentation and code how, why and under what conditions it is to be used.
2. While the name of the parameter is `useWebSocket`, the comment mentions 'Require WebSocket when connecting to backend'. If we simply changed the name to `requireWebSocket`, the code would be clearer and this comment would not be required.
3. However, your comment above "The only way to retry upgrade is to reconnect. This will not be an issue for users if they are using persistent sticky session. In case they are using soft session affinity in case of istio, then sometimes upgrade request will fail" is not trivial for any developer not fully versed in socketio and/or websockets, it seems particularly relevant to have as a comment!
4. My specific concern, as the API (the `upgradeError` event) is not documented anywhere (implying that it's not a public API) means that it may change at any time (e.g. semver doesn't apply), hence it could change without any announcement and break chainlit. As there's no unittests governing this code, we might not notice!

The last point is why I was asking specifically if there are other people having the same problem with this library, perhaps there are issues on the library's GH page, perhaps there is documentation somewhere, perhaps there's a stable/proper solution underway?

Thanks for finding the time to get into this!

sandangel:
https://github.com/socketio/socket.io/tree/main/packages/engine.io-client#events I think I found the document here. Let me update the PR.

sandangel:
Just updated the PR as discussed. Please have a look when you have time.

dokterbob (approved):

PR #1382, opened 2024-09-27
Don't run code on importing data module

dokterbob:
This avoids import loops and is a stepping stone towards pluggable data layers.


dokterbob:
@dosu Do you see any potential pitfalls or issues coming from this branch, looking at the overall code structure?

willydouhard (approved):

Issue #1394, opened 2024-10-02
Issues upgrading to 1.3.0rc0


hayescode:
**Describe the bug**
When upgrading from 1.1.404 to 1.3.0rc0 I have found the below issues -  [diff here](https://github.com/Chainlit/chainlit/compare/1.1.404...1.3.0rc0) - All of these worked properly in 1.1.404
- Step avatars disappeared
  - ![image](https://github.com/user-attachments/assets/ff990913-c827-430b-b2c5-4788120f1839)
- threads.userId changed breaking SQLAlchemyDataLayer NOTE: This only happens when steps are involved.
  - This issue is caused by a [new method that was added](https://github.com/Chainlit/chainlit/blob/79639b67adcc296430556e0e0d3f57dcdd49e984/backend/chainlit/data/sql_alchemy.py#L155-L165) where `userId` isn't encapsulated with double-quotes so the SQL engine converts `userId` --> `userid` which doesn't exist.
```sh
2024-10-02 12:12:23 - An error occurred: (sqlalchemy.dialects.postgresql.asyncpg.ProgrammingError) <class 'asyncpg.exceptions.UndefinedColumnError'>: column "userid" does not exist
HINT:  Perhaps you meant to reference the column "threads.userId".
[SQL: SELECT userId FROM threads WHERE id = $1]
[parameters: ('c15cf3b5-a38a-42cd-9e4e-82a8e69db495',)]
```

I can make a PR for the SQL Alchemy issue. @dokterbob @willydouhard can you advise on the step avatar issue?

hayescode:
@dokterbob @willydouhard The only Avatar related thing I could find in the diff was #1325 . I don't see an explanation for why this change was made in the PR and I'm not strong in Typescript/React so I'd like your feedback before I make a PR reverting the changes to `frontend/src/components/molecules/messages/Message.tsx`

AidanShipperley:
[...]

PR #1395, opened 2024-10-02
1.3.0rc0 Bug Fix | SQLAlchemy | Properly quote identifier

hayescode:
Fixes the SQLAlchemy bug raised in #1394 

Bug introduced in @1319 where `userId` wasn't encased with double-quotes so the SQL engine converts `userId` to `userid` which doesn't exist. Encasing with double-quotes fixes this issue.


dokterbob (approved):

dokterbob:
@hayescode Thanks for the fix and @desertproject thanks for the endorsement/testing.

PR #1402, opened 2024-10-04
feat: improve UX when loading images

fgalind1:
Currently when adding images in the conversation, if the image is big and is not that clear, clicking in the image will attempt to download the image.

This PR attempts to improve the UX a bit by just opening the image in a new window (aka popup) to see the image larger. It has a download option on this new popup as well to preserve the original download behavior, but now the UX is a bit better


ybilyk (approved):

przemien (approved):

dokterbob (approved):

fgalind1:
thanks @dokterbob for helping review and approve this and my last change as well 👍

Issue #1400, opened 2024-10-04
Tasklist not supported when mounting Chainlit as a fastAPI subapp


tioans:
**Describe the bug**
When the Chainlit app (v1.2.0) is mounted as a FastAPI sub-app using mount_chainlit, a TaskList element fails to load, displaying "An error occurred" in the top-right corner where the task list should be. However, the Chainlit app works correctly when run directly without mounting to FastAPI.

**To Reproduce**

1. Create a Chainlit app with the Tasklist example provided in the Chainlit documentation:

```
import chainlit as cl


@cl.on_chat_start
async def main():
    # Create the TaskList
    task_list = cl.TaskList()
    task_list.status = "Running..."

    # Create a task and put it in the running state
    task1 = cl.Task(title="Processing data", status=cl.TaskStatus.RUNNING)
    await task_list.add_task(task1)
    # Create another task that is in the ready state
    task2 = cl.Task(title="Performing calculations")
    await task_list.add_task(task2)

    # Optional: link a message to each task to allow task navigation in the chat history
    message = await cl.Message(content="Started processing data").send()
    task1.forId = message.id

    # Update the task list in the interface
    await task_list.send()

    # Perform some action on your end
    await cl.sleep(1)

    # Update the task statuses
    task1.status = cl.TaskStatus.DONE
    task2.status = cl.TaskStatus.FAILED
    task_list.status = "Failed"
    await task_list.send()

```

2. Mount the Chainlit app to a FastAPI app using mount_chainlit:

```
from fastapi import FastAPI
from chainlit.utils import mount_chainlit

app = FastAPI()

@app.get("/app")
def read_main():
    return {"message": "Hello World from main app"}

mount_chainlit(app=app, target="test_tasklist.py", path="/chainlit")
```

3. Run the FastAPI app with Uvicorn.

```
uvicorn test_mount:app --host 0.0.0.0 --port 80
```

**Expected behavior**
The TaskList element should load and display the tasks as expected, as it does when the Chainlit app is run directly without mounting to FastAPI.

**Screenshots**
![image](https://github.com/user-attachments/assets/372dbf88-f301-436b-b327-294608330fa7)

**Desktop (please complete the following information):**

- OS: Windows 11
- Browser: Firefox, Chrome, Edge

dokterbob:
@tioans Thanks so much for pointing this out! @willydouhard has created a PR attempting to fix this.

[...]

PR #1408, opened 2024-10-07
fix: tasklist when chainlit is submounted

willydouhard:
Fixing https://github.com/Chainlit/chainlit/issues/1400.

The issue is that the tasklist url is already containing the subpath.`useApi` will add the subpath again. So we just use `useSWR` raw here.


frontend/src/components/molecules/tasklist/TaskList.tsx
```diff
@@ -70,7 +73,7 @@ const TaskList = ({ isMobile }: { isMobile: boolean }) => {
     return parsedUrl.pathname + parsedUrl.search;
```
dokterbob, line 73: If you go this way, I think you can (and should) also try to remove the workaround here as well; before you'd get double base URL's (so you removed them).

Now, you are making a query for a relative URL. Might as well query the full URL and have a bit less code.

dokterbob (changes requested):
I've validated the fix! 👍🏼 

We get some free cleanup, see my other comment.

Would also be nice to remove changes from `socket.py` from this patch. Specifically, this allows us to backport it to 1.3.x without merge conflicts. Same for lock files, AFAIK they have nothing to do with this fix.

`git checkout main -- backend/chainlit/socket.py` is your friend. ;)

dokterbob:
@willydouhard Any chance you could continue with the other cleanup requested (e.g. removing unrelated changes)?

If not I can do it, but might not get to it until ~Monday.

PR #1409, opened 2024-10-07
fix: step arg input and concurrency

willydouhard:
1. Fixing concurrency issues where all steps would be accumulated in the last step. To reproduce:

```py
import chainlit as cl
from openai import AsyncOpenAI
import asyncio

cl.instrument_openai()

oai_client = AsyncOpenAI()


@cl.step(type="run")
async def agent(query: str):
   res = await oai_client.chat.completions.create(model="gpt-4o", messages=[{"role": "system", "content": "You are a chatbot."}, {"role": "user", "content": query}])
   return res.choices[0].message.content

@cl.on_message
async def f():
    coros = []
    for i in range(10):
        coros.append(agent(f"Hello {i}"))
    await asyncio.gather(*coros)
```

2. Step inputs with args (not kwargs) were not logged correctly
3. Removing useless `active_steps`. Now always using context var `local_steps`


dokterbob:
@dosubot Any suggestions how to regression test 1 and 2 and have basic test coverage for 3?

dokterbob:
Ok I think I might have replicated the 1st.

image

dokterbob (approved):
LGTM

PR #1419, opened 2024-10-09
Make Auto-Login Behavior Configurable

hayescode:
This PR makes #1362 configurable through a new `OAUTH_DISABLE_REAUTHENTICATION_ON_LOGOUT` environment variable. If this env var isn't set, or it's anything other than `true` then the current behavior in #1362 is used, so that developers don't need to do anything, but at least this behavior is controllable.

The authentication behavior before #1362 would auto-login users without user interaction. The behavior after #1362 is that my users need to re-enter their username/password and confirm 2FA on their phone. Furthermore my coporate users login to their device with a PIN so most don't remember their password. This has led to a poor experience.

Ideally, I'd like to see the behavior require the user to click the "Sign in with Microsoft" button (or different auth providers if there's multiple like #1362 suggests) to auto-login via SSO. I don't know where this auto-login trigger is coming from and how to remove, so I'm open to suggestions!
image

### DIFF
--- a/backend/chainlit/oauth_providers.py
+++ b/backend/chainlit/oauth_providers.py
@@ -39,6 +39,8 @@ class GithubOAuthProvider(OAuthProvider):
             "scope": "user:email",
             "prompt": "consent",
         }
+        if os.environ.get("OAUTH_DISABLE_REAUTHENTICATION_ON_LOGOUT", "false").lower() == "true":
+            self.authorize_params.pop("prompt", None)
 
     async def get_token(self, code: str, url: str):
         payload = {
@@ -98,6 +100,8 @@ class GoogleOAuthProvider(OAuthProvider):
             "access_type": "offline",
             "prompt": "login",
         }
+        if os.environ.get("OAUTH_DISABLE_REAUTHENTICATION_ON_LOGOUT", "false").lower() == "true":
+            self.authorize_params.pop("prompt", None)
 
     async def get_token(self, code: str, url: str):
         payload = {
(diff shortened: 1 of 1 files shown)


dokterbob:
Hi @hayescode, thanks for the feedback. I understand where you're coming from.

The UX issue we were solving was requiring users to log in again _after logging out_. The issue you seem to address is that users _always_ have to log in with the current way we addressed it.

Arguably, the proper solution would be for the frontend to _remember_ when a user logged out, only requiring a prompt if they did so. I'd much rather have this than adding an additional setting as a workaround.

Are you on Discord? I'm `mathijsfietst`. Perhaps it's a faster way to align on this or other issues.

hayescode:
@dokterbob thanks and I 100% agree that is the ideal solution, but I don't know enough about fastAPI auth to implement this. I'll start a chain on discord I'm Haze.

dokterbob:
> @dokterbob thanks and I 100% agree that is the ideal solution, but I don't know enough about fastAPI auth to implement this. I'll start a chain on discord I'm Haze.

It's kind of more of an oauth thing.

I'm really looking forward to removing all auth-specific integration stuff the hell away from chainlit. Way to big a scope to offer support for. 

@willydouhard Do you have any thoughts on how to approach this? It seems we fix an issue with a workaround on one side, we get another one in return...

dokterbob:
@hayescode Idea; how about we simply call this setting `OAUTH_PROMPT` and we pass it as-is to oauth implementations? That way we can quickly arrive at a good-enough compromise for everyone, the feature is clear (as it simply wraps standard oauth behaviour) and it would not necessarily be a workaround (as it'll still be functional once we've 'proper' oauth logout)?

We're doing 1.3.0 tomorrow, I'd be happy to include this.

dokterbob:
I'm actually working on a patch now. What I'd like to make sure is with 1.3.0 we don't actually break user's expectations. 

Hence, I'm thinking to go beyond your patch, actually default to not require consent (but allowing users to enable that).

dokterbob:
@hayescode Any chance you could take a look at #1456 to see whether it makes sense to you?

dokterbob:
Closing this in favour of #1456.

Issue #1370, opened 2024-09-24
Avatar image is broken when reading threads from history


LucasMagnum:
**Describe the bug**

The PR https://github.com/Chainlit/chainlit/pull/1326 introduces a bug when it expects the `get_avatar` to match string (https://github.com/Chainlit/chainlit/pull/1326/files#diff-9a73e67e163edc3fae6839bf2bcc8c6ab650624eff6668502388a7a86b7ece78R964) . 

The Frontend app can make requests to the avatar that do not match the Regex (https://github.com/Chainlit/chainlit/blob/main/frontend/src/components/molecules/messages/components/Avatar.tsx#L31) this creates a bug for existing applications where the `config.ui.name` contains space (e.g. `My Assistant`). 

**To Reproduce**
Set your config.ui.name to `My Assistant`

1. Create a thread
2. Publish messages
3. Create a new thread
4. Try to resume the previous thread
5. See the 400 error and the avatar not showing

**Expected behavior**
Avatar should be loaded for threads in history.

dokterbob:
@LucasMagnum Love to hear whether the fix suggested by our bot does the trick. If it does, we'd love a PR for that, preferably including a(n update to a) E2E test demonstrating the issue and it's resolution.

@dosu Any suggestion as to which is the most applicable place to write such a (regression) test, preferably without adding a new E2E (because they're slow)? What could such a test look like?

dokterbob:
> The most applicable place to write a regression test for the `get_avatar` function without adding a new E2E test is in the `backend/tests/test_server.py` file. This file already contains tests for the `get_avatar` function, so adding a regression test here would be consistent with the existing structure.
> 
> Here is an example of what such a regression test could look like:
> 
> ```python
> def test_get_avatar_regression(test_client: TestClient, monkeypatch: pytest.MonkeyPatch):
>     """Regression test for get_avatar function."""
>     # Setup: Create a custom avatar
>     custom_avatar_path = os.path.join(
>         APP_ROOT, "public", "avatars", "regression_avatar.png"
>     )
>     os.makedirs(os.path.dirname(custom_avatar_path), exist_ok=True)
>     with open(custom_avatar_path, "wb") as f:
>         f.write(b"regression test image data")
> 
>     # Test: Fetch the custom avatar
>     response = test_client.get("/avatars/regression_avatar")
>     assert response.status_code == 200
>     assert response.headers["content-type"].startswith("image/")
>     assert response.content == b"regression test image data"
> 
[...]

PR #1420, opened 2024-10-09
Sanitise avatar URL client-side

dokterbob:
More structural solution to #1370.

* 'Echo' E2E test based on [In Pure Python](https://docs.chainlit.io/get-started/pure-python) example from docs.[1]
* Assertions for avatar loading and URL.
* Sanitise avatar URL client-side, substituting anything other than `[a-zA-Z0-9]` with `_`.
* Disallow spaces again in avatar names (e.g. only allow 'clean' filenames).
* (Only) sanitise avatar filename when defaulting to config value.

[1] Instead of having chainlit start again and again, we can merge tests with identical configuration, significantly speeding up tests (chainlit takes several seconds every time to start). Making tests cheaper, will enable us to run and write more of them, faster, allowing for a more mature product.


frontend/src/components/molecules/messages/components/Avatar.tsx
```diff
@@ -28,7 +38,9 @@ const MessageAvatar = ({ author, hide }: Props) => {
     if (isAssistant && selectedChatProfile?.icon) {
       return selectedChatProfile.icon;
     }
-    return apiClient?.buildEndpoint(`/avatars/${author || 'default'}`);
+
+    const authorAvatarId = getAuthorAvatarId(author);
```
willydouhard, line 42: i don't think doing this in the client is a good idea since developers also use chainlit with custom frontends

dokterbob: Maybe we can give custom frontend implementors (2-line!) instructions to implement this?

This patch would not only improve security, but being consistent in how we deal with filenames will make the developer experience much more robust. This would replace a `.` (and other characters, think unicode) with `_`, allowing _any_ characters in avatar names.

We currently do not: a lot of author names will prevent the avatar from working. This patch solves that, fully, without compromising security again.

dokterbob: Proposed compromise:

1. We keep allowing spaces and upper characters (basically not change the backend).
2. We throw a deprecation error every time an 'invalid' URL comes in.
3. We update the frontend code already.

This would let existing users know of what's coming, recover pre-existing functionality in cases of special characters, guarantee sufficient security for now and maximal security long-term.

willydouhard (approved):

dokterbob:
I actually want to implement the compromise stated above before continuing. https://github.com/Chainlit/chainlit/pull/1420#discussion_r1799680649

Issue #230, opened 2023-07-26
findLastIndex error when loading hello in Firefox


piacentini:
findLastIndex is not a standard JS function, so Firefox can not load a Chainlit app. 
Attached is a screenshot, just run

chainlit hello

And open localhost:8000 in Firefox.
![Screenshot_20230726_164432](https://github.com/Chainlit/chainlit/assets/1039234/831c256c-b79a-4bf4-987b-bcd1c24b3313)

alimtunc:
Hi, it has been fixed recently and will be on the next release.

PR #1436, opened 2024-10-15
[Add] Implement Slack Socket Mode support

JCMarques15:
## Description
This PR introduces support for Slack Socket Mode, enabling real-time message processing using websockets. This feature enhances the Slack integration by allowing for more efficient and responsive communication, and a way to bypass ingress restrictive networks.

## Changes
- Added Slack Socket Mode handler in `server.py`
- Created `start_socket_mode` function in `slack/app.py`
- Implemented conditional logic to use Socket Mode when `SLACK_WEBSOCKET_TOKEN` is set
- Added an example script `slack_websocket_test.py` for testing the Slack websocket handler

## How to Test
1. Set the following environment variables:
   - `SLACK_BOT_TOKEN`
   - `SLACK_SIGNING_SECRET`
   - `SLACK_WEBSOCKET_TOKEN`
2. Run the application and verify that the Slack integration connects using Socket Mode
3. Test sending messages through Slack and ensure they are processed correctly

## Additional Notes
- The existing HTTP handler remains active when only `SLACK_BOT_TOKEN` and `SLACK_SIGNING_SECRET` are provided
- When `SLACK_WEBSOCKET_TOKEN` is set, the application will prioritize Socket Mode over the HTTP handler

### DIFF
--- a/backend/chainlit/slack_websocket_test.py
+++ b/backend/chainlit/slack_websocket_test.py
@@ -1,0 +1,18 @@
+# This is a simple example to test the slack websocket handler.
+# To initiate the websocket dont forget to set the variables:
+#   - SLACK_BOT_TOKEN
+#   - SLACK_SIGNING_SECRET
+#   - SLACK_WEBSOCKET_TOKEN <- this one dictates if websocket or http handler
+
+from chainlit import Message, on_message, user_session
+
+
+@on_message
+async def main(message: Message):
+    client_type = user_session.get("client_type")
+    if client_type == "slack":
+        user_email = user_session.get("user").metadata.get("email")
+        print(f"Received a message from: {user_email}")
+        await Message(
+            content=f"Hi {user_email}, I have received the following message:\n{message.content}",
+        ).send()
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -131,6 +131,13 @@ async def lifespan(app: FastAPI):
 
         discord_task = asyncio.create_task(client.start(discord_bot_token))
 
+    slack_task = None
+
+    # Slack Socket Handler if env variable SLACK_APP_TOKEN is set
+    if os.environ.get("SLACK_BOT_TOKEN") and os.environ.get("SLACK_WEBSOCKET_TOKEN"):
+        from chainlit.slack.app import start_socket_mode
+        slack_task = asyncio.create_task(start_socket_mode())
+
     try:
         yield
     finally:
@@ -143,6 +150,10 @@ async def lifespan(app: FastAPI):
             if discord_task:
                 discord_task.cancel()
                 await discord_task
+
+            if slack_task:
+                slack_task.cancel()
+                await slack_task
         except asyncio.exceptions.CancelledError:
             pass
 
@@ -230,10 +241,10 @@ app.mount(
 
 
 # -------------------------------------------------------------------------------
-#                               SLACK HANDLER
+#                               SLACK HTTP HANDLER
 # -------------------------------------------------------------------------------
 
-if os.environ.get("SLACK_BOT_TOKEN") and os.environ.get("SLACK_SIGNING_SECRET"):
+if os.environ.get("SLACK_BOT_TOKEN") and os.environ.get("SLACK_SIGNING_SECRET") and not os.environ.get("SLACK_WEBSOCKET_TOKEN"):
     from chainlit.slack.app import slack_app_handler
 
     @router.post("/slack/events")
--- a/backend/chainlit/slack/app.py
+++ b/backend/chainlit/slack/app.py
@@ -19,6 +19,7 @@ from chainlit.types import Feedback
 from chainlit.user import PersistedUser, User
 from chainlit.user_session import user_session
 from slack_bolt.adapter.fastapi.async_handler import AsyncSlackRequestHandler
+from slack_bolt.adapter.socket_mode.async_handler import AsyncSocketModeHandler
 from slack_bolt.async_app import AsyncApp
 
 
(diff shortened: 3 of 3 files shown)


backend/chainlit/slack_websocket_test.py
```diff
@@ -0,0 +1,18 @@
...
+
+@on_message
+async def main(message: Message):
+    client_type = user_session.get("client_type")
+    if client_type == "slack":
+        user_email = user_session.get("user").metadata.get("email")
+        print(f"Received a message from: {user_email}")
+        await Message(
+            content=f"Hi {user_email}, I have received the following message:\n{message.content}",
+        ).send()
```
dokterbob, line 18: It's really great that you've added this script to facilitate testing, but this really isn't the place for it!

Rather, we have several designated places for such functionality:
1. The chainlit [cookbook](https://github.com/Chainlit/cookbook/tree/main). It would be great to have a minimal example of slack bot integration there! In addition, it's also a great guidance for users to get started with this!
2. Unit tests. It would be absolutely amazing to have a unit test for Slack -- both the legacy behaviour as well as websocket behaviour. 
3. E2E tests, they live in `https://github.com/Chainlit/chainlit/tree/main/cypress/e2e`. But I don't think it's relevant for Slack as I don't think there's a browser UX in this case!

The way to test this is to [monkeypatch](https://docs.pytest.org/en/stable/how-to/monkeypatch.html) `chainlit.slack.app.AsyncSocketModeHandler` and `AsyncApp` with `spec=AsyncSocketModeHandler` and `AsyncApp`, respectively. 

You could then test:
1. Whether all Slack functions are properly called by calling assertions on the mock.
2. Where relevant, manually call Slack's callback hooks _as if_ the user sends messages. 

For now, just having _any_ testing for Slack integration would be a huuuuuuge contrib. If you can't make it, that's okay, just remove this test file and file a PR for the cookbook with it, as well as a PR for the docs discussing this new behaviour.

dokterbob (changes requested):
@JCMarques15 This is a great feature, thanks for the contrib!

I would really like to have at least basic unit test coverage here. That way, we can ensure that Slack integration stays stable going forward. Please let me know whether you're willing to give that a try! I recommend using Claude Sonnet to help along if you get stuck (but I'm happy to offer support as well!).

JCMarques15:
Hi @dokterbob, sorry for the late reply—it’s been a hectic few weeks!

Thanks for the detailed suggestions; they’re super helpful. I’ll give the unit testing a try and work on monkeypatching AsyncSocketModeHandler and AsyncApp to cover both legacy and websocket behavior. I’m still getting familiar with this kind of setup, so it might take me a bit of time, but I’m happy to dive in.

If I hit any roadblocks or it ends up taking longer than expected, I’ll focus on submitting a PR for the cookbook and documentation instead. Let me know if there’s anything else I should keep in mind as I get started!

willydouhard:
What is the status on this?

dokterbob:
@JCMarques15 Friendly poke. :)

JCMarques15:
My apologies for the delay. The end and beginning of the year have been quite demanding work-wise, and I wasn’t able to find the time to address this earlier.

I’ve now picked it up and am actively working on it. You can expect an update by the end of next week. Thank you for your patience!

JCMarques15:
Hi @dokterbob I have made the changes you had requested.

I’ve moved the testing script to the cookbook as a minimal Slack bot integration example and added documentation for the new behaviour in the docs repo.

Additionally, I’ve added unit tests for both the legacy and websocket Slack behaviours using `AsyncSocketModeHandler` and `AsyncApp` with mocks, as suggested. These tests assert that Slack functions are properly called and simulate relevant callback hooks.

I believe I’ve covered everything you outlined, but I’d really appreciate it if you could take a quick look to confirm whether this aligns with your expectations or if anything still needs to be adjusted.

The relevant PRs are:  
cookbook repository: [#195](https://github.com/Chainlit/cookbook/pull/195)  
docs repository: [#230](https://github.com/Chainlit/docs/pull/230)

gowtham3105:
@JCMarques15 @dokterbob What is the status on this ?

JCMarques15:
Hi @gowtham3105!

The PR is currently waiting for final review from the maintainers. I've added the requested unit tests and documentation, so from a technical standpoint it should be ready to go.

@dokterbob when you have a moment, would you be able to take a look? Happy to make any additional adjustments if needed.

Thanks for your interest in this feature!

gowtham3105:
Hi @hayescode @asvishnyakov 

This PR is pending for review since quite some time, can one of you please have a look ?

asvishnyakov:
@JCMarques15 Can you please fix the failing test and linting issues?

JCMarques15:
Hi @asvishnyakov. I have dealt with the unused import and weird space character. For the test error:

### Problem

The `test_slack_socket_mode.py` test was failing in CI/CD with `KeyError: 'slack'` but passing locally.

### Root Cause

The issue was with the lazy import registry system. When using:

```python
with patch("chainlit.slack.app.AsyncSocketModeHandler", autospec=True):
```

The `patch()` function tries to resolve the string path by accessing `chainlit.slack` as an attribute, which triggers the `__getattr__` method in `chainlit/__init__.py`. However, `slack` is not included in the lazy import registry, causing a `KeyError`.

### Solution

I changed the test to use object-based patching instead of string-based patching:

```python
# Import the module first
slack_app_mod = importlib.import_module("chainlit.slack.app")

# Patch the object directly
with patch.object(slack_app_mod, "AsyncSocketModeHandler", autospec=True) as handler_cls:
```

This bypasses the lazy import registry entirely by importing the module directly via `importlib.import_module()`.

An alternative would be to add `slack` to the lazy import registry in `chainlit/__init__.py`, similar to how other optional integrations are handled. However, I chose the direct import approach for this test to avoid potential side effects or changes to the main module's import behavior, especially since I'm not fully familiar with all the implications of modifying the registry.

I hope this fixes everything as requested.

JCMarques15:
Apologies, I hadn’t realized Ruff was being used as the formatter. My local setup was applying Black by mistake. I’ve corrected the formatting in the affected file, and it should now pass all CI/CD checks.

gowtham3105:
@asvishnyakov Can you please check the changes?
I want to use this in slack socket mode changes in my chatbot. This is hosted in our internal could so, exposing a public endpoint is not possible

Issue #1101, opened 2024-06-26
Files not protected when auth is enabled


puppetm4st3r:
**Describe the bug**
When you has chainlit configured with authentication, in an incognito browser you can freely access to temporal audio files generated in audio assistants, only putting the URL in the browser like: https://192.168.0.150:8888/project/file/16745f20-dddf-4cf6-84d5-6d424635c63b?session_id=e8bafcfa-ff64-4517-bdc6-8ceddc74e89d

**To Reproduce**
Inspect the audio control in an authenticated chainlit instalation, copy paste the source URL from the any audio control generated by your assistant, put the link like https://192.168.0.150:8888/project/file/16745f20-dddf-4cf6-84d5-6d424635c63b?session_id=e8bafcfa-ff64-4517-bdc6-8ceddc74e89d in an incognito tab or browser.

**Expected behavior**
Audio files are sensitive material, if auth on chainlit is activated, then audio files must be protected some way with the JWT token, nice to have: temporary generated audio files links, in order to expire at x minutes from generation.

- OS: Ubuntu 22
- Browser tested on chrome, firefox
- Version latest

**Smartphone (please complete the following information):**

Not tested

narayan-h:
Any update on this chainlit this is really, BIG Risk

dokterbob:
Really sorry, I must have missed this issue, which was created before I joined the project. Happy to see the fix, I want to try and get this solved before the next release (1.3.0).

PR #1438, opened 2024-10-16
fix(security): add auth to /project/file get endpoint

qvalentin:
fixes https://github.com/Chainlit/chainlit/issues/1101

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -878,7 +878,10 @@ async def upload_file(
 @router.get("/project/file/{file_id}")
 async def get_file(
     file_id: str,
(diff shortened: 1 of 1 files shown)


dokterbob:
Thanks @qvalentin for the report & fix! We'd like to take this along in the next release.

Any chance you could add a regression unittest demonstrating the issue and it's resolution?

dokterbob:
@qvalentin Just wrote regression test for this and several related issues. Hope to get fix in later today.

willydouhard:
This has been released in `2.0.0` as part of the cookie auth rework! Thank you for contributing

PR #1441, opened 2024-10-16
Test and resolve issue with get_file and upload_file

dokterbob:
Fixes https://github.com/Chainlit/chainlit/issues/1101 and includes and closes #1438.

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -878,7 +878,10 @@ async def upload_file(
 @router.get("/project/file/{file_id}")
 async def get_file(
     file_id: str,
-    session_id: Optional[str] = None,
+    session_id: str,
+    current_user: Annotated[
+        Union[None, User, PersistedUser], Depends(get_current_user)
+    ],
 ):
     """Get a file from the session files directory."""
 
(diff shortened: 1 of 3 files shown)


dokterbob:
Converted back to draft, there's still an issue I'd like to work on.

willydouhard (approved):

wenboown:
The commit e65f19 prevents elements from showing when the oauth are not setup. For example, I am using the simple password example in https://docs.chainlit.io/authentication/password for development, and got 401 Unauthorized on the frontend. 

Problem is in
@router.get("/project/file/{file_id}")

anyway to make it work for simple auth case as well? Thank you

hayescode:
> The commit e65f19 prevents elements from showing when the oauth are not setup. For example, I am using the simple password example in https://docs.chainlit.io/authentication/password for development, and got 401 Unauthorized on the frontend.
> 
> Problem is in @router.get("/project/file/{file_id}")
> 
> anyway to make it work for simple auth case as well? Thank you

@dokterbob Please take a look at this. This PR causes Elements to break. https://github.com/Chainlit/chainlit/issues/1472

PR #1455, opened 2024-10-21
Release prep 2.0.dev1

dokterbob:
Changelog and bump version.


hayescode:
@dokterbob suggestion: can some notes be added for the new cl.Dataframe() feature from #1373 ?

dokterbob:
> @dokterbob suggestion: can some notes be added for the new cl.Dataframe() feature from #1373 ?

Yeah we'll mention them in the release notes, the release announcement and will add them to the docs. We're not adding them to 1.3.0 though, it'll only be released as part of 2.0 or possibly 1.4.0. 1.3.0 is only fixups for 1.3.0rc0.

PR #1456, opened 2024-10-21
OAuth prompt configurable through environment

dokterbob:
Use `OAUTH_<PROVIDER>_PROMPT` or `OAUTH_PROMPT` to override behaviour.
E.g. `OAUTH_PROMPT=login` to always require login, allowing for re-login after logout, even when only a single oauth provider is configured, without breaking prior user expecations.

Supersedes #1419 and #1362, requires documentation. 

Would be great to get this into 1.3.0 as shipping as-is would break some SSO users and require release of `1.4.0` (e.g. another breaking change) to fix it.

### DIFF
--- a/backend/chainlit/oauth_providers.py
+++ b/backend/chainlit/oauth_providers.py
@@ -16,7 +16,7 @@ class OAuthProvider:
     client_secret: str
     authorize_url: str
     authorize_params: Dict[str, str]
-    default_prompt: Optional[str]
+    default_prompt: Optional[str] = None
 
     def is_configured(self):
         return all([os.environ.get(env) for env in self.env])
--- a/backend/chainlit/oauth_providers.py
+++ b/backend/chainlit/oauth_providers.py
@@ -16,6 +16,7 @@ class OAuthProvider:
     client_secret: str
     authorize_url: str
     authorize_params: Dict[str, str]
+    default_prompt: Optional[str]
 
     def is_configured(self):
         return all([os.environ.get(env) for env in self.env])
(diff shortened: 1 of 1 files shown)


dokterbob:
@ModEnter Any chance you're able to quickly test this by tomorrow?

hayescode:
@dokterbob Looks good! Since the different OAuth providers have different values for the `prompt` key, i wonder if the `OAUTH_PROMPT` should just be a boolean and the correct value is in oauth_providers.py?

GitHub OAuth has "consent" - [source](https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/oauth_providers.py#L40)

Most others have "login" - [source](https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/oauth_providers.py#L252)

dokterbob:
> @dokterbob Looks good! Since the different OAuth providers have different values for the `prompt` key, i wonder if the `OAUTH_PROMPT` should just be a boolean and the correct value is in oauth_providers.py?
> 
> GitHub OAuth has "consent" - [source](https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/oauth_providers.py#L40)
> 
> Most others have "login" - [source](https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/oauth_providers.py#L252)

The values are standardised but you can override them per-provider as well (because not all support all values).

willydouhard (approved):

eiribasjime:
Hi everyone,

@dokterbob I really appreciate the incredible job you are doing with Chainlit. I am facing some problems with oauth authentication. As I see, you have a solution for the issue that chainlit automatically logs in after logging out.

I have done some tests for a chatbot I am developing and I face the following problem:

After adding the environment variables for OAUTH_PROMPT, actually the chatbot waits for the user to click and log in another time. Everything works well if it is done in a short time since I have logged out. If a longer time has passed since that, an error of Unauthorized is shown. 

Personally I think this might be because, after logging out, it automatically starts the log in process (although it waits for the user interaction), which means an state is initialized. But, as the user waits for more time than the timeout of that state, it is not valid for the moment user wants to log in another time. I don't know how to do it but a possible solution could be that the auth callback automatically returns to the login page if something has been wrong.

And also after logging out I think there should be a button to click to: Log in with Google. This way we could avoid all the cases in which the user doesn't want to log in just after logging out, but maybe in the future. So, in that case, only after clicking: Log in with Google is when the process starts and the state is created and, eventually, user will log in just in that moment.

Thanks in advance for your help and hope you continue developing this.

Kind regards

dokterbob:
Thanks for your kind and supportive words @eiribasjime. Unfortunately, my direct collaboration with Chainlit has come to an end. However, I will be happy to offer support as a freelance consultant, feel free to reach out.

Issue #1458, opened 2024-10-22
Config spontaneous_file_upload.accept is not check in the backend


qvalentin:
**Describe the bug**
Currently the config option spontaneous_file_upload is listed under config options for the backend.
But this config is not used in the server to check files uploaded.
The server should also validate files (e.g. their mimetype) and not rely on the frontend to do it.

This is also relevant to the security issue described here: https://github.com/Chainlit/chainlit/issues/1101#issuecomment-2416021371

**To Reproduce**

1. Set `spontaneous_file_upload.accept` to `["image/png"]`
2. Upload a video with the `/project/file` endpoint
3. The file is uploaded correctly

**Expected behavior**
The server should not accept the file with an mimetype other than the configured values.

**Screenshots**

**Desktop (please complete the following information):**

n/a

**Smartphone (please complete the following information):**

n/a

**Additional context**

Mimetype validation is one aspect of the missing implementation of the config in the backend, the other options should also have effects in the backend.

dokterbob:
@qvalentin Really appreciate your support in dealing with significant technical depth on the security of chainlit (while we're doing by best to prevent this sort of issue in the future).

The relevant config is here: https://github.com/Chainlit/chainlit/blob/6f9d0331df78e8069b0d21d2c27bc2e698556be5/backend/chainlit/config.py#L223

It seems to me that none of these settings are taken into account on the backend. It will take us a couple of days to address that, let us know if you're able and willing to work on a PR ahead of that!

PR #1459, opened 2024-10-22
fix(file_upload): validate mimetype as configured

qvalentin:
https://github.com/Chainlit/chainlit/issues/1458

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -891,10 +891,10 @@ def validate_file_upload(file: UploadFile):
     Raises:
         ValueError: If the file is not allowed.
     """
-    if (
-        config.features.spontaneous_file_upload is None
-        or config.features.spontaneous_file_upload.enabled is False
-    ):
+    if config.features.spontaneous_file_upload is None:
+        """Default to allowing all file types"""
+        return
+    if config.features.spontaneous_file_upload.enabled is False:
         raise ValueError("File upload is not enabled")
 
     validate_file_mime_type(file)
@@ -908,11 +908,10 @@ def validate_file_mime_type(file: UploadFile):
     Raises:
         ValueError: If the file type is not allowed.
     """
-    if config.features.spontaneous_file_upload.accept is None:
+    accept = config.features.spontaneous_file_upload.accept
+    if accept is None:
         "Accept is not configured, allowing all file types"
         return
-
-    accept = config.features.spontaneous_file_upload.accept
 
     assert (
         isinstance(accept, List) or isinstance(accept, dict)
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -872,7 +872,7 @@ async def upload_file(
     assert file.filename, "No filename for uploaded file"
     assert file.content_type, "No content type for uploaded file"
 
-    allowed_file_upload(file)
+    validate_file_upload(file)
 
     file_response = await session.persist_file(
         name=file.filename, content=content, mime=file.content_type
@@ -881,7 +881,7 @@ async def upload_file(
     return JSONResponse(content=file_response)
 
 
-def allowed_file_upload(file: UploadFile):
+def validate_file_upload(file: UploadFile):
     if config.features.spontaneous_file_upload is None:
         return  # TODO: if it is not configured what should happen?
 
@@ -891,11 +891,11 @@ def allowed_file_upload(file: UploadFile):
             detail="File upload is not enabled",
         )
 
-    allowed_file_upload_mime(file)
-    allowed_file_upload_max_size(file)
+    validate_file_mime_type(file)
+    validate_file_size(file)
 
 
-def allowed_file_upload_mime(file: UploadFile):
+def validate_file_mime_type(file: UploadFile):
     if config.features.spontaneous_file_upload.accept is None:
         return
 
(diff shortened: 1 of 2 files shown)


backend/chainlit/server.py
```diff
@@ -870,13 +871,67 @@ async def upload_file(
...
     file_response = await session.persist_file(
         name=file.filename, content=content, mime=file.content_type
     )
 
     return JSONResponse(content=file_response)
 
 
+def validate_file_upload(file: UploadFile):
+    if config.features.spontaneous_file_upload is None:
+        return  # TODO: if it is not configured what should happen?
```
dokterbob, line 885: The most elegant solution is to change the default value for `config.features.spontaneous_file_upload` to always be populated with `SpontaneousFileUploadFeature` then to have `enabled` default to `False` and move  `max_files` and `max_size_mb` [default values from the frontend](https://github.com/Chainlit/chainlit/blob/2ba2c59c241e1227a11a9b439dabba5806e049b5/libs/copilot/src/chat/body.tsx#L41) to the backend config. This would remove the `if ... is None` in favour of more explicit code. But I get it if perhaps that's out of scope and left for a later PR.

The second best would be to follow the [frontend](https://github.com/Chainlit/chainlit/blob/2ba2c59c241e1227a11a9b439dabba5806e049b5/libs/copilot/src/chat/body.tsx#L151) in this and not allow uploads unless explicitly enabled. Definitely, we should try to avoid having a TODO in a PR. ;)

backend/chainlit/server.py
```diff
@@ -870,13 +871,67 @@ async def upload_file(
...
+    validate_file_upload(file)
+
     file_response = await session.persist_file(
         name=file.filename, content=content, mime=file.content_type
     )
 
     return JSONResponse(content=file_response)
 
 
+def validate_file_upload(file: UploadFile):
```
dokterbob, line 883: I would prefer to have validation functions raise a well-defined `Exception` and to have a docstring specifying it's behaviour.

For example, in [pydantic](https://docs.pydantic.dev/latest/concepts/validators/#field-validators) this is a `ValueError` and in [Django](https://docs.djangoproject.com/en/5.1/ref/validators/) it's a `ValidationError`. It should definitely not be a `HTTPException` as the file validation code should not have anything to do with HTTP, it's just validating an `UploadFile`.

We can catch validation errors and then re-raise them in a HTTP request handler (`upload_file`).

backend/chainlit/server.py
```diff
@@ -870,13 +871,67 @@ async def upload_file(
...
+            detail="File upload is not enabled",
+        )
+
+    validate_file_mime_type(file)
+    validate_file_size(file)
+
+
+def validate_file_mime_type(file: UploadFile):
+    if config.features.spontaneous_file_upload.accept is None:
+        return
```
dokterbob, line 899: Maybe add an explicit comment here that we'll allow any mime type unless a value is defined.

backend/chainlit/server.py
```diff
@@ -870,13 +871,67 @@ async def upload_file(
...
+        raise HTTPException(
+            status_code=400,
+            detail="File upload is not enabled",
+        )
+
+    validate_file_mime_type(file)
+    validate_file_size(file)
+
+
+def validate_file_mime_type(file: UploadFile):
```
dokterbob, line 897: Docstring, please! :)

backend/chainlit/server.py
```diff
@@ -870,13 +871,67 @@ async def upload_file(
...
+    if config.features.spontaneous_file_upload.accept is None:
+        return
+
+    accept = config.features.spontaneous_file_upload.accept
+    if isinstance(accept, List):
+        for pattern in accept:
+            if fnmatch.fnmatch(file.content_type, pattern):
+                return
+    else:
```
dokterbob, line 906: It's either a list or a dict. Maybe add an assertion here so that assumption is clear (and the code will bulk if we ever haphazardly decide to change the typing).

backend/tests/test_server.py
```diff
@@ -1,19 +1,24 @@
+import datetime  # Added import for datetime
```
dokterbob, line 1: Maybe it's a good time to remove this redundant comment. ;)

dokterbob (changes requested):
Tests look very clean. Thanks this already great as it is, but I would prefer some slight changes to validation architecture  -- separating HTTP logic from validation logic. Important as we'll add more such functionality going forward, good to have the API clean now.

qvalentin:
> Tests look very clean. Thanks this already great as it is, but I would prefer some slight changes to validation architecture -- separating HTTP logic from validation logic. Important as we'll add more such functionality going forward, good to have the API clean now.

Where would you put the validation logic?
It could be a method of `SpontaneousFileUploadFeature` or just some function somewhere else?

dokterbob:
> > Tests look very clean. Thanks this already great as it is, but I would prefer some slight changes to validation architecture -- separating HTTP logic from validation logic. Important as we'll add more such functionality going forward, good to have the API clean now.
> 
> Where would you put the validation logic? It could be a method of `SpontaneousFileUploadFeature` or just some function somewhere else?

This is what I referred to in that comment: https://github.com/Chainlit/chainlit/pull/1459#discussion_r1830842616

backend/chainlit/server.py
qvalentin: The current behaviour for the e2e tests is that it is None. So rejecting upload when it is not configured will break the tests and might also break things for users.

I agree that the best would be to not have None as the default, but this indeed seems out of scope.

qvalentin:
Hi @dokterbob is anything else missing?

dokterbob (approved):
Looks good to me! Thanks @qvalentin!

PR #1463, opened 2024-10-22
@data_layer API hook to configure data layer.

dokterbob:
`@cl.datalayer` decorator for configuring data layers, rather than forcing users to override `_data_layer`.

This is a first step towards full support of pluggable data layers.

### DIFF
--- a/backend/chainlit/data/__init__.py
+++ b/backend/chainlit/data/__init__.py
@@ -13,7 +13,13 @@ def get_data_layer():
     global _data_layer
 
     if not _data_layer:
-        if api_key := os.environ.get("LITERAL_API_KEY"):
+        from chainlit.config import config
+
+        if config.code.data_layer:
+            # When @data_layer is configured, call it to get data layer.
+            _data_layer = config.code.data_layer()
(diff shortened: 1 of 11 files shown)


backend/chainlit/data/__init__.py
```diff
@@ -11,9 +11,17 @@
 
 def get_data_layer():
     global _data_layer
+    print("Getting data layer", _data_layer)
```
willydouhard, line 14: do we want to keep this print? use logger instead?

willydouhard (commented):
The old way of setting a custom data layer still works right? Since i dont see the e2e test being updated

dokterbob:
> The old way of setting a custom data layer still works right? Since i dont see the e2e test being updated

It technically might, but it was using a variable _intentionally_ marked as private (e.g. prefixed by `_`).

The E2E test _is_ updated: https://github.com/Chainlit/chainlit/pull/1463/files#diff-dd1c256dace91e15fad2a2b0f93a821e2dd256dd48ab991c790b059f9bde2d60

I only want to support this API going forward. The old stuff _might_ work and _might_ keep working but should not support it.

willydouhard (approved):

PR #1468, opened 2024-10-22
Add missing wait in chat_profiles test

jfperusse-bhvr:
This fixes a potential failure in the second chat_profiles test as it is not currently waiting after changing the model to GPT-4.

This test was constantly failing for me on the latest main without the added wait.

### DIFF
--- a/cypress/e2e/chat_profiles/spec.cy.ts
+++ b/cypress/e2e/chat_profiles/spec.cy.ts
@@ -37,7 +37,8 @@ describe('Chat profiles', () => {
     cy.get('[data-test="select-item:GPT-4"]').click();
     cy.get('#confirm').click();
 
-    cy.wait(1000);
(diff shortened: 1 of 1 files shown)


dokterbob:
image

The tests already take a long time, I would prefer if instead of an arbitrary 1s wait you could wait for whatever it is that needs to happen before being able to submit. 

For example, through an alias to intercept HTTP events:
https://docs.cypress.io/api/commands/wait#Alias

Or using a `get()` for whatever is necessary to determine if we're ready to submit.

jfperusse-bhvr:
Updated to wait on the chat-input field to be disabled and enabled again, instead of waiting for 1 second. This seems to work fine locally after testing 10+ times.

dokterbob (approved):
Yay, instead of 1s extra, one less 1s wait in E2E tests! 

Thanks @jfperusse-bhvr!

PR #1473, opened 2024-10-23
Feature - Show images in a LightBox when clicked

buzzCraft:
# Feature - Image LightBox

## Why?
After seeing that this feature was requested back in Nov 2023, and since I couldn't find any implementation of it, I decided to implement it myself.
During my work I discovered https://github.com/Chainlit/chainlit/pull/1402 which is a PR that implements a simple popup for images. While this is an improvement over the current situation, I believe that a lightbox is a better solution for this problem.

## Added dependencies
- [Yet Another React Lightbox](https://yet-another-react-lightbox.com/) - A lightbox component for React.
I chose this libary due to its MIT license and that it is actively maintained.

## Disclaimer
Im not a frontend developer, so I might have missed some best practices or made some mistakes. Please let me know if you see anything that could be improved.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,5 +60,3 @@ dist-ssr
 .coverage
 
 backend/README.md
-
-myenv
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,5 @@ dist-ssr
 .coverage
 
 backend/README.md
+
+myenv
--- a/frontend/src/components/atoms/elements/Image.tsx
+++ b/frontend/src/components/atoms/elements/Image.tsx
@@ -1,8 +1,11 @@
 import { useState } from 'react';
-
 import Skeleton from '@mui/material/Skeleton';
+import Lightbox from 'yet-another-react-lightbox';
+import 'yet-another-react-lightbox/styles.css';
+import Zoom from 'yet-another-react-lightbox/plugins/zoom';
+import Download from 'yet-another-react-lightbox/plugins/download';
 
-import { type IImageElement } from 'client-types/';
+import { type IImageElement, useConfig } from '@chainlit/react-client';
 
 import { FrameElement } from './Frame';
 
@@ -10,67 +13,22 @@ interface Props {
   element: IImageElement;
 }
 
-const handleImageClick = (name: string, src: string) => {
-  const width = window.innerWidth / 2;
-  const height = window.innerHeight / 2;
-  const left = window.innerWidth / 4;
-  const top = window.innerHeight / 4;
-
-  const newWindow = window.open(
-    '',
-    '_blank',
-    `width=${width},height=${height},left=${left},top=${top}`
-  );
-  if (newWindow) {
-    newWindow.document.write(`
-      <html>
-        <head>
-          <title>${name}</title>
-          <link rel="icon" href="/favicon">
-          <style>
-            body {
-              margin: 0;
-              display: flex;
-              flex-direction: column;
-              justify-content: center;
-              align-items: center;
-              height: 100vh;
-              background-color: rgba(0, 0, 0, 0.8);
-            }
-            img {
-              max-width: 100%;
-              max-height: calc(100% - 50px);
-            }
-            a {
-              margin: 10px 0;
-              color: white;
-              text-decoration: none;
-              font-size: 15px;
-              background-color: rgba(255, 255, 255, 0.2);
-              padding: 8px 12px;
-              border-radius: 5px;
-            }
-            a:hover {
-              background-color: rgba(255, 255, 255, 0.4);
-            }
-          </style>
-        </head>
-        <body>
-          <img src="${src}" alt="${name}" />
-          <a href="${src}" download="${name}">Download</a>
-        </body>
-      </html>
-    `);
-    newWindow.document.close();
-  }
-};
-
 const ImageElement = ({ element }: Props) => {
   const [loading, setLoading] = useState(true);
+  const [lightboxOpen, setLightboxOpen] = useState(false);
+  const { config } = useConfig();
 
   if (!element.url) {
     return null;
   }
+
+  const enableLightbox = typedConfig?.features?.image_lightbox && element.display === 'inline';
+
+  const handleImageClick = () => {
+    if (enableLightbox) {
+      setLightboxOpen(true);
+    }
+  };
 
   return (
     <FrameElement>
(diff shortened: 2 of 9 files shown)


buzzCraft:
Simple demo of the feature: 

[chainlit.webm](https://github.com/user-attachments/assets/2ef2b6fc-3694-4999-9ce7-8db6973b9391)

.gitignore
```diff
@@ -60,3 +60,5 @@ dist-ssr
 .coverage
 
 backend/README.md
+
+myenv
```
dokterbob, line 64: This is specific to your dev environment and should not be in this PR!

Protip: update gitignore in your system's git config/home directory!

frontend/src/components/atoms/elements/Image.tsx
```diff
@@ -1,101 +1,85 @@
 import { useState } from 'react';
-
 import Skeleton from '@mui/material/Skeleton';
+import Lightbox from 'yet-another-react-lightbox';
+import 'yet-another-react-lightbox/styles.css';
+import Zoom from 'yet-another-react-lightbox/plugins/zoom';
+import Download from 'yet-another-react-lightbox/plugins/download';
```
dokterbob, line 6: That's a lot of dependencies to import, increasing binary size _even if the feature is not turned on_. How much larger is the output bundle -- and could we somehow make the lightbox optional in the frontend?

Perhaps we can set up stuff so it's a separate chunk, only importing the chunk in the browser if and when the lightbox feature is enabled?

frontend/src/components/atoms/elements/Image.tsx
```diff
@@ -1,101 +1,85 @@
...
 
+  const enableLightbox = config.config?.features.image_lightbox && element.display === 'inline';
+
+  const handleImageClick = () => {
+    if (enableLightbox) {
+      setLightboxOpen(true);
+    }
+  };
+
+
```
dokterbob, line 33: I like how this lightbox code is much more minimal. However, if I read this well, with the lightbox disabled _the old image click behaviour is entirely removed_.

That's not the way forward; developers and users expect what worked before to keep working towards the future and not suddenly disappear. So let's truly make lightbox optional and make sure that we _retain the exact legacy behaviour unless the lightbox is enabled_.

package.json
```diff
@@ -34,4 +34,4 @@
       "micromatch@<4.0.8": ">=4.0.8"
     }
   }
-}
```
dokterbob, line 37: Missing trailing newline. You might want to check your text editor's config!

dokterbob (changes requested):
This is obviously a nice feature and a proper iteration on the last fixup to image display.

But we should make sure to _either_ drop legacy behaviour entirely (and be careful not to increase our bundle size too much) _or_ not change default behaviour and _really_ make lightbox optional. In the latter case, the js code should only be loaded if the lightbox is enabled. My suggestion would be the latter. Regardless, I would like to see the increase in bundle size, please! 

In addition, I would also like to ask for E2E testing both with and without the lightbox enabled. You can prepare the test with it disabled on a fork of `main`, so you can be really sure that your patch does not break pre-existing UX. Here's some suggestions of where you might want to put that: https://github.com/search?q=repo%3AChainlit%2Fchainlit+path%3A%2F%5Ecypress%5C%2Fe2e%5C%2F%2F++image&type=code

Then, add a `lightbox` E2E test specifically testing with the lightbox enabled.

If the original contributor needs support, I imagine there's other community members who'd like to help out getting this big UX improvement in! 

Lastly a question; does this particular lightbox also work for video?

frontend/src/components/atoms/elements/Image.tsx
buzzCraft: I agree with this, but since the original feature was not released at the time of making this, i figured that it could be removed all together. Ill fix that.

frontend/src/components/atoms/elements/Image.tsx
buzzCraft: Ill look into it.

.gitignore
buzzCraft: 😅 Agree, had to swap to my private hardware, and forgot to check that the global git config was set up.

.gitignore
```diff
@@ -59,6 +59,4 @@ dist-ssr
 .aider*
 .coverage
 
-backend/README.md
-
-myenv
+backend/README.md
```
dokterbob, line 62: Trailing newline issue. https://thoughtbot.com/blog/no-newline-at-end-of-file

frontend/src/components/atoms/elements/LightboxWrapper.tsx
```diff
@@ -0,0 +1,52 @@
...
+        download: async ({ slide }) => {
+          try {
+            const response = await fetch(slide.src, { mode: 'cors' });
+            const blob = await response.blob();
+            const url = window.URL.createObjectURL(blob);
+            const link = document.createElement('a');
+            link.href = url;
+            link.download = imageName || 'image';
+            link.click();
+            window.URL.revokeObjectURL(url);
```
dokterbob, line 42: Why do we have a custom `fetch()` here rather than just `download=true`?

Download URL's do not require auth tokens, nor will it in the future (we're using cookie auth instead #1521).

Ref: https://yet-another-react-lightbox.com/plugins/download

dokterbob (commented):
Thanks for the requested changes.

There's still one missing `\n` -- please take care to set up your editor appropriately (this goes for _all_ code projects).

Other than that, there's one question about the download.

I'd also like to invite community members to play with this and give feedback before merging.

.gitignore
buzzCraft: Sorry for dropping the ball on this a bit, but sometimes too much is happening to do community work :( 
All the work has been done on my old private laptop, so yeah, my editor was just set up quick and dirty.. Sorry about that.

Im more than happy to hand this over to someone with more time, as im quite busy two more weeks

frontend/src/components/atoms/elements/LightboxWrapper.tsx
buzzCraft: I had this idea of keeping plugins to a minimum, but with the lazy load you suggested, that might not be needed anymore. 
And, to be honest, this is my first try on TS ever, so iv been confused.

Spent way to long trying to make E2E tests, before I had to put everything on hold.

dokterbob: Pretty good for a first try. :)

I know, building these tests, especially in the beginning, is quite tedious. But once you get the hang of it, additional tests (per project), go faster. Really appreciate your work on this, thanks!

Now on this issue, you're already using the download plugin. ;)

Just putting the URL there might work. Let me know if you're gonna add that to the PR. With that, I'm happy to merge.

buzzCraft: Ill look into it after work tomorrow and report back. Thanks for the guidance and for all the amazing work you and your team does <3

willydouhard:
This is part of the UI revamp in the `2.0.0` release. Feel free to reopen if it is not good enough!

buzzCraft:
Thanks, IRL work has been too much to follow up on this.. Love 2.0.0, now its all about bug hunting and fix everything that broke.

Issue #1472, opened 2024-10-23
Elements don't display when using authentication in 1.3.0 & 2.0.dev1


hayescode:
**Describe the bug**
Since updating version from 2.0.dev0 to 1.3.0 & 2.0.dev1 no elements display when using authentication or authentication+data layer. The elements are successfully persisted, and resuming chats display the elements correctly, but not when the user is actually interacting. They either say "An error has occurred, or loading, or some other error.

I believe this is related to **Authentication** but I have included examples with the data layer too.

Looking at the [diff between these versions](https://github.com/Chainlit/chainlit/compare/2.0.dev0...1.3.0) I cannot figure it out! @dokterbob @willydouhard any ideas?

> [!NOTE]  
> Elements do not work when only using authentication (no data layer)

## Example 1
```py
import chainlit as cl
import pandas as pd
import chainlit.data as cl_data
from chainlit.data.sql_alchemy import SQLAlchemyDataLayer
from chainlit.data.storage_clients.azure import AzureStorageClient

adls = AzureStorageClient(account_url=..., container=..., credential=credential, sas_token=...)
db = cl_data._data_layer = SQLAlchemyDataLayer(conninfo=..., ssl_require=True, storage_provider=adls, user_thread_limit=100, show_logger=True)

@cl.oauth_callback
def login(provider_id: str, token: str, raw_user_data: Dict[str, str], default_user: cl.User, id_token: Optional[str]) -> Optional[cl.PersistedUser]:
    return default_user

@cl.on_chat_start
async def start():
    text_content = "Hello, this is a text element."
    elements = [
        cl.Text(name="simple_text", content=text_content, display="inline")
    ]

    await cl.Message(
        content="Check out this text element!",
        elements=elements,
    ).send()
```
![image](https://github.com/user-attachments/assets/0010911a-bf2f-4f88-ac49-d44be9c06ec9)

## Example 1 (Auth Only)
```py
import chainlit as cl

@cl.oauth_callback
def login(provider_id: str, token: str, raw_user_data: Dict[str, str], default_user: cl.User, id_token: Optional[str]) -> Optional[cl.PersistedUser]:
    return default_user

@cl.on_chat_start
async def start():
    text_content = "Hello, this is a text element."
    elements = [
        cl.Text(name="simple_text", content=text_content, display="inline")
    ]

    await cl.Message(
        content="Check out this text element!",
        elements=elements,
    ).send()
```
[...]

PR #1474, opened 2024-10-23
HOT FIX - Fixes Elements by reverting bug in #1441

hayescode:
The `project/file/{file_id}` endpoint always returns 401 HTTP error. This prevent any Element from being displayed. Fixes https://github.com/Chainlit/chainlit/issues/1472

> [!IMPORTANT]  
> Users should not upgrade to versions 1.3.0 or 2.0.dev1

This comments out the problematic code and adds TODOs. This reopens #1101 and #1438 .

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -881,7 +881,7 @@ async def upload_file(
 async def get_file(
     file_id: str,
     session_id: str,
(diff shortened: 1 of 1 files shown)


dokterbob (approved):
Solving the issue properly requires moving to HTTP only cookies, which is out of scope for now (we want to properly clean up auth). We're gonna ship this ASAP.

Thanks for the contrib!

PR #1488, opened 2024-10-28
Improve UX when loading images

fgalind1:
Backport https://github.com/Chainlit/chainlit/pull/1402 already merged from main/development branch 2.X into releases/1.3

---

Currently when adding images in the conversation, if the image is big and is not that clear, clicking in the image will attempt to download the image.

This PR attempts to improve the UX a bit by just opening the image in a new window (aka popup) to see the image larger. It has a download option on this new popup as well to preserve the original download behavior, but now the UX is a bit better

### DIFF
--- a/frontend/src/components/atoms/elements/Image.tsx
+++ b/frontend/src/components/atoms/elements/Image.tsx
@@ -11,14 +11,58 @@ interface Props {
 }
 
 const handleImageClick = (name: string, src: string) => {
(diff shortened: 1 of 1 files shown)


dokterbob:
Thanks but... We're doing semver now, so the next release adding features would be 1.4 -- but we might just jump right to 2.0 (as we've limited capacity to maintain older code). 1.3 will receive only fixups.

PR #1339, opened 2024-09-13
Enables Copilot to send cookies when calling Chainlit server

mattgartman:
Enables the Copilot functionality to send cookies with requests to the Chainlit server.  This allows support for cookie based sticky sessions.

Since [withCredentials](https://socket.io/docs/v4/client-options/#withcredentials) does not work when a server sets a wildcard CORS origin policy, this update exposes an operational parameter for users needing cookies to be sent with requests.

This PR addresses #1279.

PR #1497, opened 2024-10-31
support socketio auth option in connect handler

nethi:
socketio by default uses ["polling", "websocket"] as transports in that order.  When we have multiple instances of chainlit server with a load balancer in the front, this creates a problem. The service that responds with sessionId is not necessarily the same instance when subsequent  request is made.   While there are other complex strategies such as an external cache (redis etc),  given websocket is supported broadly, a simple strategy is to use just "websocket" transport alone.

However, passing Authorization  header based token mechanism doesn't work with browser based socketio clients, as described [here](https://socket.io/docs/v4/client-options/#extraheaders), and [here](https://stackoverflow.com/a/35381472/635642)

https://github.com/Chainlit/chainlit/blob/67de9c7a1c2f1b663bfec5c28f4b20e180d415ad/libs/react-client/src/useChatSession.ts#L103

```python
      const socket = io(uri, {
        path,
        extraHeaders: {
          Authorization: accessToken || '',
          'X-Chainlit-Client-Type': client.type,
          'X-Chainlit-Session-Id': sessionId,
          'X-Chainlit-Thread-Id': idToResume || '',
          'user-env': JSON.stringify(userEnv),
          'X-Chainlit-Chat-Profile': chatProfile
            ? encodeURIComponent(chatProfile)
            : ''
        }
      });
```
[Recommended](https://stackoverflow.com/a/68062176/635642) [solution with socketio](https://socket.io/docs/v3/middlewares/#sending-credentials) is to use "auth" option

```python
const socket = io(server, {
    transports: ['websocket', 'polling', 'flashsocket'],
    auth: {
        token: accessToken
    }
});
```

However, doing this in a custom client, throws following error in chainlit backend.

```
  File "/test/.pyenv/versions/generic-agent/lib/python3.10/site-packages/engineio/async_server.py", line 494, in run_async_handler
    return await self.handlers[event](*args)
  File "/test/pyenv/versions/generic-agent/lib/python3.10/site-packages/socketio/async_server.py", line 672, in _handle_eio_message
    await self._handle_connect(eio_sid, pkt.namespace, pkt.data)
  File "/test/.pyenv/versions/generic-agent/lib/python3.10/site-packages/socketio/async_server.py", line 551, in _handle_connect
    success = await self._trigger_event(
  File "/test/.pyenv/versions/generic-agent/lib/python3.10/site-packages/socketio/async_server.py", line 637, in _trigger_event
    ret = await handler(*args)
TypeError: connect() takes 2 positional arguments but 3 were given
ERROR:engineio.server:message async handler error
  File "/test/.pyenv/versions/generic-agent/lib/python3.10/site-packages/engineio/async_server.py", line 494, in run_async_handler
    return await self.handlers[event](*args)
  File "/test/.pyenv/versions/generic-agent/lib/python3.10/site-packages/socketio/async_server.py", line 672, in _handle_eio_message
    await self._handle_connect(eio_sid, pkt.namespace, pkt.data)
  File "/test/.pyenv/versions/generic-agent/lib/python3.10/site-packages/socketio/async_server.py", line 551, in _handle_connect
    success = await self._trigger_event(
  File "/test/.pyenv/versions/generic-agent/lib/python3.10/site-packages/socketio/async_server.py", line 637, in _trigger_event
    ret = await handler(*args)
TypeError: connect() takes 2 positional arguments but 3 were given

```

Basically, if auth option is enabled in client, [python_socketio](https://python-socketio.readthedocs.io/en/latest/server.html#connect-and-disconnect-events) calls "connect" callback with a third argument "auth". But chainlit callback handler doesn't define a third argument.  

https://github.com/Chainlit/chainlit/blob/67de9c7a1c2f1b663bfec5c28f4b20e180d415ad/backend/chainlit/socket.py#L102

```python
@sio.on("connect")
async def connect(sid, environ):
    if (
        not config.code.on_chat_start
        and not config.code.on_message
        and not config.code.on_audio_chunk
    ):
        logger.warning(
            "You need to configure at least one of on_chat_start, on_message or on_audio_chunk callback"
        )
        return False
    user = None
    token = None
    login_required = require_login()
```

A simple solution is to add the third argument, auth, with default value of None and use auth["token"] if HTTP_AUTHORIZTION is not present.  This is exactly done by this PR and works great.

### DIFF
--- a/libs/react-client/README.md
+++ b/libs/react-client/README.md
@@ -42,8 +42,8 @@ This hook is responsible for managing the chat session's connection to the WebSo
 
 #### Methods
 
-- `connect`: Establishes a connection to the SocketIO server.
-- `disconnect`: Disconnects from the SocketIO server.
+- `connect`: Establishes a connection to the WebSocket server.
+- `disconnect`: Disconnects from the WebSocket server.
 - `setChatProfile`: Sets the chat profile state.
 
 #### Example
@@ -60,8 +60,7 @@ const ChatComponent = () => {
       userEnv: {
         /* user environment variables */
       },
-      accessToken: 'Bearer YOUR_ACCESS_TOKEN', // Optional Chainlit auth token
-      requireWebSocket: true // Optional, require WebSocket upgrade to be successful before user can interact with the chat bot. Will retry upgrade request every 500ms until successful. Please note if your server is behind a proxy, you will have to configure it to accept websocket upgrade request, otherwise users won't be able to interact with the app. You can check an example using nginx proxy here: https://nginx.org/en/docs/http/websocket.html. Default to false.
+      accessToken: 'Bearer YOUR_ACCESS_TOKEN' // Optional Chainlit auth token
     });
 
     return () => {
--- a/libs/react-client/src/useChatSession.ts
+++ b/libs/react-client/src/useChatSession.ts
@@ -86,12 +86,10 @@ const useChatSession = () => {
   const _connect = useCallback(
     ({
       userEnv,
-      accessToken,
-      requireWebSocket = false
+      accessToken
     }: {
       userEnv: Record<string, string>;
       accessToken?: string;
-      requireWebSocket?: boolean;
     }) => {
       const { protocol, host, pathname } = new URL(client.httpEndpoint);
       const uri = `${protocol}//${host}`;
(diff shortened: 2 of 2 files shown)


nethi:
Here is a test html I used to test with a stand-alone browser based client.
[index.html.zip](https://github.com/user-attachments/files/17588575/index.html.zip)

dokterbob:
Thanks for the contrib @nethi. We're looking into adding HTTP only cookie support, and making that the default auth for the frontend. That would significantly improve security and potentially resolve other issues as well.

Might it work for your use case, too?

dokterbob:
I don't know about socketio specifically, but in general on the FastAPI side it seems cookie-based websocket auth is a pretty standard thing: https://fastapi.tiangolo.com/advanced/websockets/#using-depends-and-others

AFAIK WebSockets start out with a normal HTTP request. 

The advantage of HTTP only cookies is that JS never has access to the auth token, so XSS attacks which are normally a pain to deal with disappear entirely (though CSRF attacks require proper config server-side).

sandangel:
Is it possible to use Server Send Event and remove socket.io? So we can have simpler architecture and more stable, scalable with stateless servers.

dokterbob:
> chainlit/libs/react-client/src/useChatSession.ts

No, we're not gonna do that. SSE's are, like you said, server-sent events. Websockets are bidirectional. The connection persistence are what allows a smooth UX, the socketio bus is basically the core of our functionality.

dokterbob:
@nethi I'm looking to migrate to httponly based cookies as the default frontend -> backend token bearer over the next 2 weeks or so. This removes XSS attack vectors and will allow us to properly secure files, which are currently served, effectively, unauthenticated.

That having been said, your patch looks like a good solution and perhaps we would like to have it in as a fixup (also for the 1.3 branch, which might not get the cookie-based auth). 

In the description, you're suggesting patching `useChatSession`. I know for a fact there's others in your situation, so I think this change should be part of this fix. That will also enable E2E tests for it.

So, could you please add your suggested changes to useChatSession to the PR? :)

dokterbob:
@nethi How about https://github.com/Chainlit/chainlit/pull/1339?

nethi:
>>Any idea why?

I can create a PR if there is interest.

PR #1508, opened 2024-11-08
Bump version to 2.1, changelog for unreleased @data_layer

dokterbob:


dokterbob:
I realised, this doesn't need to be 3.0, 2.1 will do (we're not breaking functionality, at this point).

willydouhard (approved):

PR #1529, opened 2024-11-18
feat: use async lc tracer instead of run_sync

willydouhard:
1. Remove use_sync method in Langchain integration
2. Support Pydantic outputs with Langchain/Langgraph


dokterbob (approved):
Removed formatting commits, added fallback for pydantic v1 and tested on our production app for basic functionality.

Issue #1556, opened 2024-12-01
Compatibility issue with Chainlit 1.3.2 and Pydantic 2.10.2


Hamza-cpp:
# Compatibility issue with Chainlit 1.3.2 and Pydantic 2.10.2  

## Description  
I'm encountering a `PydanticUserError` when attempting to run the `chainlit` command in my project using Chainlit 1.3.2 and Pydantic 2.10.2. The error suggests that `CodeSettings` is not fully defined and requires a rebuild of the dataclass. This issue occurs immediately upon invoking the `chainlit` CLI.  

---

## Steps to Reproduce  
1. Set up a Python environment (e.g., with `venv`).  
2. Install the following dependencies:
   - Chainlit: `1.3.2`
   - Pydantic: `2.10.2`
3. Run the `chainlit hello` command.  

---

## Expected Behavior  
The `chainlit hello` command should execute without errors, launching the Chainlit application as expected.  

---

## Observed Behavior  
The following traceback is encountered:  
```bash
(.venv) youssef@hp ~/Hamza_OK/Chef_Beldi  $ chainlit hello
Traceback (most recent call last):
  File "/home/youssef/Hamza_OK/Chef_Beldi/.venv/bin/chainlit", line 5, in <module>
    from chainlit.cli import cli
  File "/home/youssef/Hamza_OK/Chef_Beldi/.venv/lib/python3.12/site-packages/chainlit/__init__.py", line 18, in <module>
    from chainlit.action import Action
  File "/home/youssef/Hamza_OK/Chef_Beldi/.venv/lib/python3.12/site-packages/chainlit/action.py", line 5, in <module>
    from chainlit.telemetry import trace_event
  File "/home/youssef/Hamza_OK/Chef_Beldi/.venv/lib/python3.12/site-packages/chainlit/telemetry.py", line 12, in <module>
    from chainlit.config import config
  File "/home/youssef/Hamza_OK/Chef_Beldi/.venv/lib/python3.12/site-packages/chainlit/config.py", line 538, in <module>
    config = load_config()
             ^^^^^^^^^^^^^
  File "/home/youssef/Hamza_OK/Chef_Beldi/.venv/lib/python3.12/site-packages/chainlit/config.py", line 507, in load_config
    settings = load_settings()
               ^^^^^^^^^^^^^^^
  File "/home/youssef/Hamza_OK/Chef_Beldi/.venv/lib/python3.12/site-packages/chainlit/config.py", line 479, in load_settings
[...]

Issue #1544, opened 2024-11-26
Update of pydantic to version 2.10.2 breaks chainlit


nkratzke:
**Describe the bug**
The recent update of pydantic to version 2.10.2 breaks chainlit

**To Reproduce**

1. Install chainlit using `pip install chainlit`
2. Run chainlit using `chainlit`

The following error occurs:

```
Traceback (most recent call last):
  File "/Users/nane/test/chainlit/.venv/bin/chainlit", line 5, in <module>
    from chainlit.cli import cli
  File "/Users/nane/test/chainlit/.venv/lib/python3.12/site-packages/chainlit/__init__.py", line 18, in <module>
    from chainlit.action import Action
  File "/Users/nane/test/chainlit/.venv/lib/python3.12/site-packages/chainlit/action.py", line 5, in <module>
    from chainlit.telemetry import trace_event
  File "/Users/nane/test/chainlit/.venv/lib/python3.12/site-packages/chainlit/telemetry.py", line 12, in <module>
    from chainlit.config import config
  File "/Users/nane/test/chainlit/.venv/lib/python3.12/site-packages/chainlit/config.py", line 538, in <module>
    config = load_config()
             ^^^^^^^^^^^^^
  File "/Users/nane/test/chainlit/.venv/lib/python3.12/site-packages/chainlit/config.py", line 507, in load_config
    settings = load_settings()
               ^^^^^^^^^^^^^^^
  File "/Users/nane/test/chainlit/.venv/lib/python3.12/site-packages/chainlit/config.py", line 479, in load_settings
    code_settings = CodeSettings(action_callbacks={})
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/nane/test/chainlit/.venv/lib/python3.12/site-packages/pydantic/_internal/_dataclasses.py", line 121, in __init__
    s.__pydantic_validator__.validate_python(ArgsKwargs(args, kwargs), self_instance=s)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/nane/test/chainlit/.venv/lib/python3.12/site-packages/pydantic/_internal/_mock_val_ser.py", line 100, in __getattr__
    raise PydanticUserError(self._error_message, code=self._code)
pydantic.errors.PydanticUserError: `CodeSettings` is not fully defined; you should define `Action`, then call `pydantic.dataclasses.rebuild_dataclass(CodeSettings)`.

[...]

Issue #1543, opened 2024-11-26
Hello app does not work


BillShiyaoZhang:
**Describe the bug**
The hello app does not work.
```
>> $ chainlit hello
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.13/bin/chainlit", line 5, in <module>
    from chainlit.cli import cli
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/chainlit/__init__.py", line 18, in <module>
    from chainlit.action import Action
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/chainlit/action.py", line 5, in <module>
    from chainlit.telemetry import trace_event
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/chainlit/telemetry.py", line 12, in <module>
    from chainlit.config import config
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/chainlit/config.py", line 538, in <module>
    config = load_config()
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/chainlit/config.py", line 507, in load_config
    settings = load_settings()
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/chainlit/config.py", line 479, in load_settings
    code_settings = CodeSettings(action_callbacks={})
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/pydantic/_internal/_dataclasses.py", line 121, in __init__
    s.__pydantic_validator__.validate_python(ArgsKwargs(args, kwargs), self_instance=s)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/pydantic/_internal/_mock_val_ser.py", line 100, in __getattr__
    raise PydanticUserError(self._error_message, code=self._code)
pydantic.errors.PydanticUserError: `CodeSettings` is not fully defined; you should define `Action`, then call `pydantic.dataclasses.rebuild_dataclass(CodeSettings)`.

For further information visit https://errors.pydantic.dev/2.10/u/class-not-fully-defined
```

**To Reproduce**

1. pip install chainlit
2. chainlit hello
3. See error

**Expected behavior**
[...]

PR #1562, opened 2024-12-03
Exclude pydantic 2.10.2

dokterbob:
Addresses #1543, #1544, #1556.

Ref: https://github.com/pydantic/pydantic/issues/11031


dokterbob:
The CI installs from poetry lock file (for reproducibility and to avoid resolving deps all the time), which uses pydantic 1. We might consider testing both 1 and 2, although we can then start doing that with a frightening about of deps!

Trying install without lock file now, but it's taking forever to lock dependencies...

Viicos:
Hi @dokterbob, Pydantic maintainer here.

A couple things to note:
- Pydantic V2 (released a year and a half ago) is changing a lot of things from V1. Depending on your usage of Pydantic, it might be tricky to support both. But looking at your code base, you are mainly defining dataclasses/models and did not make use of validators, so seems like it should be alright.
- In 2.10, we did a complete refactor of the forward annotations evaluation logic. I'm currently investigating your use case, and I'll get back to you once this done. However, do note that https://github.com/pydantic/pydantic/issues/11031 is completely unrelated from the issues you are seeing.

dokterbob:
Thank you so much @Viicos, for this excellent contribution.

I only afforded a quick look and it seemed that the similar error message together with the recent release suggested a bug on your end. Appreciate your clarification and super appreciate your fix!

> * Try to fix the circular imports. In general, this isn't an easy task, and looking at your code architecture, it doesn't seem trivial.

:')

Every pass I make I refactor a bit -- eventually we should not have to avoid type checking _anywhere_. But there's a long way to go. ;)

Issue #1561, opened 2024-12-02
Bump Python Multipart dependency


tituslhy:
**Is your feature request related to a problem? Please describe.**
The Chainlit python-multipart dependency currently has a conflict with the [llama-deploy](https://github.com/run-llama/llama_deploy) which is the deployment library for LlamaIndex's workflow object. 

The llama-deploy library handles everything in the backend - control plane (to orchestrate between workflows) deployment, workflow deployments and message queue deployment to respective FastAPI endpoints - and is useful for scaling the backend of large scale multi-agent systems. These results can be rendered on the Chainlit frontend through a higher level abstraction without the need for using Python's `requests` library. 

**Describe alternatives you've considered**
The llama-deploy library was using python-multipart=0.0.10 and I lowered it to be >=0.0.9. This was approved. However, subsequent improvements made to the llama-deploy library required for the python-multipart version to be bumped to 0.0.18 and this will be taking place very soon.

**Desired solution**
Bump the Chainlit python-multipart library dependency from "^0.0.9" to ">=0.0.9".

dokterbob:
@tituslhy I am happy to accept a PR for `^0.0`, to ensure that we can actually support it.

tituslhy:
Thanks @dokterbob! I updated the dependency and checked in the files - see [pull request #1563](https://github.com/Chainlit/chainlit/pull/1563)

willydouhard:
Done

PR #1563, opened 2024-12-03
bump python multipart to 0.0.18

tituslhy:
Bumping multipart dependency to resolve [feature request #1561](https://github.com/Chainlit/chainlit/issues/1561)

### DIFF
--- a/backend/poetry.lock
+++ b/backend/poetry.lock
@@ -13,123 +13,109 @@ files = [
 
 [[package]]
 name = "aiohappyeyeballs"
-version = "2.4.0"
+version = "2.4.4"
 description = "Happy Eyeballs for asyncio"
 optional = false
 python-versions = ">=3.8"
 files = [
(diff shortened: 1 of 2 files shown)


dokterbob:
Why this consistently fails is beyond me...

dokterbob (commented):
Thanks for the contrib but... this updates a lot more than just python-multipart!

Potentially causing a lot of side-effects (maybe that's why tests are failing). A patch updating python-multipart should really just be updating that.

In the end I did `poetry add 'python-multipart>=0.0.18,<0.1'`, to keep with my suggested `^0.0` but also require `>0.0.18` as there's a security issue (https://github.com/Chainlit/chainlit/security/dependabot/99).

To save both of us a bit of time, I already created the PR. https://github.com/Chainlit/chainlit/pull/1570

Love to hear your feedback though, does it solve your issue?

dokterbob:
Superseded by #1570

tituslhy:
Yes it does! The failing tests are weird. I only updated the pyproject.toml file and later just ran `poetry lock` and checked in both the pyproject.toml and poetry.lock file. I think the locking process caused Python to automatically update the rest of the dependencies which caused the tests to fail.

Thanks @dokterbob!

PR #1568, opened 2024-12-04
Import `Field` from `pydantic` instead of `pydantic.dataclasses`

Viicos:
`Field` just so happens to be imported internally by the `pydantic.dataclasses` module, but this can change at any time.


dokterbob (approved):
Super, thanks!

PR #1584, opened 2024-12-11
Remove numpy depdendency

dokterbob:
I can't find where we're explicitly installing numpy, so propose to remove it as a (bulky!) dependency.


willydouhard:
Duplicate of https://github.com/Chainlit/chainlit/pull/1633

PR #1585, opened 2024-12-11
Remove unused asyncpg dependency

dokterbob:
Doesn't seem to get used anywhere.


desaxce:
Now it's used by ChainlitDataLayer :)

PR #1594, opened 2024-12-19
Preserve root path when mounting as FastAPI sub-application

AnysSido:
Currently, when Chainlit is mounted as a FastAPI sub-application, the `mount_chainlit` function overrides the root path with the sub-application path. This prevents Chainlit from correctly serving static files since the full path is not preserved.

This change ensures that when a root path is provided, it is preserved and combined with the sub-application path during mounting.


willydouhard:
I think I fixed it in the `2.0.0` release. Can you confirm?

dokterbob:
> I think I fixed it in the `2.0.0` release. Can you confirm?

@AnysSido Could you please let us know whether this is already fixed in 2.0?

AnysSido:
> > I think I fixed it in the `2.0.0` release. Can you confirm?
> 
> @AnysSido Could you please let us know whether this is already fixed in 2.0?

I haven't had chance to migrate to 2.0.0 yet, I will try to find some time tomorrow.

AnysSido:
@dokterbob @willydouhard I have just migrated to Chainlit 2.0.2 and the issue is not fixed. I can also confirm that this MR does still fix the issue.

AnysSido:
> @AnysSido Thanks for the quick feedback.
> 
> Could you share a minimal FastAPI/chainlit app so we can replicate this problem -- and verify your solution?

I could but it would be just the most basic FastAPI application with Chainlit mounted as a sub-application using Chainlit's `mount_chainlit` function.

If you serve the FastAPI application behind a reverse proxy, say at `/myservice`, and you mount the chainlit sub-app at `/chat`, then chainlit will be served from `/myservice/chat`.

In the function modified in this PR you can see that the root path is originally discarded and replaced with the sub-app path. This means chainlit will try to serve static files from `/chat` instead of `/myservice/chat`.

This PR ensures that the root path is preserved when both are used.

dokterbob:
> > @AnysSido Thanks for the quick feedback.
> > Could you share a minimal FastAPI/chainlit app so we can replicate this problem -- and verify your solution?
> 
> I could but it would be just the most basic FastAPI application with Chainlit mounted as a sub-application using Chainlit's `mount_chainlit` function.
> 
> If you serve the FastAPI application behind a reverse proxy, say at `/myservice`, and you mount the chainlit sub-app at `/chat`, then chainlit will be served from `/myservice/chat`.
> 
> In the function modified in this PR you can see that the root path is originally discarded and replaced with the sub-app path. This means chainlit will try to serve static files from `/chat` instead of `/myservice/chat`.
> 
> This PR ensures that the root path is preserved when both are used.

Thanks but there's more ways you could do all that than you or I could possibly fathom. ;)

Having specific examples and replication cases, code that runs or a Docker container or whatever makes it easier for us to evaluate PR's really helps us accept contributions faster.

In the mean time, another user contributed https://github.com/Chainlit/chainlit/pull/1723, which seems to address similar issues, but also cleans up a lot of other stuff.

Would you be able to confirm whether it does indeed address your issue? As it cleans up a lot more, I would prefer to mention the other PR (but happy to mention/credit you in the commit msg).

dokterbob:
Closing this in favour of #1723.

PR #1609, opened 2024-12-25
feat: source maps

asvishnyakov:
This PR adds source map support to the Chainlit frontend and the react-client library.

Source maps make debugging easier, particularly in production environments. They allow you to debug pre-compiled code without needing to deploy a debug version or run Chainlit from the source code.

This PR enables source map generation as separate files, ensuring it does not affect the resulting bundle, package, or wheel size, except for a single line added at the end of each file to reference the corresponding source map file.

You can learn more about source maps [here](https://web.dev/articles/source-maps) and the settings used [here](https://v3.vitejs.dev/config/build-options.html#build-sourcemap) and [here](https://www.typescriptlang.org/tsconfig/#sourceMap).

After this PR, the build system may generate numerous warnings about missing source maps and incorrectly (though this doesn't seem to have much impact) add the source map link to files twice, with the wrong file name the first time (`out.js.map` instead of `index.js.map`).

The first issue can be resolved by upgrading to `vite-plugin-react-swc` version 3.7.1 or later, and the second by upgrading `tsup` to version 8.1.2 or later.

Unfortunately, I wasn't able to get source maps working for `libs/*` when they are included in the `frontend` app compiled in production mode. This seems to be an issue with Vite/React/SWC, as it affects all dependencies, not just `libs/*`. However, source maps for `libs/*` are generated correctly and can be used in other apps that depend on these libraries.

Feel free to edit this PR if needed as I allowed edits by maintainers.


willydouhard:
Thank you for the contribution! Is this impacting the bundle size?

asvishnyakov:
@willydouhard It depends on what you mean by bundle size. The bundle itself (i.e., the size of `index.js`) will remain the same, and end users will not notice any change in the overall size of loaded files in the browser. However, the zip file (`npm` or `pip` wheel) containing it will include an additional large file, and developers who open the browser's developer console will load this additional data.

dokterbob (approved):
Great, thanks!

PR #1610, opened 2024-12-25
fix: auto load threads after migration to cookie

asvishnyakov:
"It seems that migrating to cookie authentication broke automatic thread loading because `accessToken` may no longer be available. This PR restores the functionality by using `isAuthenticated` from `useAuth` instead.


willydouhard:
This has been fixed in the `2.0.0` release! Thank you for contributing :)

PR #1615, opened 2024-12-30
Willy/ui2

willydouhard:


backend/chainlit/element.py
```diff
@@ -388,14 +390,15 @@ def __post_init__(self) -> None:
...
+    """Useful to send a custom element to the UI."""
 
-    type: ClassVar[ElementType] = "component"
+    type: ClassVar[ElementType] = "custom"
     mime: str = "application/json"
     props: Dict = Field(default_factory=dict)
 
     def __post_init__(self) -> None:
         self.content = json.dumps(self.props)
+        self.updatable = True
```
Atry, line 402: Why is `updatable` hard-coded as `True`? Is it a bug?

PR #1617, opened 2024-12-30
fix upload button render logic

### LINKED ISSUE #1548: AskUploadButton renders as "0" due to isAsk sometimes being zero
**Describe the bug**

With chat history using SQLAlchemy data layer with SQLite, the UI becomes like this (see the zeros):

Näyttökuva 2024-11-28 kello 10 28 07

This is because Message.tsx renders following into zero when isAsk is zero and isRunning is falsy:

```
{!isRunning && isAsk && (
<AskUploadButton onError={onError} />
)}
```     

That zero value originates from SQLAlchemy layer implementation. Perhaps there should be casting to boolean there.

In my humble opinion, the above UI logic is not optimal either, as evidenced by this issue. There's nothing wrong with preferring readability and simplicity over cleverness and reliance on what could be here considered a side effect.
            
**To Reproduce**

Implement chat history with SQLAlchemy & SQLite.

petri:
Fix #1548

### DIFF
--- a/frontend/src/components/molecules/messages/Message.tsx
+++ b/frontend/src/components/molecules/messages/Message.tsx
@@ -185,9 +185,9 @@ const Message = memo(
                         allowHtml={allowHtml}
                         latex={latex}
                       />
-                      {!isRunning && isAsk && (
+                      {!isRunning && isAsk ? (
                         <AskUploadButton onError={onError} />
-                      )}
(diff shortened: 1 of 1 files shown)


willydouhard:
Thank you for contributing! I think I fixed it in the Chainlit `2.0.0` UI revamp

PR #1618, opened 2024-12-31
Update message.py

cta2106:
`step_dict` has a `content` key, not an `output` key


willydouhard:
A message is a variant of a step. For simplicity reasons, they are stored int the same table. The convention is that a message has a content instead of input/output but in the table it is stored in the output column.

willydouhard:
Correct!

PR #1720, opened 2025-01-16
add json file to translate into japanese

myksyut:
just add json file


willydouhard (approved):

willydouhard:

Issue #1629, opened 2025-01-06
Please add timezone support and config.


hexart:
`from datetime import datetime
import pytz
from src.app.core.config import settings
timezone = pytz.timezone(settings.timezone)
def utc_now():
    now = datetime.now(timezone)
    return now`
put timezone config in .env and config.toml:
`TIMEZONE=Asia/Shanghai`

hexart:
<img width="545" alt="截屏2025-01-09 18 28 32" src="https://github.com/user-attachments/assets/a12690b4-dd6f-4cb7-96e8-da0fc7409cb6" />
if there's no timezone configuration, the newest thread was in Last 7 days, and the createAt shows like this:
<img width="383" alt="截屏2025-01-09 18 29 03" src="https://github.com/user-attachments/assets/a01645cd-b495-4d69-9e5e-d8cd8b50494d" />

PR #1732, opened 2025-01-17
fix: optimize thread date grouping logic

hexart:
## Problem
The current implementation has potential timezone issues and inconsistent grouping due to:
1. Date Boundary Issues
   - Using `toDateString()` for comparison makes grouping sensitive to specific time
   - Example: A thread created at "2024-01-17 23:59 UTC" might show in different groups for users in UTC+8 vs UTC-8
2. Overlapping Group Logic
   - Direct `>=` comparisons with `sevenDaysAgo` and `thirtyDaysAgo` can cause threads to appear in multiple groups
   - Example: A thread from yesterday might also match `>= sevenDaysAgo` condition
3. Time Sorting
   - Lacks time-based sorting within groups
   - Leads to inconsistent thread ordering in UI

### 1. **Normalized Date Comparison**
```typescript
// Before
const isToday = createdAt.toDateString() === today.toDateString();

// After
today.setHours(0, 0, 0, 0);
threadDate.setHours(0, 0, 0, 0);
const daysDiff = Math.floor((today.getTime() - threadDate.getTime()) / 86400000);
```
- Normalizes dates to start of day in user's timezone
- Ensures consistent grouping regardless of the specific time

### 2. **Clear Group Boundaries**
```typescript
// Before
const isLast7Days = createdAt >= sevenDaysAgo;  // Can overlap with Today/Yesterday

// After
if (daysDiff === 0) {...}        // Today only
else if (daysDiff === 1) {...}   // Yesterday only
else if (daysDiff <= 7) {...}    // 2-7 days ago
```

### 3. **Time-based Sorting**
```typescript
[...data].sort((a, b) => 
  new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime()
)
```
- Ensures consistent thread ordering within groups
- Improves user experience with predictable sorting

## Benefits
1. **Timezone Consistency**
  - Threads appear in the same groups regardless of user timezone
  - Date boundaries respect local time

2. **Accurate Grouping**
  - No more overlapping between groups
  - Clear and logical group boundaries

3. **Better Performance**
  - Reduces Date object creation
  - Simpler and more efficient comparison logic

4. **Improved UX**
  - Consistent thread ordering
  - Predictable group assignments

## Testing
Tested with:
- Different user timezones (UTC+8, UTC-8, UTC)
- Date boundary cases (23:59, 00:01)
- Various creation dates spanning multiple groups

### Related Issue
#1629

### DIFF
--- a/libs/react-client/src/utils/group.ts
+++ b/libs/react-client/src/utils/group.ts
@@ -2,45 +2,35 @@ import { IThread } from 'src/types';
 
 export const groupByDate = (data: IThread[]) => {
   const groupedData: { [key: string]: IThread[] } = {};
-
+  
   const today = new Date();
-  const yesterday = new Date();
-  yesterday.setDate(today.getDate() - 1);
-  const sevenDaysAgo = new Date();
-  sevenDaysAgo.setDate(today.getDate() - 7);
-  const thirtyDaysAgo = new Date();
-  thirtyDaysAgo.setDate(today.getDate() - 30);
-
-  data.forEach((item) => {
-    const createdAt = new Date(item.createdAt);
-    const isToday = createdAt.toDateString() === today.toDateString();
-    const isYesterday = createdAt.toDateString() === yesterday.toDateString();
-    const isLast7Days = createdAt >= sevenDaysAgo;
-    const isLast30Days = createdAt >= thirtyDaysAgo;
-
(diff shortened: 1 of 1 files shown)


willydouhard (approved):
LGTM

PR #1729, opened 2025-01-17
Add i18n Support for Chat Components

hexart:
## Changes
This PR adds internationalization support for several chat-related components:
- Related issue:
- #1710
- #1681
- #1640
- #1551
- #1550  

### New Chat Dialog
- Added translation paths in NewChat.tsx
- Updated corresponding strings in en-US.json under newChatDialog

### Thread List Component
- Added translation paths for renaming dialog and deletion dialog strings in ThreadList.tsx
- Added translation paths for time group labels (Today, Yesterday, etc.)
- Updated corresponding strings in en-US.json under threadList section
- Fixed e2e test selector to work with translated dialog buttons
   - Changed button selector to use [role='alertdialog'] button.bg-primary instead of text content

### Ask File Button Component
- Added translation paths in AskFileButton.tsx
- Updated corresponding strings in en-US.json under askFileButton section

### Implementation Details
- All translations follow existing path patterns
- Organized strings hierarchically based on component structure
- Maintained consistent format between component files and translation JSON
- Used Translator component for all hardcoded strings
- Added missing Translator imports where needed
- Used CSS classes/attributes for test selectors instead of text content

### Notes
- No functional changes, purely i18n support additions
- Maintained existing UI/UX while adding translation capability
- Updated test selectors to be more resilient with translated content

The main challenge was updating the e2e tests to work with translated content. Instead of relying on button text content, we now use role attributes and CSS classes to select dialog buttons reliably.

PR #1740, opened 2025-01-17
Fix Broken Translation Paths in Left Sidebar in 2.0.4

AidanShipperley:
The translation additions in #1729 broke text from appearing for some interactions in the left sidebar due to the incorrect path being used in [ThreadList.tsx](https://github.com/Chainlit/chainlit/blob/32d7fd821fdb74a8be74fda46aeb75cdc410d921/frontend/src/components/LeftSidebar/ThreadList.tsx), so this fixes the paths to match the [en-US translation file](https://github.com/Chainlit/chainlit/blob/32d7fd821fdb74a8be74fda46aeb75cdc410d921/backend/chainlit/translations/en-US.json).

Currently in Chainlit 2.0.4, some panels have dialogue missing:

#### Renaming and deleting a thread before the fix:
image
image

#### Renaming and deleting a thread after this fix:
image
image

### DIFF
--- a/frontend/src/components/LeftSidebar/ThreadList.tsx
+++ b/frontend/src/components/LeftSidebar/ThreadList.tsx
@@ -92,7 +92,7 @@ export function ThreadList({
   if (!threadHistory || size(threadHistory?.timeGroupedThreads) === 0) {
     return (
       <Alert variant="info" className="m-3">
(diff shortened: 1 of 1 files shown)


AidanShipperley:
Saw that a fix was already submitted by in #1738, so I'm closing.

PR #1743, opened 2025-01-18
fix: add partitioned cookie to avoid browser error

sandangel:
Use partitioned cookie as discussed in https://discord.com/channels/1088038867602526210/1328452879870197851


willydouhard:
I think this is a duplicate of https://github.com/Chainlit/chainlit/pull/1758. If that's the case I'll close this PR. Thank you for your contribution nonetheless!

PR #1748, opened 2025-01-19
Set uvicorn log level to warning to minimize non-essential logs

hexart:
If this is not appropriate, please feel free to close this PR.


dokterbob:
We definitely want max. debug output in debug mode (for now).

Long term, we should look into having proper/full Python log config support.

Thanks for understanding!

Issue #1728, opened 2025-01-16
Fix for Installing Chainlit on ARM Macs


ishaansehgal99:
**Describe the bug**
Bug arises on M3 Pro max mac - during installation of Chainlit project following first step of `CONTRIBUTING.md`

```
poetry install --with tests --with mypy --with dev
Installing dependencies from lock file

No dependencies to install or update

Installing the current project: chainlit (2.0.3)Preparing build environment with build-system requirements poetry-core==1.9.1

-- Building frontend, this might take a while!

   If you don't need to build the frontend and just want dependencies installed, use:
   `poetry install --no-root`

-- Running: /Users/ishaansehgal/Library/pnpm/pnpm install --frozen-lockfile
Scope: all 4 workspace projects
Already up to date
Done in 60ms
-- Running: /Users/ishaansehgal/Library/pnpm/pnpm buildUi

> @ buildUi /Users/ishaansehgal/Documents/chainlit
> cd libs/react-client && pnpm run build && cd ../copilot && pnpm run build && cd ../../frontend && pnpm run build


> @chainlit/react-client@0.2.3 build /Users/ishaansehgal/Documents/chainlit/libs/react-client
> tsup src/index.ts --clean --format esm,cjs --dts  --external react --external recoil --minify --sourcemap --treeshake

Error: Cannot find module @rollup/rollup-darwin-x64. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
    at requireWithFriendlyError (/Users/ishaansehgal/Documents/chainlit/libs/react-client/node_modules/.pnpm/rollup@4.24.0/node_modules/rollup/dist/native.js:59:9)
    at Object.<anonymous> (/Users/ishaansehgal/Documents/chainlit/libs/react-client/node_modules/.pnpm/rollup@4.24.0/node_modules/rollup/dist/native.js:68:76)
    at Module._compile (node:internal/modules/cjs/loader:1364:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1422:10)
    at Module.load (node:internal/modules/cjs/loader:1203:32)
    at Module._load (node:internal/modules/cjs/loader:1019:12)
    at Module.require (node:internal/modules/cjs/loader:1231:19)
    at require (node:internal/modules/helpers:177:18)
[...]

PR #1752, opened 2025-01-20
fix: Update package.json for ARM Mac

ishaansehgal99:
https://github.com/Chainlit/chainlit/issues/1728

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "devDependencies": {
+    "@rollup/rollup-darwin-arm64": "^4.30.1",
     "@trivago/prettier-plugin-sort-imports": "^4.3.0",
     "@types/react": "^18.2.15",
     "@types/react-dom": "^18.2.7",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "devDependencies": {
+    "@rollup/rollup-darwin-arm64": "^4.30.1",
```
dokterbob, line 3: We don't want to have platform-dependent stuff here.

package.json
```diff
@@ -12,6 +13,7 @@
     "kill-port": "^2.0.1",
     "lint-staged": "^13.3.0",
     "prettier": "^2.8.8",
+    "rollup": "3.29.4",
```
dokterbob, line 16: This is an indirect dependency and AFAIK we don't need to have it explicitly mentioned here.

dokterbob (changes requested):
Thanks for the contribution!

However, updating indirect dependencies with pnpm is simply a matter of updating either the direct dependencies, which then update indirect dependencies _or_ _just_ updating the lockfile: https://pnpm.io/cli/update

Do please try again (e.g. `git reset --hard main`) -- as I don't have an ARM platform available for testing.

PR #1757, opened 2025-01-20
fix: update file upload MIME types to use standard format

hexart:
- Replace '*/*' with specific MIME type categories
- Add support for application, audio, image, text and video files
- Remove invalid MIME type warning


dokterbob (changes requested):
Thanks! It took me a while to find out why this makes sense, until I landed through Drop zone's documentation at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#unique_file_type_specifiers

Perhaps we could add some comment describing the why and how of this, referring to aforementioned (or some equivalent) URL?

It's nice to be kind to our future selves! ;)

dokterbob (approved):

PR #1758, opened 2025-01-20
Fix from anthonym1991 #1685

### LINKED ISSUE #1685: Users with large JWT tokens are unauthorized (401) to access chailit due to having a cookie larger than 4kb
Some users with large JWT tokens get a 401 error when trying to log on to Chainlit. The Set-Cookie operation fails.

For these users, the login process fails for all devices/browsers.

jpolvto:
Splits the JWT token over multiple cookies if it's too big. Should resolve #1685. It's a bit hard to check, since my account doesn't have a large JWT token. @dokterbob

### DIFF
--- a/backend/chainlit/auth/cookie.py
+++ b/backend/chainlit/auth/cookie.py
@@ -46,7 +46,7 @@ class OAuth2PasswordBearerWithCookie(SecurityBase):
 
     async def __call__(self, request: Request) -> Optional[str]:
         # First try to get the token from the cookie
-        token = request.cookies.get(_auth_cookie_name)
+        token = reconstruct_token_from_cookies(request.cookies)
 
         # If no cookie, try the Authorization header as fallback
         if not token:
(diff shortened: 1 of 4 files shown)


fcestari:
@jpolvto, I have tested a few times with a large JWT token and I can confirm it has worked on my side.

backend/chainlit/auth/cookie.py
```diff
@@ -76,26 +76,60 @@ async def __call__(self, request: Request) -> Optional[str]:
         return token
 
 
+def reconstruct_token_from_cookies(request_cookies: dict) -> Optional[str]:
```
dokterbob, line 79: I would suggest a clearer signature.

Something like `get_token_from_cookies(cookies: dict[str, str]) -> Optional[str]`.

backend/chainlit/auth/cookie.py
```diff
@@ -76,26 +76,60 @@ async def __call__(self, request: Request) -> Optional[str]:
...
+            break
+        chunk_parts.append(request_cookies[cookie_key])
+        i += 1
+
+    joined = "".join(chunk_parts)
+
+    return joined if joined != "" else None
+
+
 def set_auth_cookie(response: Response, token: str):
```
dokterbob, line 99: We should be careful here to ensure the case where we're overwriting cookies and not deleting old ones.

Imagine:
1. `set_auth_cookie(r, 'incredibly_<...>_long_token')` sets `access_token_0`, `access_token_1`.
2. `set_auth_cookie(r, 'more_reasonable_token')` sets only `access_token_0`.
3. `get_token_from_cookies(🍪)` returns the concatenation of the new `access_token_0` and the _old_ `access_token_1`.

dokterbob (commented):
Thanks for the quick contrib!

We'd love to get this in soon, but as this code will affect nearly all users we need to thread carefully.

In addition to the issues pointed out in the code, I'd love to see unit test coverage of cookie getting and setting. This includes a test for the specific bug pointed out! If you need help on approach to testing, DM me on Discord.

I would also suggest implementing this workaround (which is what it is!) such that the behaviour for typical tokens doesn't change. Plus a test that indeed it doesn't.

One way to do this:
* `access_token_0` present -> look for `_1` etc.
* `access_token` present -> cookie fits,  no need to look for `_1` etc.

When setting cookies, we need to check for chunked cookies and, if they exist, ensure cookies beyond what we're currently setting get deleted!

As a bonus, this will provide a clear distinction between chunked and non-chunked codepaths, making it easy to remove this once a more elegant/sane response than sending extra  4k+

dokterbob (changes requested):
See prior review. ;)

Forgot to set status properly.

dokterbob:
@jpolvto Happy to see your applying requested changes!

If you need support writing tests, let me know. I've a bit of capacity available as this is blocking lot's of users.

dokterbob (approved):

PR #1763, opened 2025-01-21
fix: resolve React warnings and accessibility issues

hexart:
## Description
This PR fixes multiple React warnings, accessibility issues and DOM nesting problems across several components:
- Add DialogTitle for accessibility in Search component
- Prevent state updates during render in ChatProfiles component
- Fix invalid DOM nesting of buttons in ThreadOptions
- Fix invalid DOM nesting of `<h2>` and `<p>` tags in Markdown component
- Add proper key prop to React Fragments in Messages component
- Fix invalid DOM nesting of Alert component in Markdown paragraphs

## Warning before fix
<details><summary>Screenshots</summary>
<p>

截屏2025-01-30 11 58 13
截屏2025-01-30 11 57 54
截屏2025-01-30 11 57 09
</p>
</details> 

## Fixes
1. Fixed Accessibility Issues in `LeftSidebar/Search.tsx`:
   - Added screen reader accessible DialogTitle to Search component
   - Implemented sr-only pattern to maintain visual design
   - Resolved Radix UI Dialog accessibility warning

2. Fixed React Component State Issues in `header/ChatProfiles.tsx`:
   - Moved state updates from render to useEffect in ChatProfiles component
   - Added early return check to prevent invalid renders
   - Improved type safety with optional chaining

3. Optimized Profile Selection Logic in `header/ChatProfiles.tsx`:
   - Split profile selection into separate effects:
     - Handle initial profile selection
     - Handle invalid profile fallback
   - Removed redundant chatProfiles existence checks
   - Improved error handling and edge cases

4. Fixed DOM Nesting Issues:
   - Resolved invalid button nesting in `LeftSidebar/ThreadOptions.tsx`
     - Replace Button element with div element in DropdownMenuTrigger
     - Apply equivalent button styles using buttonVariants function
     - Keep all existing behavior and visual styles intact
     - Maintained original styling and functionality
   - Improved DOM nesting handling in `Markdown.tsx` component
     - Enhanced block element detection to handle headings and Alert components
     - Implemented comprehensive checks for nested divs and component types
     - Added fallback to div container for proper DOM hierarchy
     - Preserved existing markdown styling and functionality
     - Added type safety checks for React elements

5. Fixed `e2e/command/spec.cy.ts` Test Cases:
   - Updated command test to handle whitespace properly

## Testing Done
- **Confirmed browser console is clean with no error messages**
- Verified accessibility of Search dialog with screen readers
- Tested chat profile selection functionality
- Confirmed thread options and list still work as expected
- Validated markdown content renders correctly
- Verified Alert components render properly within markdown content

## Types of changes
- Bug fix (non-breaking change which fixes warnings/issues)
- Accessibility improvements
- No new feature added


frontend/src/components/Markdown.tsx
```diff
@@ -12,9 +12,11 @@ import { visit } from 'unist-util-visit';
 import { ChainlitContext, type IMessageElement } from '@chainlit/react-client';
 
 import { AspectRatio } from '@/components/ui/aspect-ratio';
+/* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
willydouhard, line 15: why is this required?

frontend/src/components/Markdown.tsx
```diff
@@ -12,9 +12,11 @@ import { visit } from 'unist-util-visit';
...
 import { Card } from '@/components/ui/card';
 import { Separator } from '@/components/ui/separator';
 import {
+  /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
willydouhard, line 19: why is this required?

frontend/src/components/LeftSidebar/ThreadList.tsx
```diff
@@ -274,32 +274,32 @@ export function ThreadList({
...
-                            setThreadNewName(thread.name);
-                          }}
-                          className={cn(
-                            'absolute z-20 bottom-0 top-0 right-0 bg-sidebar-accent hover:bg-sidebar-accent hover:text-primary flex opacity-0 group-hover/thread:opacity-100',
-                            isSelected && 'bg-sidebar-accent opacity-100'
+                    <div className="relative">
+                      <Link to={isResumed ? '' : `/thread/${thread.id}`}>
+                        <SidebarMenuButton
+                          isActive={isSelected}
+                          className="relative truncate h-9 group/thread pr-10" // 添加右边距给 ThreadOptions 留空间
```
willydouhard, line 281: I'd like the codebase to completely written in english, could you remove the chinese comment?

frontend/src/components/Markdown.tsx
```diff
@@ -230,18 +232,18 @@ const Markdown = ({
...
         p(props) {
-          return (
-            <p
-              {...omit(props, ['node'])}
-              className="leading-7 [&:not(:first-child)]:mt-4 whitespace-pre-wrap break-words"
-            />
+          const containsBlockElement = React.Children.toArray(props.children).some(
+            child => 
+              React.isValidElement(child) && 
+              (/^h[1-6]$/.test((child.type as string)) || child.type === 'p')
```
willydouhard, line 238: could you provide some details on this?

frontend/src/components/chat/Messages/Message/Content/index.tsx
```diff
@@ -38,7 +38,7 @@ const MessageContent = memo(
     const isMessage = message.type.includes('message');
 
     const outputMarkdown = (
-      <p className="flex flex-col gap-2">
+      <div className="flex flex-col gap-2">  {/* change to div */}
```
willydouhard, line 41: we can remove this comment

willydouhard (changes requested):

frontend/src/components/Markdown.tsx
hexart: Card and Table are no longer used, this is for lint error.

frontend/src/components/Markdown.tsx
hexart: #### 1. Problem Context:
- In HTML, a `<p>` tag cannot contain block-level elements (like `<div>`, `<h1>-<h6>`, or another` <p>`)
- When React encounters such invalid nesting, it throws a warning
- This commonly happens in Markdown rendering when complex content is parsed

#### 2. Solution Breakdown:
```typescript
// Convert children to an array for easier processing
React.Children.toArray(props.children)
``` 
- `React.Children.toArray()` normalizes the children prop into a flat array
- This makes it easier to inspect each child element
```typescript
.some((child) => {
  // First check: Is it a valid React element?
  React.isValidElement(child) &&
  // Second check: Is it a heading (h1-h6) or paragraph?
  (/^h[1-6]$/.test(child.type as string) || child.type === 'p')
})
``` 
- `.some()` checks if any child matches our criteria
- `React.isValidElement(child)` verifies we're dealing with a React element
- The regex `/^h[1-6]$/` checks if the element is any heading tag (h1 through h6)
- `child.type === 'p'` checks if it's a paragraph element

#### 3. Using the Result:
```typescript
return containsBlockElement ? (
  <div className={commonClassNames} {...props} />
) : (
  <p className={commonClassNames} {...props} />
);
``` 
- If block elements are detected, render a `<div>` (which can contain block elements)
- If no block elements are found, safely render a `<p>` tag

#### 4. Why This Works:
- Instead of blindly wrapping content in a `<p>`, we first inspect what we're trying to render
- If we detect any block-level elements that would cause invalid HTML, we use a `<div>` instead
- This ensures we generate valid HTML while maintaining the desired styling and layout

The key insight is that it's better to adapt our container element based on its content rather than forcing content into a container that might not be appropriate for it.

hexart:
### refactor(markdown): enhance paragraph element handling logic
```typescript
// Modify the p component in Markdown.tsx components configuration
p(props) {
  const containsBlockElement = React.Children.toArray(props.children).some(
    (child) => {
      if (!React.isValidElement(child)) return false;
      
      // Check if it's a standard block element
      if (/^h[1-6]$/.test(child.type as string) || child.type === 'p') {
        return true;
      }
      
      // Check if it's an Alert component
      if (child.props?.['data-type'] === 'Alert' || 
          (typeof child.type === 'function' && child.type.name === 'AlertComponent')) {
        return true;
      }

      // Check if there are div elements as children
      if (React.isValidElement(child) && child.props?.children) {
        return React.Children.toArray(child.props.children).some(
          (grandChild) => React.isValidElement(grandChild) && 
            ((grandChild.type === 'div') || (typeof grandChild.type === 'string' && grandChild.type.toLowerCase() === 'div'))
        );
      }
      return false;
    }
  );

  const commonClassNames =
    'leading-7 [&:not(:first-child)]:mt-4 whitespace-pre-wrap break-words';

  // Use div if contains block elements
  if (containsBlockElement) {
    return <div {...omit(props, ['node'])} className={commonClassNames} />;
  }

  // Otherwise use p
  return <p {...omit(props, ['node'])} className={commonClassNames} />;
}
```

sandangel:
Hi, I think this PR is ready to be merged right? cc @willydouhard

frontend/src/components/Markdown.tsx
```diff
@@ -238,12 +238,40 @@ const Markdown = ({
           );
         },
         p(props) {
-          return (
-            <p
-              {...omit(props, ['node'])}
-              className="leading-7 [&:not(:first-child)]:mt-4 whitespace-pre-wrap break-words"
-            />
-          );
+          const containsBlockElement = React.Children.toArray(
```
willydouhard, line 241: this is going to run everytime a token is streamed, it might be a bit costly

hexart: Thank you for reminding me, I'll try to add useMemo  to check the result in cache tomorrow.

hexart:
> there is another issue, which is in #chat-input, I type something then remove all, then the `send` button is not disabled properly. And users now can send empty messages. I wonder if that can also be fixed in this PR? Or should it be fixed in another PR.

I suggest handling the empty message validation in a separate PR. This current one focuses on DOM/rendering fixes, while input validation is a different concern. I'll create a new PR to address the send button and empty message issues.

sandangel:
No worry, I submitted a PR to fix the issue.

willydouhard:
I am not sure what this section does

``` typescript
 const containsBlockElement = useMemo(
            () =>
              React.Children.toArray(props.children).some((child) => {
                if (!React.isValidElement(child)) return false;
                if (
                  /^h[1-6]$/.test(child.type as string) ||
                  child.type === 'p'
                ) {
                  return true;
                }
                if (
                  child.props?.['data-type'] === 'Alert' ||
                  (typeof child.type === 'function' &&
                    child.type.name === 'AlertComponent')
                ) {
                  return true;
                }
                if (React.isValidElement(child) && child.props?.children) {
                  return React.Children.toArray(child.props.children).some(
                    (grandChild) =>
                      React.isValidElement(grandChild) &&
                      (grandChild.type === 'div' ||
                        (typeof grandChild.type === 'string' &&
                          grandChild.type.toLowerCase() === 'div'))
                  );
                }
                return false;
              }),
            [props.children]
```

Other changes look good to me but it would be nice to have some comments explaining why we need to know if the paragraph contains certain blocks.

sandangel:
@willydouhard As I understand, that is for checking when to use `p` and when to use `div` for outer most element. If the elements have h1-h6 or have `p` etc, then the outer element should be `div`, otherwise it can be `p`.

willydouhard:
Can't we always use a div then instead of adding logic?

sandangel:
@willydouhard the intention seems to keep the original `<p>` tag implementation. I think we can just use `<div>` tag and avoid this complex check. I can submit another PR without this check.

hexart:
> Can't we always use a div then instead of adding logic?

While using div everywhere might seem simpler, using semantic HTML tags like <p> when appropriate is important for:
- Accessibility (screen readers)
- SEO
- Document structure clarity

I'll add comments to explain this rationale in the code.

sandangel:
I don't think SEO is related in the chatbot context.

Can you explain why would div affect screen reader and document structure clarity? I do think this logic affect the maintenance effort of the project. Not sure if the benefit worth adding this complex check.

hexart:
> I don't think SEO is related in the chatbot context.
> 
> Can you explain why would div affect screen reader and document structure clarity? I do think this logic affect the maintenance effort of the project. Not sure if the benefit worth adding this complex check.

You make great points about the context of our application. You're right that in a chatbot UI, we might be over-engineering by adding complex checks for semantic HTML.

Would you be open to this simpler approach? Something like:
```typescript
p(props) {
  const commonClassNames = 'leading-7 [&:not(:first-child)]:mt-4 whitespace-pre-wrap break-words';
  return <div {...omit(props, ['node'])} className={commonClassNames} role="paragraph" />;
}
```

sandangel:
That totally makes sense. changing "role" to paragraph and avoid this complex logic LGTM.

sandangel:
Thanks for fixing this. We kind of need this change before we can release with chainlit v2.

sandangel:
I guess we can also use role="article" if needed as mentioned here https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/article_role

hexart:
> I guess we can also use role="article" if needed as mentioned here https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/article_role

Yes, you're right, already fix it.

willydouhard (approved):
lgtm

PR #1787, opened 2025-01-22
Netlify functions

datalayer0:


willydouhard:
A description would be nice! This PR intends to make deployment of chainlit apps on netlify easier if I understand correctly?

Issue #1789, opened 2025-01-22
Replace asyncio event pool with with uvloop


asvishnyakov:
It seems that, for some reason, you replaced `uvloop`, which is used by `uvicorn` by default, with the default `asyncio` event loop. This is negatively impacts performance:

https://github.com/Chainlit/chainlit/blob/85c98ea98508d1b4d87f6e476d268ad3cdd293a7/backend/chainlit/cli/__init__.py#L91-L92
https://github.com/Chainlit/chainlit/blob/85c98ea98508d1b4d87f6e476d268ad3cdd293a7/backend/chainlit/cli/__init__.py#L104-L108

or at least, it feels like performance improved when I replaced this code with the following:

```python
    # Start the server
    uvicorn_config = uvicorn.Config(
        app,
        host=host,
        port=port,
        ws=ws_protocol,
        log_level=log_level,
        ws_per_message_deflate=ws_per_message_deflate,
        ssl_keyfile=ssl_keyfile,
        ssl_certfile=ssl_certfile,
    )
    server = uvicorn.Server(uvicorn_config)
    server.run()
```

Is this true? If so, is there any reason to keep it that way? Is there a workaround or solution to configure it to use `uvloop` instead?

dokterbob:
I'd be extremely surprised if there's an objectively measurable change in performance in a single user case, as most modern machines are perfectly capable of relaying short strings of text at 100Mbit+ rates.

Could you provide a benchmark?

PR #1790, opened 2025-01-22
fix: replace asyncio default event loop with uvloop

asvishnyakov:
This is my proposed solution to #1789. Feel free to edit or close if I'm wrong


willydouhard:
Unfortunately I don't think this is possible, we need re-entrancy for stuff like AskUser

dokterbob:
> Unfortunately I don't think this is possible, we need re-entrancy for stuff like AskUser

But then the tests are passing.

@asvishnyakov Does this mean we can do away with the asyncio dependency?

As I'm not (yet) convinced there's a strong use case for this and @willydouhard suggests it's not possible, I propose we leave this open for a bit until either:

1. A benchmark clearly shows there's performance issues in the current version.
2. Other contributors weigh in on this and/or are able to do additional testing.

The rationale being: if it's not (demonstrably) broken...

asvishnyakov:
@dokterbob I’m new to Python, so I’m not entirely sure myself 😄. I’m trying to investigate performance issues in my app, and while this is definitely not a bottleneck, I was surprised to see it working better after the change. I think we need proper performance tests. Also, it looks like, as a workaround, I can run chainlint as a sub-app and configure uvloop for the main app, if needed.

Issue #1792, opened 2025-01-23
CustomElement with PostGreSQL data layer error:  duplicate key value violates unique constraint "Element_pkey"


yves03:
**Describe the bug**
Using version 2.0.5 of Chainlit with the new official PostGreSQL data layer for data persistence.
When using the newly added custom elements (defined under "./public/elements/custom_element.jsx"), the use of the "updateElement" along with the "callAction" API causes the data layer to insert a new element in the "Element" table with the same ID, resulting in the following error:
> Database error: duplicate key value violates unique constraint "Element_pkey"

I tried changing the code to only use the "callAction" API to call an action that would print something, but that causes the same issue.
Using the "updateElement" alone doesn't cause the same issue.

**To Reproduce**
To reproduce, take the example "Counter" custom element defined at https://docs.chainlit.io/api-reference/elements/custom and add a "callAction" API call that prints something, for example:

```jsx
import { Button } from "@/components/ui/button"
import { X, Plus } from 'lucide-react';

export default function Counter() {
    return (
        <div id="custom-counter" className="mt-4 flex flex-col gap-2">
                <div>Count: {props.count}</div>
                <Button id="increment" onClick={() => {
                    updateElement(Object.assign(props, {count: props.count + 1}));
                    callAction({name: "print_something", payload: {}});
                }}><Plus /> Increment</Button>
                <Button id="remove" onClick={deleteElement}><X /> Remove</Button>
        </div>
    );
}
```

And the action_callback decorated function in my main python file:
```python
@cl.action_callback("print_something")
async def test_print(action):
    print("Hello World !")
```

In the terminal, you'll find the following error:
> Database error: duplicate key value violates unique constraint "Element_pkey"

**Desktop**

- OS: Windows 11
- Browser Microsoft Edge
- Version 132.0.2957.115

desaxce:
> Using the "updateElement" alone doesn't cause the same issue.

I am able to reproduce the same issue with `updateElement` alone on the `Counter` example. 

I believe it comes from `BaseDataLayer.create_element` which performs an `INSERT INTO` without specifying an `ON CONFLICT` strategy. Let me try that.

desaxce:
That works, here's the PR: https://github.com/Chainlit/chainlit/pull/1794

PR #1794, opened 2025-01-23
fix: add conflict strategy to create_element

desaxce:
@yves03 reported https://github.com/Chainlit/chainlit/issues/1792

### Changes:
- override `props` in `create_element` with on conflict strategy
- remove persisted condition on deletion -> need to check why `self.persisted` is false upon element removal :/

### To test:
- set up data layer with https://github.com/Chainlit/chainlit-datalayer/
- use the counter example from https://docs.chainlit.io/api-reference/elements/custom linked to above data layer
- validate that counter increments do show on `props` (count key incremented in DB)
- validate that clicking "Remove" deletes the `Element` record from DB
- add the `callAction` as described in issue above and validate the print works (no errors)


willydouhard (approved):
lgtm

Issue #1819, opened 2025-01-28
Custom FastAPI routes no longer work


s0l4r:
**Describe the bug**
After upgrade from 2.0.2 to 2.0.603 I cannot get custom routes to work, for example https://github.com/Chainlit/cookbook/tree/main/custom-frontend/backend. I've been trying to create the custom endpoint /custom-auth, but the route is never reached.

I believe this may have been introduced in https://github.com/Chainlit/chainlit/pull/1723/files?diff=split&w=0

The custom route is now overwritten of routes by the chainlit catch-all route (e.g., /{full_path:path}). 

```
INFO:     127.0.0.1:62200 - "GET /custom-auth HTTP/1.1" 200 OK
Intercepted path: custom-auth
```

I modified server.py to check if my custom route is activated, but it's caught by the catch all route:

```
@router.get("/{full_path:path}")
async def serve(full_path: str, request: Request):
    """Serve the UI files."""
    print(f"Intercepted path: {full_path}")
    html_template = get_html_template(request.scope["root_path"])
    response = HTMLResponse(content=html_template, status_code=200)

    return response
```

This results in that the default html template is being rendered when I try to access /custom-auth.

**To Reproduce**

Run the example at https://github.com/Chainlit/cookbook/tree/main/custom-frontend/backend with custom auth.

I tried starting the app using the new root-path parameter, but still doesn't work (also tried without --root-path set)

```
"args": [
       "run",
        "--root-path=/chainlit",
        "-h",
        "app.py"
]
```

**Expected behavior**
Custom routes should not be overwritten by Chainlit. 

**Desktop (please complete the following information):**

- OS: MacOS
- Browser Chrome
- Version 131.0.6778.265

**Additional context**

The error might be in how the sub application is mounted, and the changes done in #1723 
```
mount_chainlit(app=app, target="cl-app.py", path="/chainlit"
```

s0l4r:
I've managed to work around it, but it isn't pretty. This is what I did:

create new init_chainlit.py:
```
from fastapi import Request
from chainlit.server import app as chainlit_app # I have to use the main app, the custom app with routes is overwritten
from starlette.middleware.cors import CORSMiddleware
from chainlit.server import _authenticate_user
from chainlit.user import User

# Add CORS middleware
chainlit_app.add_middleware(
    CORSMiddleware,
    allow_origins=["http://localhost:4200"],
    allow_credentials=True,
    allow_methods=["*"],
    allow_headers=["*"],
)

[...]

PR #1825, opened 2025-01-29
fix: prevent chainlit sub app to default all endpoints

willydouhard:
Fixing https://github.com/Chainlit/chainlit/issues/1819

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -447,9 +447,7 @@ def _get_oauth_redirect_error(request: Request, error: str) -> Response:
             "error": error,
         }
     )
-    response = RedirectResponse(
-        url=str(request.url_for("login")) + "?" + params
-    )
+    response = RedirectResponse(url=str(request.url_for("login")) + "?" + params)
     return response
 
 
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -440,7 +440,7 @@ def _get_auth_response(access_token: str, redirect_to_callback: bool) -> Respons
     return JSONResponse(response_dict)
 
 
-def _get_oauth_redirect_error(error: str) -> Response:
+def _get_oauth_redirect_error(request: Request, error: str) -> Response:
     """Get the redirect response for an OAuth error."""
     params = urllib.parse.urlencode(
         {
@@ -448,8 +448,7 @@ def _get_oauth_redirect_error(error: str) -> Response:
         }
     )
     response = RedirectResponse(
-        # FIXME: redirect to the right frontend base url to improve the dev environment
-        url=f"{config.run.root_path.rstrip('/')}/login?{params}",
+        url=request.url_for("login") + "?" + params,
     )
     return response
 
@@ -617,7 +616,7 @@ async def oauth_callback(
         )
 
     if error:
-        return _get_oauth_redirect_error(error)
+        return _get_oauth_redirect_error(request, error)
 
     if not code or not state:
         raise HTTPException(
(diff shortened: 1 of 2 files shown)


backend/chainlit/server.py
```diff
@@ -449,7 +449,7 @@ def _get_oauth_redirect_error(error: str) -> Response:
     )
     response = RedirectResponse(
         # FIXME: redirect to the right frontend base url to improve the dev environment
-        url=f"/login?{params}",  # Shouldn't there be {root_path} here?
+        url=f"{config.run.root_path.rstrip('/')}/login?{params}",
```
dokterbob, line 452: Instead of writing the URL, including `root_path` manually, I would suggest using [`request.url_for()`](https://fastapi.tiangolo.com/reference/request/?h=url_for#fastapi.Request.receive) to generate the URL with root path automatically.

Yes, that does mean we add a request parameter there, but it also means less custom code and so less maintenance and bugs in the future!

backend/chainlit/utils.py
```diff
@@ -128,4 +130,13 @@ def mount_chainlit(app: FastAPI, target: str, path="/chainlit"):
 
     ensure_jwt_secret()
 
+    class ChainlitMiddleware(BaseHTTPMiddleware):
```
dokterbob, line 133: This deserves a docstring describing what it does, and why.

backend/chainlit/utils.py
```diff
@@ -128,4 +130,13 @@ def mount_chainlit(app: FastAPI, target: str, path="/chainlit"):
...
+        async def dispatch(self, request: Request, call_next):
+            if not request.url.path.startswith(path):
+                return JSONResponse(status_code=404, content={"detail": "Not found"})
+
+            return await call_next(request)
```
dokterbob, line 138: If I read this correctly this:
1. Looks whether a request path starts with the app's mount path.
2. Returns 404 if it doesn't.

But this is standard functionality, we should not have to implement this.

Simply mounting chainlit at `path` should ensure that no requests outside of `path` ever get to it. If that is not the case, that in itself is a bug and it should be addressed appropriately.

dokterbob (changes requested):
It's a good start but I think we need a more rigorous solution: we should not have to refer to `root_path` or such things anywhere in our code, but rather move to use of `request.url_for()`. That way our code becomes less error prone and less sensitive to (the many) edge cases in HTTP routing.

## References:
* https://fastapi.tiangolo.com/advanced/templates/?h=url_for#template-context-values
* https://fastapi.tiangolo.com/reference/request/?h=url_for#fastapi.Request.url_for

## Regression unit tests
In addition, I would like to request we start writing unit tests for this -- I think https://github.com/Chainlit/chainlit/blob/main/backend/tests/test_server.py is the place.

I'm not 100% sure it's required but we could programmatically spin up a uvicorn. If we do, we can [run uvicorn programmatically](https://www.uvicorn.org/#running-programmatically), e.g. in a pytest fixture using [yield](https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended) for teardown. Make sure to use a random port and check for availability.

backend/chainlit/utils.py
willydouhard: This is not the case because in server.py, we have an endpoint catching all routes to serve the index.html which is a common pattern in SPWA

```python
@router.get("/{full_path:path}")
async def serve(request: Request):
    """Serve the UI files."""
    html_template = get_html_template(request.scope["root_path"])
    response = HTMLResponse(content=html_template, status_code=200)

    return response
```

When submounted on a path, it collides with the main app routes. Adding the middleware ensure we only reach this route if it starts with the sub path

PR #1826, opened 2025-01-29
feat(pwa): Adds PWA support

Alihassanc5:


SeaRedm:
Do PWAs work properly on Chainlit? I know you can install them as one but it seems to be riddled with issues.

PR #1836, opened 2025-01-30
feat: support adding custom links on the nav header

fgalind1:
This PR allows to configure custom links to be added in the menu/navigation header by adding them in the config.toml file

For example:

```
[[UI.header_links]]
    name = "Confluence"
    icon_url = "https://www.atlassian.com/favicon.ico"
    url = "https://www.atlassian.com"

[[UI.header_links]]
    name = "Issues"
    icon_url = "https://mintlify.s3-us-west-1.amazonaws.com/chainlit-43/_generated/favicon/favicon.ico?v=3"
    url = "https://github.com/Chainlit/chainlit/issues"
```

That will result in the following:

image

### DIFF
--- a/frontend/src/components/ButtonLink.tsx
+++ b/frontend/src/components/ButtonLink.tsx
@@ -34,6 +34,7 @@ export default function ButtonLink({ name, iconUrl, url }: ButtonLinkProps) {
                     ? apiClient.buildEndpoint(iconUrl)
                     : iconUrl
                 }
+                className={'h-6 w-6'}
                 alt={name}
               />
             </a>
--- a/frontend/src/components/ButtonLink.tsx
+++ b/frontend/src/components/ButtonLink.tsx
@@ -1,3 +1,7 @@
+import { useContext } from 'react';
+
+import { ChainlitContext } from '@chainlit/react-client';
+
 import { Button } from '@/components/ui/button';
 import {
   Tooltip,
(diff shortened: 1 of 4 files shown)


frontend/src/components/ButtonLink.tsx
```diff
@@ -0,0 +1,39 @@
...
+        <Tooltip>
+          <TooltipTrigger asChild>
+            <Button
+              variant="ghost"
+              size="icon"
+              className="text-muted-foreground hover:text-muted-foreground"
+            >
+            <a href={url} target="_blank">
+              <img
+                src={iconUrl}
```
lamylio, line 27: Shouldn't we check if the path starts with `/public` ?

Like this : 
```ts
src={
  iconUrl?.startsWith('/public')
  ? apiClient.buildEndpoint(iconUrl)
  : iconUrl
}
```

fgalind1: good call 👍 added in https://github.com/Chainlit/chainlit/pull/1836/commits/4cc1f3aea3a8d07156c4bc9d50bd71ac476b61ff

willydouhard:
That's great! The size of the button is too big i think. Also we should remove the current `github` config which does exactly that buy specifically for github, this can be now generalised with this PR

fgalind1:
> That's great! The size of the button is too big i think. Also we should remove the current `github` config which does exactly that buy specifically for github, this can be now generalised with this PR

indeed, I've adjusted the size, now they show more uniform with the theme/user icons
image

prev size:
image

fgalind1:
I believe it is indeed not used, so I've created a PR https://github.com/Chainlit/chainlit/pull/1865 that removes this in favor of this generalized approach

willydouhard (approved):
lgtm

PR #1840, opened 2025-01-31
Fix: prevent showing 0's on resumed thread if AskAction/File was used

lamylio:
Hello, the redesign of Chainlit on V2 introduced back an issue (https://github.com/Chainlit/chainlit/issues/1548) where the assistant messages would display "0" in some case when thread contains a AskActionMessage or AskFileMessage element.

This simply applies back the same fix as the PR https://github.com/Chainlit/chainlit/pull/1617 to change the logic on the frontend.

---
Before : 
image

After:
image


willydouhard (approved):
lgtm

PR #1842, opened 2025-01-31
fix: reintroduce including URL location after UI refactor

fgalind1:
Partial code of the feature introduced in https://github.com/Chainlit/chainlit/pull/1403 was removed after the recent UI refactoring.

This PR reintroduces the missing parts to allow including the current location/URL where the user is as part of the message metadata for further processing in `on_message(message: cl.Message)`


asvishnyakov:
This is useful not only in Copilot, but, for example, when user references question to something on current page, where RAG app displayed. This will be very useful for me, as currently I need to send this information through window message and store in memory. Thank you @fgalind1 !

willydouhard (approved):
lgtm

Issue #1845, opened 2025-02-03
Azure AD OAUTH requires fully qualified scope (https://graph.microsoft.com/offline_access)


GhimBoon:
**Describe the bug**
AzureAD authentication started failing from 2.0.603 when offline_access was added without the Microsoft Graph prefix.

**To Reproduce**

1. Upgrade to 2.0.603 or above
2. Login using AzureAD
3. Will encounter 502 error

**Expected behavior**
Login will work as expected

**Screenshots**
Forgot to take the screenshot

**Additional context**
Patched the scope to https://graph.microsoft.com/offline_access and it started working. Some production environments require fully qualified scopes to work.

ajosegun:
<img width="370" alt="Image" src="https://github.com/user-attachments/assets/8408dc51-89f3-4f49-badc-6c760cb5cbc1" />

Seems I have the same issue

fcestari:
The PR #1846 actually had a different behavior to me - the refresh_token are no longer saved in the user metadata as per #1599

GhimBoon:
Update: I'm reverting it in #1869.

My issue is actually caused by kubernetes nginx that got a 502 due to the limited proxy_buffer sizing. The offline_access refresh_tokens requires a larger buffer size that kubernetes doesn't handle by default.

Adding these to my ingress solved it:
    nginx.ingress.kubernetes.io/proxy-buffer-size: "64k"
    nginx.ingress.kubernetes.io/proxy-buffers: "8 64k"

GhimBoon:
Closing this issue since this is not a bug.

PR #1846, opened 2025-02-03
Update offline_access scope to be fully qualified with the prefix

GhimBoon:
Modify offline_access to have fully qualified Microsoft graph prefix.
Fix for #1845

### DIFF
--- a/backend/chainlit/oauth_providers.py
+++ b/backend/chainlit/oauth_providers.py
@@ -183,7 +183,7 @@ class AzureADOAuthProvider(OAuthProvider):
         self.authorize_params = {
             "tenant": os.environ.get("OAUTH_AZURE_AD_TENANT_ID"),
             "response_type": "code",
-            "scope": "https://graph.microsoft.com/User.Read offline_access",
+            "scope": "https://graph.microsoft.com/User.Read https://graph.microsoft.com/offline_access",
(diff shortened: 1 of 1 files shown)


willydouhard (approved):

fcestari:
Hi @GhimBoon, how can one fetch the refresh token now? Does it need any extra config in the service principal?

fcestari:
Hi @GhimBoon, how can one fetch the refresh token? After this PR, it returns null for me. Is there any extra config needed in the service principal?

GhimBoon:
> Hi @GhimBoon, how can one fetch the refresh token? After this PR, it returns null for me. Is there any extra config needed in the service principal?

Have you checked if admin consent is approved for offline_access in your enterprise app?

GhimBoon:
image

fcestari:
Yes, it has been granted

arnemorten:
I believe offline_access is part of the openid spec and should not use a qualified domain.

GhimBoon:
That is weird. OAUTH only worked for me with the fully qualified domain.

GhimBoon:
OK, I tried again. Refresh_token is not available when fully qualified domain is added. This needs to be reverted, but the issue of getting 502 is still a problem.

PR #1852, opened 2025-02-04
fix: disable submit button when input is empty

sandangel:

when I type something then delete to empty string in the input, the actual value is `\n`, so the button was not disabled.


willydouhard (approved):
lgtm

PR #1856, opened 2025-02-05
chore: allow chainlit to be built as dependencies

sandangel:

Adding composite: true so that chainlit can built incrementally when referenced in another projects.
This is to support monorepo cloning chainlit and build . This should not be an issue with current chainlit project.


willydouhard (approved):
lgtm

PR #1865, opened 2025-02-05
chore: remove github config not used

fgalind1:
the github button in the UI wasn't really working, however this could now be achieved using `UI.header_links` from https://github.com/Chainlit/chainlit/pull/1836

### DIFF
--- a/backend/chainlit/config.py
+++ b/backend/chainlit/config.py
@@ -127,9 +127,6 @@ name = "Assistant"
 # Chain of Thought (CoT) display mode. Can be "hidden", "tool_call" or "full".
 cot = "full"
 
-# Link to your github repo. This will add a github button in the UI's header.
-# github = ""
-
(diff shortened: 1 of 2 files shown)


willydouhard:
I think this is also used in the frontend

fgalind1:
> I think this is also used in the frontend

it looks like it was removed here https://github.com/Chainlit/chainlit/commit/5ecfbcedb62382998c49011d7bcb247f332a4ed0#diff-bd1b43018992726f6f2bb7330dd53a3b717d82e44a3214640f2a01f0bd8b57ad and never reintroduced back

willydouhard (approved):
lgtm

akotlar:
As a result of this, you now force Chainlit into the header og:url tag, would you all consider re-adding configuration of the header?

asvishnyakov:
@akotlar You may open an issue with feature request

PR #1881, opened 2025-02-09
fix: Smooth Scroll Button Behavior

hexart:
# Smooth Scroll Button Behavior Fix

This PR addresses the issue where the scroll button briefly reappears during smooth scrolling animation.

## Problem
When using behavior: 'smooth' for scrolling, the scroll button would flicker due to handleScroll being triggered during the animation process, causing showScrollButton to be set to true before reaching the bottom.

## Solution
Added an isScrolling state to track the scroll animation status. Instead of using a fixed timeout, we now monitor the actual scroll position to detect when scrolling has completely stopped. During the animation, handleScroll updates are skipped to prevent unwanted button state changes.

## Edge Case Handling
The solution also handles user interaction during scroll animation. When users interrupt auto-scroll with mouse wheel, the scroll button visibility is immediately updated when scrolling stops, accurately reflecting the current position without requiring additional scroll input.

## Verification Process
<details><summary>Details</summary>
<p>
I created a test component to verify both the issue and the solution:

Created a scroll container with logging capabilities to track realtime scroll events and button states
Reproduced the original issue, logs showed:
```
01:17:10 - Scroll button clicked
01:17:10 - Set showScrollButton to false
01:17:10 - handleScroll - atBottom: false, showScrollButton set to: true
// Multiple similar logs during animation
01:17:11 - handleScroll - atBottom: true, showScrollButton set to: false
``` 
3. Tested the solution, new logs showed:
``` 
01:18:32 - Scroll button clicked
01:18:32 - Set isScrolling to true
01:18:32 - Set showScrollButton to false
01:18:32 - handleScroll skipped - isScrolling: true
// Multiple skipped events during animation
01:18:33 - Set isScrolling to false
``` 
4. On chainlit, test logs showed:

This demonstrates that the solution successfully detects the actual end of scroll animation and prevents button flickering while maintaining smooth scrolling behavior.

</p>
</details>

### DIFF
--- a/libs/react-client/src/useChatSession.ts
+++ b/libs/react-client/src/useChatSession.ts
@@ -62,6 +62,7 @@ const useChatSession = () => {
   const setIsAiSpeaking = useSetRecoilState(isAiSpeakingState);
   const setAudioConnection = useSetRecoilState(audioConnectionState);
   const resetChatSettingsValue = useResetRecoilState(chatSettingsValueState);
+  const setChatSettingsValue = useSetRecoilState(chatSettingsValueState);
   const setFirstUserInteraction = useSetRecoilState(firstUserInteraction);
   const setLoading = useSetRecoilState(loadingState);
   const wavStreamPlayer = useRecoilValue(wavStreamPlayerState);
@@ -189,6 +190,9 @@ const useChatSession = () => {
         }
         if (thread.metadata?.chat_profile) {
           setChatProfile(thread.metadata?.chat_profile);
+        }
+        if (thread.metadata?.chat_settings) {
+          setChatSettingsValue(thread.metadata?.chat_settings);
         }
         setMessages(messages);
         const elements = thread.elements || [];


willydouhard:
I had issue with the scroll animation on my end, on which browser did you test it?

hexart:
> I had issue with the scroll animation on my end, on which browser did you test it?

I tested it on macOS with:
- Chrome 133.0.6943.54
- Safari 18.3

Both browsers support smooth scrolling behavior well during my testing. No issues found.

willydouhard:
The issue is that the animation retriggers the `handleScroll` method and resets showScrollButton to true I think

hexart:
> The issue is that the animation retriggers the `handleScroll` method and resets showScrollButton to true I think

You're right! I've resubmitted the PR with a fix and detailed explanation of how we verified and solved the issue.

willydouhard:
I think this is working but it also brings a lot of complexity for "just" a scroll animation. Not sure this is worth it

hexart:
> I think this is working but it also brings a lot of complexity for "just" a scroll animation. Not sure this is worth it

I appreciate your minimalist approach, that's a fair point. We actually have a few options with increasing complexity:

1. Simply change the 'instant' to 'smooth' parameter, ignoring the handleScroll retriggering issue
2. Add a 500ms timeout(or more) to prevent handleScroll during animation, which would handle most cases
3. Monitor scroll position to detect actual animation end, ignoring user wheel interruptions
4. Monitor scroll position and also handle user wheel interruptions (current implementation)

Which approach do you think strikes the best balance between user experience and code simplicity?

Issue #1697, opened 2025-01-14
Oauth Login Page - Use An Image Other Than Favicon - 2.0.2


mihidumh:
**Is your feature request related to a problem? Please describe.**
The Oauth login page has a half-screen image of the favicon, and it is always grayscaled.

**Describe the solution you'd like**
I would like to be able to use an image of my own choice, and preferably not grayscaled.

Huxb12138:
### To change Customise LoginPage image
You can follow steps in docs, by renaming your image to 
![image](https://github.com/user-attachments/assets/c5a29ba1-b97e-4c28-97dd-70af801e5928)
> https://docs.chainlit.io/customisation/custom-logo-and-favicon

### To change grayscale 
you can go to external libraries to find a site-packages names chainlit.
Go to 
![image](https://github.com/user-attachments/assets/04a30b6c-a838-4d68-8fdc-3cb201084196)
Filter and change `brightness` and `grayscale`  to 1 and 0% like this:
![image](https://github.com/user-attachments/assets/ff7ef83f-40e5-40c7-a0ff-5e3779fc5f66)
![image](https://github.com/user-attachments/assets/ffd70463-117f-4962-bf35-f2dbc0ea515d)

Better clean the cache of browser, and you will get your dark LoginPage image the same as light theme.

DawarWaqar:
@Huxb12138 as asked by @mihidumh what if we want to use an image other than the favicon?

mihidumh:
added in 2.4.400

PR #1885, opened 2025-02-11
feat: Add custom login background image and filter

hexart:
## Feature Description
This PR introduces the ability to customize the login page background image and apply image filters, providing more flexible UI customization options for Chainlit applications.

### Screenshot
example: `login_page_image_filter = "sepia -hue-rotate-90 blur-sm"`

- Related issue #1697 

## Implementation Details
- Added `login_page_image` and `login_page_image_filter` to the configuration return in `backend/chainlit/auth/__init__.py`
- Add usage description in `backend/chainlit/config.py`
- Updated type definitions in `libs/react-client/types/config.ts`
- Added safelist in `tailwind.config.js`
- Modified `frontend/src/pages/Login.tsx` to support custom image and filter

## Technical Details
<details><summary>Supports filter types: </summary>
<p>

```
filter-none
blur-sm
blur
blur-md
blur-lg
blur-xl
blur-2xl
blur-3xl

brightness-0
brightness-50
brightness-75
brightness-90
brightness-95
brightness-100
brightness-105
brightness-110
brightness-125
brightness-150
brightness-200

contrast-0
contrast-50
contrast-75
contrast-100
contrast-125
contrast-150
contrast-200

grayscale-0
grayscale

hue-rotate-0
hue-rotate-15
hue-rotate-30
hue-rotate-60
hue-rotate-90
hue-rotate-180
-hue-rotate-0
-hue-rotate-15
-hue-rotate-30
-hue-rotate-60
-hue-rotate-90
-hue-rotate-180

invert-0
invert

saturate-0
saturate-50
saturate-100
saturate-150
saturate-200

sepia-0
sepia
``` 

</p>
</details>


michelem09:
Please merge this because actual favicon is not usable at all in this context.

frontend/src/lib/utils.ts
```diff
@@ -15,6 +15,93 @@ export const hasMessage = (messages: IStep[]): boolean => {
   );
 };
 
+export const generateFilterStyle = (filter?: string): string => {
```
willydouhard, line 18: why not let the user define tailwind classes and accept those? do we really need this validation?

willydouhard (changes requested):
lgtm overall

frontend/src/lib/utils.ts
hexart: Based on my testing, I initially tried using Tailwind classes for styling, but found several limitations:

1. Tailwind doesn't support many filter effects beyond basic grayscale
2. It doesn't allow for precise parameter adjustments (like grayscale-50% or blur-5px)
3. Many filter classes like brightness and contrast weren't working in the environment

This is why I implemented standard CSS style handling instead. For validation, I believe it provides better security and predictability without significantly impacting performance, since filter styles are only processed once during rendering rather than continuously.

The current approach gives users much more flexibility while maintaining good performance and security practices. However, if you feel the validation is unnecessary, I'm happy to remove that portion of the code.

asvishnyakov: @hexart I'm sorry I interrupt, but
1. There is literally a list with all or almost all filters you used, like [saturate](https://v3.tailwindcss.com/docs/saturate)
2. You can use arbitrary values, like `blur-[5px]`
3. This is probably because Tailwind strip unused classes at build time, but you can use [safe lists](https://v3.tailwindcss.com/docs/content-configuration#safelisting-classes) to prevent it for certain classes

hexart: @asvishnyakov thank you for the information, I'll try it later.

hexart: @asvishnyakov Yes, it works. But what filters should we add into tailwind.config.js?
```
safelist: [
  "sepia-[1]",
  "hue-rotate-[310deg]",
  "blur-[5px]",
],
```

hexart: > @hexart I'm sorry I interrupt, but
> 
> 1. There is literally a list with all or almost all filters you used, like [saturate](https://v3.tailwindcss.com/docs/saturate)
> 2. You can use arbitrary values, like `blur-[5px]`
> 3. This is probably because Tailwind strip unused classes at build time, but you can use [safe lists](https://v3.tailwindcss.com/docs/content-configuration#safelisting-classes) to prevent it for certain classes

Tailwind safelist does not support dynamic variable concatenation, unless we fully enumerate all possibilities, which I feel is not appropriate. Therefore, the current expression supports all internal image filters, but does not support dark, light, hover, and focus variants.
```
safelist: [
{
	pattern: /^(filter-none|blur(?:-\w+)?|brightness-\d+|contrast-\d+|grayscale(?:-\d+)?|hue-rotate-\d+|-hue-rotate-\d+|invert(?:-\d+)?|saturate-\d+|sepia(?:-\d+)?)$/
}
],
```

willydouhard: I confirm the issue is probably about unused class stripping

hexart: I’ve tried, it doesn’t work.

frontend/src/pages/Login.tsx
```diff
@@ -101,11 +101,16 @@ export default function Login() {
...
           <img
-            src={apiClient.buildEndpoint('/favicon')}
+            src={
+              config?.ui?.login_page_image ||
+              apiClient.buildEndpoint('/favicon')
+            }
             alt="Image"
-            className="absolute inset-0 h-full w-full object-cover dark:brightness-[0.2] dark:grayscale"
+            className={`absolute inset-0 h-full w-full object-cover ${
+              config?.ui?.login_page_image_filter ?? ''
```
willydouhard, line 112: should we default to the current value?

willydouhard (changes requested):
if the dark modifier does not work maybe we should just have to configs, one for light and the other for dark? we could make that similar to how logos work and that would be simpler? https://docs.chainlit.io/customisation/custom-logo-and-favicon

frontend/src/pages/Login.tsx
hexart: The issue with supporting arbitrary values like `brightness-[0.2]` in the config.toml is related to how Tailwind CSS works during the build process. Even though the project has a safelist that attempts to preserve filter classes, there are fundamental limitations:

When Tailwind builds the final CSS, it purges unused classes to reduce file size. While standard values like `grayscale` are preserved, arbitrary values in square brackets don't survive this process unless they appear as literal strings in the source code.

The project's current safelist pattern only matches predefined values:
```js
/^(filter-none|blur(?:-\w+)?|brightness-\d+|contrast-\d+|grayscale(?:-\d+)?|hue-rotate-\d+|-hue-rotate-\d+|invert(?:-\d+)?|saturate-\d+|sepia(?:-\d+)?)$/
```

This means dynamically interpolated strings with arbitrary values like `brightness-[0.2]` or variants like `dark:` combined with arbitrary values will be removed during the build process.

I've implemented the separate light/dark filter approach as you suggested, which elegantly solves these technical limitations by allowing users to specify different filters for each theme.

MiesnerJacob:
would love to use this!

willydouhard (approved):
lgtm

Issue #1764, opened 2025-01-21
Vite Build Warnings: Sourcemap Errors and Large Chunk Sizes


hexart:
### Problem Description

During the Chainlit frontend build process using `pnpm build`, the following issues were encountered:

#### 1. Sourcemap Resolution Errors
The build process reported unresolvable sourcemap locations in several files:
- `src/components/ui/dialog.tsx`
- `src/components/ui/sheet.tsx`
- `src/components/chat/Messages/Message/Buttons/FeedbackButtons.tsx`

These errors might impact debugging experience by preventing accurate error source tracking.

#### 2. Large Chunk Size Warnings
Two JavaScript chunks significantly exceed the recommended size limit of 500 kB:
- `dist/assets/index-C1aQOH-P.js` (3,702.30 kB)
- `dist/assets/react-plotly-Comr0_VG.js` (3,762.38 kB)

Oversized chunks can negatively affect:
- Initial page load performance
- Application responsiveness
- User experience, especially on slower networks

### Environment Details
- Vite version: 5.4.8
- Package manager: pnpm
- Build status: Completed successfully with warnings
- Operating System: macOS15.2 (24C101) on M1 Mac

dokterbob:
Thanks for the feedback.

We just merged a PR enabling source maps, could you check whether that solves the 1st issue?

In general, in situations like this, it would be best to create 2 different issues, as they need to be addressed separately!

PR #1888, opened 2025-02-12
fix: Remove unnecessary 'use client' directives to fix sourcemap errors

hexart:
## Issue Description
When running `poetry install --with tests --with mypy --with dev`, the build process fails with sourcemap errors in three React components. These errors occur during the Vite build process for both `@chainlit/copilot` and `@chainlit/app`:
- Related issue #1764 

```
vite v5.4.14 building for production...
Error when using sourcemap for reporting an error: Can't resolve original location of error.
``` 
The errors consistently appear in three files:
- `src/components/ui/dialog.tsx`
- `src/components/ui/sheet.tsx`
- `src/components/chat/Messages/Message/Buttons/FeedbackButtons.tsx`

## Root Cause
The affected components include the `'use client'` directive, which is specific to Next.js and unnecessary in a Vite project. These components appear to be generated using shadcn/ui CLI tool, which defaults to Next.js templates and automatically adds the 'use client' directive. While this directive is essential in Next.js for declaring client-side components, it's not recognized or needed in our Vite-based build system and interferes with sourcemap generation.

## Changes
This PR removes the `'use client'` directive from:
- `dialog.tsx`
- `sheet.tsx`
- `FeedbackButtons.tsx`

## Testing
To verify this fix:
1. Run `poetry install --with tests --with mypy --with dev`
2. Confirm that the build completes without sourcemap errors

## Impact
This change:
- Fixes the sourcemap generation errors
- Removes unnecessary code
- Has no functional impact on the components
- Does not require any configuration changes
- Introducing PR `#1888` - a lucky number in Chinese culture that represents prosperity! 🎉

### DIFF
--- a/frontend/src/components/chat/Messages/Message/Buttons/FeedbackButtons.tsx
+++ b/frontend/src/components/chat/Messages/Message/Buttons/FeedbackButtons.tsx
@@ -1,5 +1,3 @@
-'use client';
-
 import { MessageContext } from '@/contexts/MessageContext';
 import { MessageCircle, ThumbsDown, ThumbsUp } from 'lucide-react';
 import { useCallback, useContext, useState } from 'react';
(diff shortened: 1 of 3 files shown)


willydouhard (approved):
lgtm

PR #1909, opened 2025-02-19
fix: allow setting custom cookie name via env var

sandangel:


willydouhard (approved):
lgtm

PR #1910, opened 2025-02-19
chore(slack): Remove redundant thread ID generation

0xRichardH:
**Thread ID Generation Inconsistency:** 

There are two different thread ID generation approaches in the handle_message function. The first one is immediately overwritten by the second:

```diff 
# backend/chainlit/slack/app.py

- thread_id = str(
-     uuid.uuid5(
-         uuid.NAMESPACE_DNS,
-         message["channel"] + datetime.today().strftime("%Y-%m-%d"),
-     )
- )
 thread_ts = message.get("thread_ts", message["ts"])
 thread_id = str(uuid.uuid5(uuid.NAMESPACE_DNS, thread_ts))
```

### DIFF
--- a/backend/chainlit/slack/app.py
+++ b/backend/chainlit/slack/app.py
@@ -2,7 +2,6 @@ import asyncio
 import os
 import re
 import uuid
-from datetime import datetime
 from functools import partial
 from typing import Dict, List, Optional, Union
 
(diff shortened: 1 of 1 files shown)


backend/chainlit/slack/app.py
```diff
@@ -339,12 +338,6 @@ async def handle_app_mentions(event, say):
 
 @slack_app.event("message")
 async def handle_message(message, say):
-    thread_id = str(
-        uuid.uuid5(
-            uuid.NAMESPACE_DNS,
-            message["channel"] + datetime.today().strftime("%Y-%m-%d"),
-        )
-    )
```
willydouhard, line 347: do we want to keep the per day segmentation? If not this means that we will get one thread per channel which can grow pretty fast.

0xRichardH: `thread_ts` is a thread message key, which means we will get one chainlit thread per Slack thread.

https://api.slack.com/messaging/retrieving
image

PR #1929, opened 2025-02-26
fix(react-client): Chat settings won't be resumed when resuming thread

tcx4c70:
When resuming thread, the backend has already sent the chat setting values to the frontend. However,
the frontend doesn't resume the values. As a result, the setting button will be disappeared after resuming
the thread.

This PR will resume the setting values in the frontend when resuming the thread to fix the problem.


willydouhard (approved):
LGTM

Issue #1904, opened 2025-02-18
Does --root-path require reverser proxy after version 2.0.6, setup on local docker?


ChirawatCh:
I run chainlit app in docker (MacOS), before version 2.0.6 the root path setup with this command work perfectly. But after version 2.0.6 when i access my app with this URL: http://localhost:8000/chatbot/v1, it got white page!!. How do solve this issue? Or does it require reverse proxy server?

CMD ["python", "-m", "chainlit", "run", "app.py", "-h", "--port", "8000", "--root-path", "/chatbot/v1","--host", "0.0.0.0"]

ee-juni:
I'm experiencing the same problem! 
I'm running v.2.2.1.
Please send help 😢

willydouhard:
Will be fixed in the next release!

tibbetson42:
Still an issue in 2.5.5, so I don't think this was resolved!

PR #1723, opened 2025-01-16
Use asgi root path to handle mount point

amol-:
Changes chainlit server to rely on ASGI root_path variable to know where the application is mounted on.
This addresses both cases where chainlit is proxied to non root paths, mounted as a subapp or both combined.

I tested this by serving the following app

```
from fastapi import FastAPI
from chainlit.utils import mount_chainlit

app = FastAPI()

@app.get("/app")
def read_main():
    return {"message": "Hello World from main app"}

mount_chainlit(app=app, target="clapp.py", path="/chainlit")
```

served with

```
uvicorn app:app --host 0.0.0.0 --port 80 --root-path=/proxy
```

The application is correctly available at `http://localhost/proxy/chainlit/` and loads, previously it would not load trying to fetch css and js files from wrong path.

This might also fix https://github.com/Chainlit/chainlit/issues/1440 https://github.com/Chainlit/chainlit/issues/1707 and https://github.com/Chainlit/chainlit/issues/1397 but I didn't have the opportunity to confirm that.

PR #1939, opened 2025-02-28
revert PREFIX of fastapi server.

shkangomelet:
After the changes introduced in [this pull request](https://github.com/Chainlit/chainlit/pull/1723/files), issue [#1904](https://github.com/Chainlit/chainlit/issues/1904) occurred, leading to a rollback of the changes.

I tested running the following script locally using the command:

```
chainlit run demo.py -w --root-path /app
```

demo.py

```python
import chainlit as cl


@cl.step(type="tool")
async def tool():
    # Fake tool
    await cl.sleep(2)
    return "Response from the tool!"


@cl.on_message  # this function will be called every time a user inputs a message in the UI
    """
    This function is called every time a user inputs a message in the UI.
    It sends back an intermediate response from the tool, followed by the final answer.

    Args:
        message: The user's message.

    Returns:
        None.
    """


    # Call the tool
    tool_res = await tool()

    await cl.Message(content=tool_res).send()

```

The script runs without any issues in this setup. However, I have not yet investigated any potential side effects. I would appreciate it if you could review this further.

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -74,8 +74,6 @@ from ._utils import is_path_inside
 mimetypes.add_type("application/javascript", ".js")
 mimetypes.add_type("text/css", ".css")
 
-PREFIX = os.environ.get("CHAINLIT_ROOT_PATH", "")
-
 @asynccontextmanager
 async def lifespan(app: FastAPI):
     """Context manager to handle app start and shutdown."""
@@ -201,7 +199,7 @@ asgi_app = socketio.ASGIApp(
     socketio_path="",
 )
 
-app.mount(f"{PREFIX}/ws/socket.io", asgi_app)
+app.mount(f"{config.run.root_path}/ws/socket.io", asgi_app)
 
 app.add_middleware(
     CORSMiddleware,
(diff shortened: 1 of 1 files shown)


willydouhard:
Can you attach a screen shot of the network tab when you get the blank page? I would love to see the (wrong) path of the js/css files the browser attempts to load.

shkangomelet:
image
image
image
image

willydouhard:
I merged https://github.com/Chainlit/chainlit/pull/1955. I will release it tomorrow, please let me know if it helps. I would rather not revert this if possible.

shkangomelet:
@willydouhard https://github.com/Chainlit/chainlit/pull/1955 still not work. I don't know the reason why, but I still get blank page.

image
image
image

shkangomelet:
[a96fd80](https://github.com/Chainlit/chainlit/pull/1939/commits/a96fd80d668cdff06744bcd7c994dc0bfe51c34e) This is working! If you don't want to call environment variable from this file, it may be better.

willydouhard:
I was not able to make it work so I reverted to prefix but with some other details -> https://github.com/Chainlit/chainlit/pull/1968. Thank you for your help!

PR #1944, opened 2025-03-02
Upgrade aiofiles to 24.1.0

GitAlexxx:


willydouhard:
Why do you need to force the upgrade? This might hinder other developers which use different dependencies.

AlexAisec:
Hi @willydouhard !

Chainlit is actively used, and the limitation on the old version of aiofiles may prevent its integration into modern projects, aiofiles 24.1.0 includes fixes and improvements, which increases stability.
There are no breaking changes in 24.1.0 that could disrupt the operation of chainlit.

I understand that the forced upgrade may affect other developers, but the current limitation is already creating problems. It might be worth at least extending the range (>=23.1.0,<25.0.0) or conducting compatibility testing.

GitAlexxx:
Hi @willydouhard !

Chainlit is actively used, and the limitation on the old version of aiofiles may prevent its integration into modern projects, aiofiles 24.1.0 includes fixes and improvements, which increases stability.
There are no breaking changes in 24.1.0 that could disrupt the operation of chainlit.

I understand that the forced upgrade may affect other developers, but the current limitation is already creating problems. It might be worth at least extending the range (>=23.1.0,<25.0.0) or conducting compatibility testing.

willydouhard:
I suggest we start by extending the range as you proposed!

willydouhard (approved):
LGTM

Issue #1799, opened 2025-01-24
Custom element props not showing after switching chat or refreshing the page, Persistency issue


komikndr:
## Information
1. I am using a custom data layer with PostgreSQL (SQLAlchemy) and MinIO.
2. I am unsure whether this issue is a bug or my mistake, but I am inclined to believe it is the latter.

## Checklist
1. Other elements, such as Image, DataFrame, and Plotly, have been tested and are fully functional.
2. The custom element exists and is successfully populated in both PostgreSQL and MinIO.

When I send a custom element using .send(), its props are successfully populated. However, if I refresh the page, I only see a blank UI without any props.

As a sanity check, I created a DataFrame, refreshed the page, and inspected the browser network activity. The DataFrame data and its corresponding UI component were successfully loaded.

![Image](https://github.com/user-attachments/assets/87d20c67-6e60-42b5-a5ee-7861b4ec271f)

3. But if i am using my own component 

![Image](https://github.com/user-attachments/assets/7c06128d-5412-40eb-8886-530d01263baa)

What it supposed to be

![Image](https://github.com/user-attachments/assets/7a7a500a-9d24-4672-b87a-ea32aceb642b)

And yes this is from the example of custom element. My gut feeling is i should implement update method on the JSX file

desaxce:
> When I send a custom element using .send(), its props are successfully populated. However, if I refresh the page, I only see a blank UI without any props.

If the new props do appear in the database, a page refresh triggers `get_thread` and that function returns the DB `props` for each custom element. 

One example I tested to update/delete custom elements is:
https://docs.chainlit.io/api-reference/elements/custom#example-of-a-counter-element

komikndr:
Thanks for reply.

I’ve already checked my PostgreSQL and MinIO, and both have the custom properties. For now, I’m offloading the custom properties by building another React app and embedding it as an iframe. Alternatively, I might just need to create a custom front end.

This is my postgre
| id                                   | threadId                            | type    | url                                                                                     | chainlitKey                           | name         | display | objectKey                                                   | size | page | language | forId                              | mime              | props                                                                                                                 |
|--------------------------------------|-------------------------------------|---------|-----------------------------------------------------------------------------------------|---------------------------------------|--------------|---------|-----------------------------------------------------------|------|------|----------|------------------------------------|-------------------|-----------------------------------------------------------------------------------------------------------------------|
[...]

PR #1949, opened 2025-03-03
fix: Include element `props` in SQLAlchemyDataLayer query for elements

ckrapu-nv:
# Description
Fixes an issue arising with `SQLAlchemyDataLayer` in which a custom element's state would not be properly reloaded after resuming a chat thread due to a null value for the element's `props`. https://github.com/Chainlit/chainlit/issues/1799 documents the issue in more detail, including a fix.

# Problem
In `sql_alchemy.py`, the SQL query contained in `elements_query` is missing the `props`. Since this is missing, any reloaded custom element will have an empty value for `props` thereby ignoring its latest state.

Also, there was an error on insertion of new elements into the database in an identical fashion to https://github.com/Chainlit/chainlit/pull/1794 which fixes the same error in the default Postgres data layer. This PR includes the same fix for the SQLAlchemyDataLayer which also blocked a successful element refresh workflow.

# Steps to Reproduce
- Create an application with a custom Element
- Use the SQLAlchemyDataLayer for persistence
- Create a Chainlit app instantiating the custom Element
- Update the state of the element
- Refresh the page or create a new conversation, then switch back to the original one

# Solution and Changes
Add the `props` field to the query and use `props=element.get("props", "{}")` instead of `props=json.loads(element.get("props", "{}"))`

# Testing
- Ran `pytest` from `backend/` with no failures or errors and all tests finishing.
- Successfully tested with my application which uses the S3 storage client and SQLAlchemyDataLayer with Azure PostgreSQL (flexible server). I created two new chat threads, triggered the custom element creation, and created a third chat. I then switched back to the two earlier chat threads and verified that the elements' props were loaded and the element state was properly set.

### DIFF
--- a/backend/chainlit/data/sql_alchemy.py
+++ b/backend/chainlit/data/sql_alchemy.py
@@ -519,7 +519,12 @@ class SQLAlchemyDataLayer(BaseDataLayer):
 
         columns = ", ".join(f'"{column}"' for column in element_dict_cleaned.keys())
         placeholders = ", ".join(f":{column}" for column in element_dict_cleaned.keys())
-        query = f"INSERT INTO elements ({columns}) VALUES ({placeholders})"
+        updates = ", ".join(
+            f'"{column}" = :{column}'
+            for column in element_dict_cleaned.keys()
+            if column != "id"
+        )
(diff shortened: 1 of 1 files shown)


willydouhard (approved):
lgtm

Issue #1892, opened 2025-02-14
WebSocket Session Affinity Broken in Istio due to Missing Cookie/Header


munday-tech:
**Issue**

Last year, a change was made to remove `X-Chainlit-Session-Id` and other `extraHeaders` from WebSocket connections (see PR: [#1575](https://github.com/Chainlit/chainlit/pull/1575)). The intention was to move towards cookie-based session affinity, but no cookie is currently being set to maintain this affinity.

**Impact**

For Istio-managed Kubernetes clusters, sticky sessions are broken because:

- Istio relies on consistent hashing (ConsistentHashLB) to provide session affinity. [Docs](https://istio.io/latest/docs/reference/config/networking/destination-rule/#LoadBalancerSettings-ConsistentHashLB)
- This requires either a header or a cookie to consistently route WebSocket requests to the same pod.
- Without a header/cookie, subsequent WebSocket connections get randomly load-balanced, leading to session disruptions.

**Expected Behavior**

- A session cookie (e.g., X-Chainlit-Session-Id) should be set if headers are removed.

This would allow Istio to hash on the cookie and properly maintain sticky sessions.

willydouhard:
My understanding is that is it up to the load balancer to set/read that cookie and not the application

munday-tech:
That applies if you're relying on an external load balancer for sessions. However, in our EKS deployment, Istio manages internal traffic instead of the ALB handling sticky sessions.

This basically means with the removal of headers and no addition of a session cookie, Chainlit just simply won't work in multi-pod deployments.

munday-tech:
Hey @willydouhard any plan to make this accessible?

PR #1953, opened 2025-03-04
feat: add cookie-based session affinity for Istio

munday-tech:
Last year a change was made to remove `X-Chainlit-Session-id` header from websocket connections, causing kubernetes clusters using istio to lose the ability to maintain sticky sessions. Discussed here: https://github.com/Chainlit/chainlit/issues/1892

This PR adds new functionality to allow users to enable the addition of the users session id as a cookie, that istio can use for sticky sessions. This simply sets a cookie on the client-side when a websocket connection occurs.

Changes:
- Added `enable_session_cookie` to enable this option under project settings. Defaults to off.
- Added a FastAPI endpoint to set session cookie
- Added websocket command to trigger the above API


willydouhard:
This would be useful for everybody, I don't think we need to make it opt-in.
Also, instead of relying on an extra websocket message to get the cookie, we could make the call directly [here](https://github.com/Chainlit/chainlit/blob/main/libs/react-client/src/useChatSession.ts#L108)

backend/chainlit/server.py
```diff
@@ -302,6 +299,25 @@ async def teams_endpoint(req: Request):
 # -------------------------------------------------------------------------------
 
 
+@app.get("/set-session-cookie")
+async def set_session_cookie(session_id: str, response: Response, request: Request):
```
willydouhard, line 303: an auth check would be welcomed here, example available [here](set_auth_cookie)

willydouhard (changes requested):

willydouhard:
Are you willing to implement the changes @munday-tech? I think the whole community would benefit from this

munday-tech:
Can do it this week @willydouhard.

backend/chainlit/server.py
munday-tech: This link is dead :(

willydouhard: here you go https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/server.py#L1243

munday-tech:
@willydouhard Made the changes.

1. Removed config option, happens every websocket connect now
2. Updated the set cookie to `POST` from `GET`
3. Added in the auth check as requested
4. Added in test cases

The `POST` happens in the `.connect()` of the websocket as it has to occur after the websocket has been created (So we can fetch session id and map to user for auth)

My ts/react isn't great so let me know if anything needs to be improved :)

munday-tech:
@willydouhard any plans for when we can review / get this one in :)?

willydouhard (approved):
lgtm

willydouhard:
I will merge this and make two small modifications:

1. Use the chainlit api client to make the call
2. Move the set cookie before the connection and not once the connection is successful (since the session is created on the server during the connection and we need stickiness right from there)

PR #1956, opened 2025-03-04
fix: Fix Element mime type by fixing from_dict()

### LINKED ISSUE #1831: Mime type for different file formats merged (How to read?)
In Previous versions every file type had its own mime type i.e (application/*), but as of 2.0.603 release, Docx/Pptx/Xlsx are just given mime type "**file**" how can i read these files now accordingly?

hayescode:
Fixes #1831 

Element type "file" was being used for both `type` and `mime`
image

After changing mime to "mime" we get the intended results
image


willydouhard (approved):
lgtm

PR #1957, opened 2025-03-05
fix: When upload fail server side the error message was undefined front

pmercier:
This patch allow to display the server error message correctly in the toast.


willydouhard (approved):
LGTM

PR #1963, opened 2025-03-06
Introducing the Semantic Kernel Filter

eavanvalkenburg:
This PR adds a Semantic Kernel Filter class to Chainlit.

This allows a developer using Chainlit with Semantic Kernel to easily add function calling steps to their app.

All you need to do is (assuming you have a `kernel` variable of type semantic_kernel.Kernel):
```python
from chainlit.semantic_kernel import SemanticKernelFilter

filter = SemanticKernelFilter()
filter.add_to_kernel(kernel)
#or
SemanticKernelFilter(kernel=kernel)

```


willydouhard:
Thank you! We should expose this in the `__init__.py` so that everyone can import it from the root. However we should lazy load it since we don't want to load the semantic kernel package unless it is actually used. Here is an example for other third parties

https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/__init__.py#L82
https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/__init__.py#L111
https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/__init__.py#L141

eavanvalkenburg:
@willydouhard updated

willydouhard (approved):
LGTM, we should add an example in the docs!

PR #1967, opened 2025-03-07
fix: Sort thread history groups in fixed order

goda6565:
This PR fixes the inconsistent order of thread history groups in the left sidebar. The groups (Today, Yesterday, Previous 7 days, Previous 30 days) were stored as object keys, and JavaScript does not guarantee their order. I added explicit sorting using a fixed order array so that the groups always appear in the intended order.

### DIFF
--- a/frontend/src/components/LeftSidebar/ThreadList.tsx
+++ b/frontend/src/components/LeftSidebar/ThreadList.tsx
@@ -1,6 +1,6 @@
 import { cn } from '@/lib/utils';
 import { size } from 'lodash';
-import { useContext, useState } from 'react';
+import { useContext, useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Link, useNavigate } from 'react-router-dom';
 import { useSetRecoilState } from 'recoil';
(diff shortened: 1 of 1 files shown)


frontend/src/components/LeftSidebar/ThreadList.tsx
```diff
@@ -260,51 +260,73 @@ export function ThreadList({
...
-                  idToResume === thread.id && !threadHistory.currentThreadId;
-                const isSelected =
-                  isResumed || threadHistory.currentThreadId === thread.id;
-                return (
-                  <SidebarMenuItem key={thread.id} id={`thread-${thread.id}`}>
-                    <Link to={isResumed ? '' : `/thread/${thread.id}`}>
-                      <SidebarMenuButton
-                        isActive={isSelected}
-                        className="relative truncate h-9 group/thread"
+      {Object.keys(threadHistory.timeGroupedThreads ?? {})
```
willydouhard, line 263: we could use `useMemo` here

willydouhard (changes requested):
LGTM overall, curious to know in which scenario the keys were not sorted correctly naturally

goda6565:
It’s difficult to reproduce intentionally, but I’ve seen instances—and others have too—where the keys ended up in a different order than expected. This isn’t something we can easily trigger on demand, but those sporadic cases were enough to warrant adding explicit sorting for consistency.

willydouhard (approved):
lgtm

PR #1974, opened 2025-03-09
feat: send user message from custom element

willydouhard:


SAIVENKATARAJU:
@willydouhard  is this live yet? I need customelement in AskUserMessage. where I am using dropdown to select values from user

PR #1975, opened 2025-03-09
Improve smooth scroll behavior to prevent button flickering

hexart:
# Smooth Scroll Button Behavior Fix

This PR addresses the issue where the scroll button briefly reappears during smooth scrolling animation.

## Problem
When using `behavior: 'smooth'` for scrolling, the scroll button would flicker due to `handleScroll` being triggered during the animation process, causing `showScrollButton` to be set to `true` before reaching the bottom.

## Solution

1. Added an `isScrolling` state to track scroll animation status
2. Created a `checkScrollEnd` function at component level to monitor actual scroll position and detect when scrolling has completely stopped
   - Placing this function at component level improves performance by avoiding function recreation on each button click
4. Enhanced both  `scrollToBottom`   and  `scrollToPosition`   functions to properly handle smooth scrolling:
   - Setting `isScrolling` state to `true` at start
   - Hiding scroll button during animation
   - Using `checkScrollEnd` to detect end of animation
   - Updating button visibility based on final position

## Edge Case Handling
The solution also handles user interaction during scroll animation. When users interrupt auto-scroll with mouse wheel, the scroll button visibility is immediately updated when scrolling stops, accurately reflecting the current position without requiring additional scroll input.


willydouhard (approved):
lgtm

PR #1979, opened 2025-03-10
Enhance CORS configuration consistency between Socket.IO and main application

hexart:
## Background
While investigating a CORS-related issue after user logout, I discovered that Socket.IO and the main application use inconsistent CORS configurations:
```python
# Socket.IO uses an empty list
sio = socketio.AsyncServer(cors_allowed_origins=[], async_mode="asgi")

# Main application uses the config value
app.add_middleware(
    CORSMiddleware,
    allow_origins=config.project.allow_origins,
    allow_credentials=True,
    allow_methods=["*"],
    allow_headers=["*"],
)
``` 

## Issue Context
In development environments with frontend and backend running on different ports (e.g., frontend on port 5173, backend on port 8000), CORS errors occur after logout:
```
Cannot use wildcard in Access-Control-Allow-Origin when credentials flag is true.
Fetch API cannot load http://localhost:8000/auth/config due to access control checks.
``` 
Root cause analysis revealed two important findings:

1. When using `credentials: 'include'` in frontend requests, CORS specifications require that `Access-Control-Allow-Origin` must specify exact origins rather than wildcards (`*`)
2. Both the main application and Socket.IO need consistent CORS configurations

## Solution
For the first point, users should update their .chainlit/config.toml file to use specific origins:
```toml
[project]
# Instead of this:
allow_origins = ["*"]

# Use specific origins:
allow_origins = ["http://localhost:5173", "http://localhost:8000"]
``` 

This PR addresses the second point by aligning Socket.IO's CORS configuration with the main application:
```python
# Modified code
sio = socketio.AsyncServer(cors_allowed_origins=config.project.allow_origins, async_mode="asgi")
``` 
## Benefits

1. **Configuration consistency** - Socket.IO server automatically adopts the same CORS settings as the main application
2. **Improved developer experience** - Changes to allow_origins in config.toml apply universally
3. **Better cross-environment compatibility** - Facilitates deployment in various environments with different domain configurations
4. **Future-proofing** - As more Socket.IO-based features are added (like MCP functionality), CORS handling remains consistent

## Testing
This enhancement has been tested in a development environment with frontend on port 5173 and backend on port 8000, confirming that with proper configuration, authentication flows work correctly before and after logout.

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -195,7 +195,9 @@ copilot_build_dir = get_build_dir(os.path.join("libs", "copilot"), "copilot")
 
 app = FastAPI(lifespan=lifespan)
 
-sio = socketio.AsyncServer(cors_allowed_origins=[], async_mode="asgi")
+sio = socketio.AsyncServer(
+    cors_allowed_origins=config.project.allow_origins, async_mode="asgi"
+)
 
 asgi_app = socketio.ASGIApp(
     socketio_server=sio,


willydouhard:
It is expected that the developer should configure the allowed origins in the `config.toml`. As for the websocket cors, my understanding is that it inherits the main application cors and that we should not configure them, otherwise they are going to be "doubled" and cause issue.

See https://github.com/miguelgrinberg/python-socketio/discussions/1304

hexart:
> It is expected that the developer should configure the allowed origins in the `config.toml`. As for the websocket cors, my understanding is that it inherits the main application cors and that we should not configure them, otherwise they are going to be "doubled" and cause issue.
> 
> See [miguelgrinberg/python-socketio#1304](https://github.com/miguelgrinberg/python-socketio/discussions/1304)

Thank you for the clarification! I wasn't aware that Socket.IO would inherit the main application's CORS configuration.
Based on your explanation and the linked discussion, it seems the empty list (`cors_allowed_origins=[]`) is intentional, telling Socket.IO to inherit the FastAPI CORS settings. This makes sense and explains why my solution worked, but perhaps for the wrong reason.
The real issue was indeed in the `config.toml` configuration, where `allow_origins = ["*"]` needed to be changed to specific origins like `allow_origins = ["http://localhost:5173"]` when using credentialed requests.
However, I'm still curious about one thing: why didn't we need to set specific origins in the `config.toml` during development before, but after the MCP update, it became necessary? Was there a change in how credentials are handled in recent updates?
I'll close this PR since no code change is needed. Thank you for the explanation - this helps me better understand how CORS works with Socket.IO in Chainlit.

PR #1989, opened 2025-03-12
feat(backend/mcp): Support set env vars from command line

tcx4c70:
Some MCP servers need env vars to set API keys, configure, etc. This commit will add support to set env vars from command line.


willydouhard (approved):
LGTM

PR #2009, opened 2025-03-16
fix: chat_profiles model description

shinishiho:
Fix an issue where model description (in Chat Profiles) does not disappear when quickly hovered

## Screenshot
Screenshot 2025-03-16 at 2 35 39 PM


willydouhard (approved):
lgtm

PR #2014, opened 2025-03-17
feat(mcp): Support custom cmd allow list for mcp server

tcx4c70:
Some mcp servers are implemented using compiled language (such as go), and can't run via `npx` or `uvx`. This PR supports to custom allow list for the commands to run mcp servers.

### DIFF
--- a/backend/chainlit/config.py
+++ b/backend/chainlit/config.py
@@ -116,6 +116,11 @@ edit_message = true
     # Sample rate of the audio
     sample_rate = 24000
 
+[features.mcp]
+    # Only the executables in the allow list can be used for MCP stdio server.
+    # Only need the base name of the executable, e.g. "npx", not "/usr/bin/npx".
+    stdio_executable_allow_list = [ "npx", "uvx" ]
+
 [UI]
 # Name of the assistant.
 name = "Assistant"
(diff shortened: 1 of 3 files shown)


backend/chainlit/server.py
```diff
@@ -762,7 +762,7 @@ async def project_settings(
         config.features.audio.enabled = True
 
     if config.code.on_mcp_connect:
-        config.features.mcp = True
+        config.features.mcp.enabled = True
```
willydouhard, line 765: we could rely on the config value instead of the decorator presence

backend/chainlit/mcp.py
```diff
@@ -47,22 +50,24 @@ def validate_mcp_command(command_string: str):
     # Look for the actual executable in the command
     executable = None
     executable_index = None
+    allowed_executables = config.features.mcp.stdio_executable_allow_list or [
```
willydouhard, line 53: let's make those the default config values (in the config generation) and not hardcode them

backend/chainlit/config.py
```diff
@@ -213,11 +218,17 @@ class AudioFeature(DataClassJsonMixin):
     enabled: bool = False
 
 
+@dataclass
+class MCPFeature(DataClassJsonMixin):
+    enabled: bool = False
+    stdio_executable_allow_list: Optional[list[str]] = None
```
willydouhard, line 224: i would love to add transport config here. default would be ["stdio", "sse"], that way developers could refuse to use stdio servers for security reasons

willydouhard (changes requested):
this is cool stuff, few change to make it more coherent and complete

backend/chainlit/server.py
```diff
@@ -1109,12 +1106,24 @@ async def connect_mcp(
             exit_stack = AsyncExitStack()
 
             if payload.clientType == "sse":
+                if not config.features.mcp.sse.enabled:
```
willydouhard, line 1109: I think we are missing a check on `config.features.mcp.enabled`. It seems like it is never checked and it could be false and sse could be enabled. in that case we should have failed earlier

willydouhard (changes requested):
LGTM overall, last detail

PR #2017, opened 2025-03-18
Fix: Resolve WebSocket connection error when deleting chat history

hexart:
## Issue
When deleting a chat history session from the sidebar, a WebSocket error occurs:
"WebSocket is closed before the connection is established."

## Solution
Added a short delay (300ms) after clearing the current session before proceeding with the delete operation. This gives the WebSocket enough time to properly reconnect before continuing with the session deletion process.


willydouhard (approved):
lgtm

PR #2026, opened 2025-03-19
Added tags tracing on LangChain tracer

LoreDema:
Tags are not traced on LangChain Tracer due to this missing line. I added it so that they can also be traced.


willydouhard (approved):
lgtm

PR #2030, opened 2025-03-20
Fix Plotly width in #1919

### LINKED ISSUE #1919: Plotly Element fixed at 600px width?
I want to display Plotly figure inline using `cl.Plotly` but I cannot get it to render with width greater than 600px. Then I saw that frontend [InlinedPlotlyList.tsx](https://github.com/Chainlit/chainlit/blob/main/frontend/src/components/chat/Messages/Message/Content/InlinedElements/InlinedPlotlyList.tsx) is hardcoding the `maxWidth` of parent container to `600px`.

Is this intentional or a bug?

```js
const InlinedPlotlyList = ({ items }: Props) => (
  <div className="flex flex-col gap-2">
    {items.map((element, i) => {
      return (
        <div
          key={i}
          className="max-w-[600px] h-[400px]"
          style={{
            maxWidth: '600px',
            height: '400px'
          }}
        >
          <PlotlyElement element={element} />
        </div>
      );
    })}
  </div>
);
```

hayescode:
Fixes #1919

### DIFF
--- a/frontend/src/components/chat/Messages/Message/Content/InlinedElements/InlinedPlotlyList.tsx
+++ b/frontend/src/components/chat/Messages/Message/Content/InlinedElements/InlinedPlotlyList.tsx
@@ -14,7 +14,7 @@ const InlinedPlotlyList = ({ items }: Props) => (
           key={i}
           className="max-w-[600px] h-[400px]"
           style={{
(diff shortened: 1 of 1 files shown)


willydouhard (approved):
lgtm

PR #2033, opened 2025-03-20
[Feature] Add Azure AD Group Memberships to User Metadata

parthbs:
This PR updates the oauth_providers.py file to include Azure Active Directory (Azure AD) group memberships in the user metadata. 

The following changes were made:

1. Fetch Group Memberships: Added logic to retrieve group memberships using the Microsoft Graph API.
2. Error Handling: Implemented a try-except block to gracefully handle any errors during the group membership fetch process.
3. User Metadata Update: Updated the User object to include the fetched group memberships under the groups key in the metadata.

**Testing:**
- Verified that group memberships are correctly fetched and added to the user metadata.
- Tested error handling to ensure the application continues to function if the group membership fetch fails.

**Dependencies:**

- No new dependencies were added.

**Notes:**

- This change ensures that user group memberships are available for downstream processes that rely on user metadata.
- Backward compatibility is maintained as errors during group membership fetch are ignored.

### DIFF
--- a/backend/chainlit/data/storage_clients/azure_blob.py
+++ b/backend/chainlit/data/storage_clients/azure_blob.py
@@ -1,8 +1,14 @@
 from datetime import datetime, timedelta, timezone
 from typing import Any, Dict, Union
 
-from azure.storage.blob import BlobSasPermissions, ContentSettings, generate_blob_sas
-from azure.storage.blob.aio import BlobServiceClient as AsyncBlobServiceClient
+from azure.storage.blob import (  # type: ignore
+    BlobSasPermissions,
+    ContentSettings,
+    generate_blob_sas,
+)
+from azure.storage.blob.aio import (  # type: ignore
+    BlobServiceClient as AsyncBlobServiceClient,
+)
 
 from chainlit.data.storage_clients.base import EXPIRY_TIME, BaseStorageClient
 from chainlit.logger import logger
--- a/backend/chainlit/oauth_providers.py
+++ b/backend/chainlit/oauth_providers.py
@@ -239,12 +239,31 @@ class AzureADOAuthProvider(OAuthProvider):
                 # Ignore errors getting the photo
                 pass
 
+            try:
+                # Fetch group memberships
+                groups_response = await client.get(
+                    "https://graph.microsoft.com/v1.0/me/memberOf",
+                    headers={"Authorization": f"Bearer {token}"},
+                )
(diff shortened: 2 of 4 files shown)


hayescode:
@parthbs can we put this under a Boolean parameter in @cl.login? My company has 10s of thousands of groups for some reason and this could blow up the metadata for some.

package.json
```diff
@@ -24,7 +24,7 @@
     "lint": "pnpm run lintUi && pnpm run lintPython",
     "lintUi": "pnpm run --parallel lint",
     "formatUi": "pnpm run --parallel format",
-    "lintPython": "cd backend && poetry run dmypy run --timeout 600 -- chainlit/ tests/",
```
willydouhard, line 27: why remove the timeout?

willydouhard (commented):
lgtm overall

backend/chainlit/data/storage_clients/azure_blob.py
```diff
@@ -1,8 +1,14 @@
 from datetime import datetime, timedelta, timezone
 from typing import Any, Dict, Union
 
-from azure.storage.blob import BlobSasPermissions, ContentSettings, generate_blob_sas
-from azure.storage.blob.aio import BlobServiceClient as AsyncBlobServiceClient
+from azure.storage.blob import (  # type: ignore
```
willydouhard, line 4: why is this needed?

parthbs: Not supposed to be there. Removed accidental lint ignore.

package.json
parthbs: accidental changes, reverted.

parthbs:
@willydouhard can you check this PR now?

backend/chainlit/oauth_providers.py
```diff
@@ -239,12 +239,31 @@ async def get_user_info(self, token: str):
                 # Ignore errors getting the photo
                 pass
```
willydouhard, line 241:
```suggestion
user_groups = []
```
If the call fails, user_groups will not be defined

willydouhard (changes requested):

willydouhard:
I see that you initialise a `user_data` variable but I was talking about `user_group`

hayescode:
@parthbs you can retrieve group memberships using the token this already retrieves. No need to force this into the framework when most users won't need this.

Issue #1756, opened 2025-01-20
Azure Blob Storage issues with SQLAlchemyDataLayer


GhimBoon:
**Describe the bug**
get_read_url is not called in all other data layers (like SQLAlchemyDataLayer) hence AzureBlob doesn't work for them.

**To Reproduce**

1. Enable Azure Blob Data Storage Client
2. Upload to Chainlit
3. Upload will fail to display as URL field is empty

**Expected behavior**
URL field will be populated with SAS and image will be displayed.

**Screenshots**
N/A

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
N/A

**Additional context**
Chainlit 2.0.4

desaxce:
You mention SAS token. Do you use `AzureStorageClient` or `AzureBlobStorageClient`?

I expect the issue to be that `AzureBlobStorageClient.upload_file` does not return a `url` key in its output. 

An issue with `get_read_url` not being called by `SQLAlchemyDataLayer` would only arise upon resuming a thread, when `BaseDataLayer.get_thread` gets triggered.

GhimBoon:
`AzureBlobStorageClient` is the one causing the issue. Yes, it only arises when resuming a thread as the url is not existent.

GhimBoon:
`AzureBlobStorageClient` is also returning a non standard dictionary key(path) rather than (object_key) as compared to all the other implementations.

desaxce:
I can update `AzureBlobStorageClient.upload_file` to align and return a dict with `object_key` and `url` keys. 
That would work with `SQLAlchemyDataLayer` because its `create_element` implementation stores the URL returned by `upload_file`. 

However, storing the `upload_file` URL means it's a long-living URL, which lets anyone with access to it upload to that location. 
I recommend using the `url` column to only store the original URL link (when the user opts for that kind of attachment).

On thread resume, instead of relying on the persisted URL (the one coming from `upload_file` for `SQLAlchemyDataLayer`), the data layer should ask the storage client for a read-only URL which expires in 1 hour. The resumed thread uses that URL to display elements. 

Note that `SQLAlchemyDataLayer` currently fetches *all* threads when resuming a thread. Something to refactor along the way.

PR #2035, opened 2025-03-21
Fix: Images disappear after page reload when using AzureBlobStorageClient

shuntagami:
fix https://github.com/Chainlit/chainlit/issues/1756

## What
This PR updates the Azure Blob Storage implementation of `upload_file` to return a consistent response format that includes both `object_key` and `url` keys, aligning with other storage providers like S3 and GCS.

The current implementation returns:
```python
{
    "path": object_key,
    "size": properties.size,
    # other properties but no url or object_key
}
```

While the S3 implementation returns:
```python
{
    "object_key": object_key,
    "url": url
}
```

## Why
Currently, when using `SQLAlchemyDataLayer` with Azure Blob Storage, uploaded images disappear after page reload. This is because:

1. The Element data model expects either a `url` or `chainlit_key` to display media
2. Frontend components check for `element.url` first when rendering elements
3. When saving to the database, `SQLAlchemyDataLayer` tries to get the URL from `uploaded_file.get("url")`, which is `None` for Azure Blob Storage

This inconsistency causes files to appear initially (using temporary URLs) but disappear after reload since the persisted element lacks a valid URL.

Note: As mentioned in the related issue, in future releases it might be better to further improve this by generating temporary read-only URLs that expire after a certain period rather than storing long-lived URLs directly.

### DIFF
--- a/backend/chainlit/data/storage_clients/azure_blob.py
+++ b/backend/chainlit/data/storage_clients/azure_blob.py
@@ -70,6 +70,8 @@ class AzureBlobStorageClient(BaseStorageClient):
 
             return {
                 "path": object_key,
+                "object_key": object_key,
+                "url": self.get_read_url(object_key),
                 "size": properties.size,
                 "last_modified": properties.last_modified,
                 "etag": properties.etag,
--- a/backend/chainlit/data/storage_clients/azure_blob.py
+++ b/backend/chainlit/data/storage_clients/azure_blob.py
@@ -71,7 +71,7 @@ class AzureBlobStorageClient(BaseStorageClient):
             return {
                 "path": object_key,
                 "object_key": object_key,
-                "url": self.get_read_url(object_key),
+                "url": await self.get_read_url(object_key),
                 "size": properties.size,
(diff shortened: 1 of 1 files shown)


shuntagami:
@willydouhard 
Thanks for your contributions, I love this project❤️
so, review please !

backend/chainlit/data/storage_clients/azure_blob.py
```diff
@@ -70,6 +70,8 @@ async def upload_file(
 
             return {
                 "path": object_key,
+                "object_key": object_key,
+                "url": self.get_read_url(object_key),
```
willydouhard, line 74: `get_read_url` returns a coroutine, you need to await it

willydouhard (changes requested):

backend/chainlit/data/storage_clients/azure_blob.py
shuntagami: @willydouhard Thank, I fixed it!

willydouhard (approved):
LGTM

PR #2037, opened 2025-03-21
feat: Add config options to load avatar and logo from CDN URLs

hasitparmar-snyk:
This PR adds two new configurable fields—`logo_file_url` and `default_avatar_file_url`—in `config.toml`, allowing media to be loaded directly from CDN URLs.

#### ✨ What's New?
- 🆕 New optional configuration: `logo_file_url` & `default_avatar_file_url`
- Updated **Logo** component to load the logo from `logo_file_url` (if present).
- Updated **Avatar** component to load the avatar image from `default_avatar_file_url` (if present).
- **Fallback mechanism:** If these fields aren’t configured, the existing logic executes and fetches media from the Chainlit backend as usual.


willydouhard:
I understand, the issue with the suggested implementation is that it only works for the avatar using the default author name (specified in the config). Maybe we can add an avatar subsection in the UI section of the config where you can specify an url for each avatar name?

willydouhard:
Sorry for the confusion. If I read the PR correctly, setting `avatar_file_url` will return the same avatar regardless of the author name. The current behavior is that the avatar is loaded given the author name (and defaults to the favicon). Such a mapping could be done in the config.toml (or somewhere else).

Unless you want to make the default configurable?

willydouhard:
In that case we can just keep it as is but add default in the name like default_avatar_file_url

willydouhard (approved):
lgtm

PR #2038, opened 2025-03-21
feat: support additionalQueryParamsForAPI setting in copilot

hasitparmar-snyk:
This PR introduces a new configuration option: `additionalQueryParamsForAPI` in Copilot settings. This allows passing extra query parameters in API requests to the Chainlit server.

### ✨ What's New?
- 🆕 **New copilot setting:** `additionalQueryParamsForAPI`, which accepts key-value pairs.
- 🔗 **Automatic query param injection:** When configured, all API requests from the Copilot client will include these parameters.
- 📌 **Example:** If set as `{ "version": "v3", "region": "europe" }`, API calls will be formatted as:
`/auth/config?version=v3&region=europe`
`/project/settings?version=v3&region=europe`

### Why This Matters?
- When Chainlit backend server is behind the API Gateway which enforce some strict API versioning rules, then this configurable option is savior.
- Improves request tracking and API monitoring. For example, if multiple websites embed Chainlit’s Copilot while using a single Chainlit backend, this feature allows better tracking of requests from individual sites, offering more flexibility.

### Implementation changes:
- Added `additionalQueryParamsForAPI` option in Copilot settings to accept key-value pairs.
- Updated `buildEndpoint` method in `APIBase` (React client) to append these parameters to the API URL.

### DIFF
--- a/frontend/src/api/index.ts
+++ b/frontend/src/api/index.ts
@@ -25,6 +25,7 @@ const onError = (error: ClientError) => {
 export const apiClient = new ChainlitAPI(
   httpEndpoint,
   'webapp',
+  {}, // Optional - additionalQueryParams property.
   on401,
   onError
 );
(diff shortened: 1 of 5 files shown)


willydouhard (approved):
lgtm

PR #2040, opened 2025-03-22
fix: Use await instead of asyncio.create_task

tcx4c70:
For almost all persistence operations, we use `asyncio.create_task` to do async operations. However, `asyncio.create_task` only create a task, it won't wait the operation to complete. It might cause that messages/steps don't persistent in db even if `await msg.send()` returns.


willydouhard:
I agree but it will add latency for the end user right? This will result to a higher latency and maybe a poor UX?

Issue #2022, opened 2025-03-19
Add support for headers for MCP SSE Server


meabhishekkumar:
Scenario : MCP server (with SSE ) that requires additional configuration to be provided. May be through headers such as Auth header, configuration settings that cannot be provided at user prompt.

Challenge : Current interface only have option to provide "name" and "server".  

Expected Feature : MCP protocol does allow to add additional information . Even in MCP inspector they can be added as environment variables. It would be good to have the feature to support adding additional dictionary / headers.

venkatest:
This would be a good feature. We are also building mcp server which will accept azure ad token as header to provide information based on user and their rights

PR #2047, opened 2025-03-24
feat: Rework MCP settings and support headers for SSE

tcx4c70:


willydouhard:
Can you provide a description of why these changes are necessary?

GhimBoon:
Changes are necessary to support #2022.

jiriklein:
> Can you provide a description of why these changes are necessary?

These changes are valuable if you want to secure your MCP tool endpoints behind some sort of auth (with token passthrough), especially if you're building on top of existing APIs. 
That way, you can set a bearer token at the point of the `chainlit` app making requests, e.g. at user session init, and provide that token to the MCP server which can pass it through to the relevant tool endpoints. 

A good example of this is https://github.com/tadata-org/fastapi_mcp/blob/main/examples/08_auth_example_token_passthrough.py

hayescode:
@tcx4c70 this isn't working for me.
image

image

image

hayescode:
For headers i'd recommend a form with seperate key/value input areas, and a `+` sign to add a new row/header.

senj:
Will there be another approach? Imo this feature is really important.

PR #2055, opened 2025-03-25
Feature: Add custom env file loading

adh-wonolo:
Hi there,

In trying to use chainlit within our application we realized it will always load `.env` even though we have various `.env` files for our different environments such as `.env.production`. Putting up this small PR so that we can just specify the environment ourselves and hopefully this is an acceptable contribution!


willydouhard (approved):
lgtm

PR #2057, opened 2025-03-25
feat: add share button

celeriev:


celeriev:
Hi @willydouhard, I started working on this idea of a sharing thread feature, but I'd love to get your guidance. I know you've been considering this feature for a while. I'm curious if you've already thought of the best way to handle this backend-wise.

backend/chainlit/socket.py
```diff
@@ -65,6 +66,71 @@ async def resume_thread(session: WebsocketSession):
...
+    from chainlit.context import init_http_context
+    from chainlit.element import ElementDict
+    from chainlit.step import StepDict
+
+    data_layer = get_data_layer()
+    if not data_layer or not session.user or not session.thread_id_to_resume:
+        return
+
+    new_thread_id = str(uuid.uuid4())
+    init_http_context(thread_id=new_thread_id, user=user)
```
willydouhard, line 88: i don't think we need this line

frontend/src/components/header/Share.tsx
```diff
@@ -0,0 +1,134 @@
+import { hasMessage } from '@/lib/utils';
```
willydouhard, line 1: I think we should move this button in the `...` popover menu in the chat history (where rename and delete are)

willydouhard (changes requested):
Thank you for this contribution, it looks great! I added a few comments

frontend/src/components/header/Share.tsx
celeriev: Do you think we should put it in both places? I see Claude, OpenAI have it on the top right as well

willydouhard: we can do both but I prefer to use this [icon](https://lucide.dev/icons/share) in a button then like chat gpt

celeriev:
hello! I was in offline, but I'm taking this back this week

sanchezedgar:
Excited about this!

Internally, we've implemented a basic approach for sharing threads using a `?thread_id` query param in the `start_chat` handler. One feature we've appreciated is the ability to include a "header" and "footer" message for shared threads. The header thread includes the original author info. Would this solution support something similar? For instance, could we check if a thread is shared in an `on_chat_resume` handler and append/prepend header/footer messages accordingly? I reviewed the PR but am not yet familiar enough with the Chainlit source to determine if this is possible.

Screenshot 2025-04-23 at 5 20 03 PM

celeriev:
Thanks for the feedback @sanchezedgar , I believe we could, yes, I like the header Idea. Let me start by tackling the first blocking points for the PR and we can work on this as well

mkay1375:
Thanks for the PR! I have a couple of questions:
1. Do you have an estimate for when this will be completed?
2. Would it be possible to copy the shared chat to another (possibly temporary) table at the moment it’s shared, and then create a new thread from that when someone opens the share link? That way, any future updates by the original user wouldn’t be visible to others viewing the shared link.

backend/chainlit/socket.py
celeriev: I cannot make it work without actually. I may be missing something

celeriev:
@willydouhard I wonder if it doesn't raise more complexity actually since one user could share a thread using a specific profile and the end user loading this shared thread might not

claude89757:
Hi, thanks for your PR ❤️

Really looking forward to this feature!

hayescode:
@celeriev thanks for the work on this much needed PR. Is this good to go from your perspective? If so please mark it ready for review.

hayescode:
@celeriev if you're up for it I think we're ready for this. This is a highly desirable feature and you e done a great job with it. Do you plan on finishing this?

celeriev:
hello @hayescode , unfortunately, I believe it is not yet fully tested, but I would welcome anyone to take it from there. My main concern would be the way this feature would act with active chat profiles, in case the receiver of the shared thread doesn't have access to the chat profile used by the user owning the shared thread.

Luttik:
I would love to see this feature in production!

Luttik:
@celeriev I think since there is still an unresolved change request (even though you left a comment), this PR might be out of focus with the core team... even though it's a very cool and wanted feature. Maybe if you re-request the review or revoke the suggestion we can get this moving again.

pindaroso:
bump!

asvishnyakov:
Superseded by #2489

Issue #2064, opened 2025-03-26
Update Contribution guidelines?


asvishnyakov:
I followed contribution guidelines to install dependencies, build, test & create PR, but had a few issues:

1. Pre-commit checks fails with the following error:

![Console log](https://github.com/user-attachments/assets/7d6417e5-e736-4a8e-ac08-fba1e353651d)

2. `poetry build` produces wheel without `frontend` and `copilot` folders inside, but with the following log:

![Console log](https://github.com/user-attachments/assets/43e216d0-ce6c-4728-85a4-98aca741546c)

Ubuntu 24.04
Python 3.12.3
Poetry 2.0.1
pnpm 9.13.2 (selected automatically by poetry and/or corepack)

asvishnyakov:
May anyone look at it, please? May be suggest which changes I need to make and I'll contribute to contribution guidelines after that :)

asvishnyakov:
@willydouhard Based on this https://github.com/Chainlit/chainlit/pull/2093#issuecomment-2782728274 - do you use ruff now instead of black? Seems like the first issue caused by black is still used in `package.json`'s `lintPython`

PR #2065, opened 2025-03-26
feature: add configuration option for default sidebar state

### LINKED ISSUE #1200: Hide sidebar
I want to be able to activate data persistence for logging purposes, but I don't want resumability of previous threads. Activating a data layer also causes the app to start with the sidebar expanded and lists the previous threads. 

I would like an option to hide the sidebar, independent of data persistence.

Is this already possible? I couldn't find a way to do it. Maybe adding this as an option to the config.toml could work?

asvishnyakov:
This PR introduces configuration option for default sidebar state and by so closes #1200

P.S. Commit in this PR was made without pre-commit checks because of #2064

### DIFF
--- a/backend/chainlit/config.py
+++ b/backend/chainlit/config.py
@@ -134,6 +134,8 @@ name = "Assistant"
 
 # layout = "wide"
 
+# default_sidebar_state = "open"
+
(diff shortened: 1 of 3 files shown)


willydouhard (approved):
lgtm

PR #2068, opened 2025-03-27
fix: flickering on app start

asvishnyakov:
This PR:
- Adds a loader that is used until authentication information is ready. This prevents flickering on app load when the initial UI (for an unauthenticated user) is displayed first and then replaced with the authenticated user's version.
- Removes the blinking cursor on the `on_chat_start` message loading (the large black dot below the chat input).

Before:
https://github.com/user-attachments/assets/f4cd157a-0620-400e-94d1-08703d328373

After:
https://github.com/user-attachments/assets/16108af1-5aa9-44b8-b304-8f54e151fce6

### DIFF
--- a/frontend/src/components/chat/Messages/index.tsx
+++ b/frontend/src/components/chat/Messages/index.tsx
@@ -72,7 +72,8 @@ const Messages = memo(
                     scorableRun={scorableRun}
                   />
                 ) : null}
-                {showToolCoTLoader || showHiddenCoTLoader ? (
+                {(showToolCoTLoader || showHiddenCoTLoader) &&
+                m.name !== 'on_chat_start' ? (
                   <BlinkingCursor />
                 ) : null}
               </React.Fragment>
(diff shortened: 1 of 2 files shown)


frontend/src/components/chat/Messages/index.tsx
```diff
@@ -72,7 +72,8 @@ const Messages = memo(
                     scorableRun={scorableRun}
                   />
                 ) : null}
-                {showToolCoTLoader || showHiddenCoTLoader ? (
+                {(showToolCoTLoader || showHiddenCoTLoader) &&
```
willydouhard, line 75: why not showing the loader indicator while on chat start runs?

asvishnyakov: It show a large black blinking dot below chat at the start:
image
This code removes it

It may indicate bug somewhere else, for example, if you filter messages, because when you start chatting `on_chat_start` disappear from message list, but at least this code fix one specific issue with UI without breaking anything else

willydouhard (approved):
lgtm

PR #2086, opened 2025-04-02
fix: GCS private key was in incorrect format

markperfect:
1. GCP creates service account keys in IAM through a JSON file. Currently, the private key is a string in the format of `-----BEGIN PRIVATE KEY-----`. However, when using this private key with the `GCS_PRIVATE_KEY` environment variable, the following error occurs:

```
storage_client = GCSStorageClient(
                     ^^^^^^^^^^^^^^^^^
  File "/Library/Python/3.11/lib/python/site-packages/chainlit/data/storage_clients/gcs.py", line 16, in __init__
    private_key = base64.b64decode(private_key).decode("utf-8")
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/base64.py", line 88, in b64decode
    return binascii.a2b_base64(s, strict_mode=validate)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
binascii.Error: Incorrect padding
```

The GCP authentication issue was caused by unnecessary encoding of the private key. This PR removes the unnecessary key encoding.

2. When using `display="side"` for elements in the side panel, the following error occurs when trying to retrieve persisted data elements:

`
Fetch API cannot load gs://{BUCKET}. URL scheme "gs" is not supported.
`

The Fetch API in browsers doesn't support the `gs://` URL natively as it expects HTTPS URLs. This ticket changes the URL from `gs://` to `https://storage.googleapis.com`.

### DIFF
--- a/backend/chainlit/data/storage_clients/gcs.py
+++ b/backend/chainlit/data/storage_clients/gcs.py
@@ -58,7 +58,7 @@ class GCSStorageClient(BaseStorageClient):
 
             return {
                 "object_key": object_key,
-                "url": f"gs://{self.bucket.name}/{object_key}",
+                "url": f"https://storage.googleapis.com/{self.bucket.name}/{object_key}",
             }
 
         except Exception as e:
--- a/backend/chainlit/data/storage_clients/gcs.py
+++ b/backend/chainlit/data/storage_clients/gcs.py
@@ -14,9 +14,6 @@ class GCSStorageClient(BaseStorageClient):
         self, project_id: str, client_email: str, private_key: str, bucket_name: str
     ):
         # Go to IAM & Admin, click on Service Accounts, and generate a new JSON key
-        private_key = base64.b64encode(private_key.encode("utf-8")).decode("utf-8")
-        private_key = base64.b64decode(private_key).decode("utf-8")
-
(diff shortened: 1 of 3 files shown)


backend/chainlit/data/storage_clients/gcs.py
```diff
@@ -13,6 +13,8 @@ class GCSStorageClient(BaseStorageClient):
     def __init__(
         self, project_id: str, client_email: str, private_key: str, bucket_name: str
     ):
+        # Go to IAM & Admin, click on Service Accounts, and generate a new JSON key
+        private_key = base64.b64encode(private_key.encode("utf-8")).decode("utf-8")
```
willydouhard, line 17: this new line should replace the one below it right?

markperfect: You actually don't need to encode at all to connect to the GCS service account. I updated it and removed the unnecessary line

willydouhard (approved):
lgtm

willydouhard:
Looks like ruff complains, can you run `ruff format` and `ruff check --fix`?

PR #2089, opened 2025-04-03
Enable CopilotFunction for main application

jdb78:
This ensures the copilot functions can also be executed in the main application. This is useful, if you want to further customize the experience.


willydouhard:
is this for custom js or custom elements? Custom elements can already access the copilot function

jdb78:
It is for custom js, so when no custom elements are rendered.

willydouhard (approved):
lgtm

PR #2091, opened 2025-04-03
feature: allow set attributes for the tags of the custom_js or custom_css file

asvishnyakov:
Case: I'm building `custom_js` used for chainlit in our project using `vite` and due to certain conditions I can produce only native ECMAScript (JS) module (`<script type="module">`) build. But currently only non-modular scripts can be used in `custom_js`

Solution: allow set attributes for the tags of the `custom_js` or `custom_css` file

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -363,9 +363,7 @@ def get_html_template(root_path):
 
     css = None
     if config.ui.custom_css:
-        css = (
-            f"""<link rel="stylesheet" type="text/css" href="{config.ui.custom_css}" {config.ui.custom_css_attributes}>"""
-        )
+        css = f"""<link rel="stylesheet" type="text/css" href="{config.ui.custom_css}" {config.ui.custom_css_attributes}>"""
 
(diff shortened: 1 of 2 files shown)


backend/chainlit/server.py
```diff
@@ -363,12 +363,10 @@ def get_html_template(root_path):
 
     css = None
     if config.ui.custom_css:
-        css = (
-            f"""<link rel="stylesheet" type="text/css" href="{config.ui.custom_css}">"""
-        )
+        css = f"""<link rel="stylesheet" type="text/css" href="{config.ui.custom_css}" {config.ui.custom_css_attributes}>"""
```
willydouhard, line 366: here is custom css attributes is None (default value) it will actually add "None" right?

willydouhard (changes requested):

backend/chainlit/server.py
asvishnyakov: Fixed, tested - now works as expected

willydouhard (approved):
lgtm

PR #2093, opened 2025-04-04
feat: add typed session accessors for user session management

zboyles:
This is a simple, non-breaking feature `SessionAccessor`, that acts as a wrapper with generic type safety, around the values stored with `user_session`. It does not modify the existing functionality except for adding a `create_accessor` helper method in the `UserSession` class. The core functionality is not touched and should lead to cleaner code, fewer runtime errors, and more advanced solutions from beginners looking to experiment with `chainlit`.

The `create_accessor` helper method provides a convenient factory pattern directly on the `user_session` object. If preferred, we can remove `create_accessor` from the `UserSession` class and simply rely on the `SessionAccessor` init to create accessor instances.

- Introduced `create_accessor` method in `UserSession` to create typed accessors for session keys with default values and optional transformation functions.
- Added `SessionAccessor` class to encapsulate session value management with type safety, including methods for getting, setting, resetting, and applying transformations to session values.

Examples
--------
```python
count = cl.user_session.create_accessor("count", 0)
count.get() # returns 0
count.set(5)  # type-safe setter
count.get() # returns 5

# With transform function
counter = cl.user_session.create_accessor("counter", 0, apply_fn=lambda x: x + 1)
counter.apply() # increments value and returns new value (1)

@cl.on_message
async def on_message(message: cl.Message):
    await cl.Message(content=f"You sent {counter.apply()} messages").send() # You sent 2 messages
```

### DIFF
--- a/backend/chainlit/user_session.py
+++ b/backend/chainlit/user_session.py
@@ -1,8 +1,10 @@
-from typing import Dict
+from typing import Dict, Generic, TypeVar, Optional, Callable
 
 from chainlit.context import WebsocketSession, context
 
 user_sessions: Dict[str, Dict] = {}
+
+T = TypeVar("T")
 
 
 class UserSession:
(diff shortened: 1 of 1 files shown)


willydouhard:
Try to run ruff format and ruff check --fix

PR #2059, opened 2025-03-25
Add HeaderLink display_name

hayescode:
Builds on #1836 by adding `display_name` optional. I didn't want to re-use `name` so that this isn't a breaking change. @willydouhard I'm no tailwinds expert so I copied the rendered className from the ReadMe over to ButtonLink. This works, but maybe isn't ideal so feel free to update this PR. Thanks!

**Without display_name (current state)**
```toml
[[UI.header_links]]
    name = "Confluence"
    icon_url = "https://www.atlassian.com/favicon.ico"
    url = "https://www.atlassian.com"

[[UI.header_links]]
    name = "Issues"
    icon_url = "https://mintlify.s3-us-west-1.amazonaws.com/chainlit-43/_generated/favicon/favicon.ico?v=3"
    url = "https://github.com/Chainlit/chainlit/issues"
```
image

**With display_name**
```toml
[[UI.header_links]]
    name = "Confluence"
    display_name = "Confluence"
    icon_url = "https://www.atlassian.com/favicon.ico"
    url = "https://www.atlassian.com"

[[UI.header_links]]
    name = "Issues"
    display_name = "Report Issue"
    icon_url = "https://mintlify.s3-us-west-1.amazonaws.com/chainlit-43/_generated/favicon/favicon.ico?v=3"
    url = "https://github.com/Chainlit/chainlit/issues"
```
image

PR #2115, opened 2025-04-16
fix: keep icon style when no display is set

fgalind1:
The PR #2059 introduced an optional display name, however if no display name is provided the style gets a bit bat with too much spacing between icons.

This PR preserves the original style when no display is specified and keeping the proposed style when display is set

Currently after #2059
image

Original before #2059 and after this PR
image


willydouhard (approved):
lgtm

Issue #2084, opened 2025-04-02
Feedback is not restored on chat resume


vio-p:
**Describe the bug**
When resuming a thread, the feedback buttons show empty even if feedback has been previously provided for a message.

**To Reproduce**
Create a simple Chainlit application with data persistence and chat resume enabled.

1. Type a message and wait for the answer. Give a feedback to that answer.
2. Create a new conversation.
3. Open the previous conversation. The message you've just given feedback to looks like no feedback has been provided for it previously.

**Expected behavior**
The feedback component should maintain the state it had before "ending" the conversation.

**Desktop:**

- OS: Windows 11
- Browser: Microsoft Edge
- Version: 134.0.3124.93

**Additional context**
It looks like even if chat resume is not enabled, the feedback component still doesn't keep its state.

PR #2130, opened 2025-04-22
fix: issue #2084 fetching steps feedbacks in get_thread

ale-delfo:
### Description  
This pull request addresses a bug in the `ChainlitDataLayer` class, specifically in the `get_thread` method. As pointed out in #2084, when resuming a thread, previous feedback entries were not being loaded from the database.

### Problem  
The `Feedback` table was not being queried at all.

### Solution  
The query used to retrieve steps from the database has been updated to also fetch related entries from the `Feedback` table. The method responsible for creating steps to be included in the thread has been updated accordingly. Now, when reloading a thread, previous feedback for each step is displayed correctly.


willydouhard (approved):
lgtm

Issue #2136, opened 2025-04-28
Cognito scope not valid


brnaba-aws:
**Describe the bug**
When using Cognito oauth provider, the default scope is not working with the default cognito setup

**To Reproduce**
Create a cognito user pool as mentioned in the AWS documentation
Set up your .env variables for cognito as mentioned in the documentation.
Start your chainlit app.
When, clicking on `continue with cognito` you would have this result:

![Image](https://github.com/user-attachments/assets/95d6c270-93e1-48c7-bada-ee4baf122259)

**Expected behavior**
I should be redirected to Cognito login

**Screenshots**

**Desktop (please complete the following information):**

- OS: MacOS
- Browser: firefox
- Version chainlit 2.5.5

**Additional context**
To fix the issue, I modified the code to include a `OAUTH_COGNITO_SCOPE=openid` and it worked.

brnaba-aws:
I dived deeper into this and notice that by changing the login page of the client application I could add profile and email.
But there could be more options.

so I suggest to add a `OAUTH_COGNITO_SCOPE` variable to let the user change those options.

dor15:
hey @brnaba-aws 
I'm also using AWS Cognito and get the same error.
we are using "cognito user pool".

I tried your solution, but it didn't work for me- 
added OAUTH_COGNITO_SCOPE to env file 
and 
self.scopes = os.environ.get("OAUTH_COGNITO_SCOPE", "openid profile email")
"scope": self.scopes 


These are my scopes 

<img width="1397" alt="Image" src="https://github.com/user-attachments/assets/29fdd352-e7e6-4e99-b5be-a116e5c39fff" />

can you tell me what is the correct OAUTH_COGNITO_SCOPE I need?

PR #2137, opened 2025-04-28
added scope as env variable to Cognito auth provider

brnaba-aws:
fixing #2136 
This PR makes sure that Cognito scopes can be set using env variable, giving end-user flexibility on the scope they want to add for their authentication.


backend/chainlit/oauth_providers.py
```diff
@@ -550,17 +550,19 @@ class AWSCognitoOAuthProvider(OAuthProvider):
         "OAUTH_COGNITO_CLIENT_ID",
         "OAUTH_COGNITO_CLIENT_SECRET",
         "OAUTH_COGNITO_DOMAIN",
+        "OAUTH_COGNITO_SCOPE",
```
willydouhard, line 553: we don't have to add it here otherwise it is going to be mandatory for the provider to be enabled

willydouhard (changes requested):

PR #2143, opened 2025-05-01
fix: Return GCS signed URL instead of direct URL

markperfect:
The Chainlit data layer stores files in Google Cloud Storage. However, an error occurs when a user attempts to access those files from a private GCS bucket. 

The issue stems from the `sync_upload_file` method, which returns the direct public URL `f"https://storage.googleapis.com/{self.bucket.name}/{object_key}"`. A direct URL will return a 403 Forbidden error for non-publicly accessible buckets.

1. This PR modifies the `sync_upload_file` method to return a signed URL instead of a direct URL. Now, the application consistently uses authenticated signed URLs throughout, which will work with private buckets
2. This PR creates unit tests for `GCS.py`, which didn't have any previously
3. Updated `CONTRIBUTING.md`


willydouhard:
The signed URL is supposed to be short lived. That's why we store the object key and sign URLs on the fly when we get a thread for instance.

hayescode (changes requested):
There seems to be a lot of unrelated changes here. Also GCS has existed for a year and private access is common so I wonder how anybody has this working before?

markperfect:
The Chainlit codebase and LLMs in general are an area of active development. A lot of work is still needed to make Chainlit robust in production, as demonstrated by the lack of comprehensive test coverage. 

Since there were only a couple of active maintainers on the repo, many features have been neglected due to prioritization and capacity. Now that we have additional maintainers, we can hopefully speed up the development cycle for the package that we collectively value.

From what I can tell, the Chainlit data layer was released 6 months ago, and there haven’t been many major contributions since then. My intuition also tells me that GCS is less likely to be used than either AWS or Azure

markperfect:
@hayescode @asvishnyakov all checks have passed

hayescode:
Can you remove all of the unrelated changes like timeout, formatting, erc

markperfect:
@hayescode removed changes to Cypress and `package.json`

hayescode (approved):

PR #2151, opened 2025-05-06
fix: crlf on paste

desaxce:
On Windows pasting multiple lines with CRLF inserts an additional new lines, [discord comment](https://discord.com/channels/1088038867602526210/1088038868344914056/1368940293835198526). 

I don't have a Windows to test, but replace full CRLF would fix the issue.


hayescode:
Fixed in #2246

PR #2152, opened 2025-05-06
mod(sql_alchemy.py): support for additional connection arguments such…

stuenofotso:
… as the schema to use for a postgres connection


stuenofotso:
mod(sql_alchemy.py): support for additional connection arguments such as the schema to use for a postgres connection

hayescode (approved):

Issue #2155, opened 2025-05-07
Makes files directory configurable


brooklynbagel:
**Is your feature request related to a problem? Please describe.**
Right now the files directory (`.files`) is hard-coded into the application and is not configurable. This can be a problem when Chainlit is being run on an environment that doesn't give the user direct control of the directory the application is installed and run in. 

**Describe the solution you'd like**
Add a configuration to `config.toml` to override the default files directory. I'd suggest something like `features.spontaneous_file_upload.files_dir`

**Describe alternatives you've considered**
Keep the status-quo which is problematic on certain environments.

**Additional context**
I have a PR that is ready for review and comment I will link to the issue.

rakeshvikhar:
this is very much needed

Br1an67:
I'\''d like to work on this. Added `files_dir` option to `[features.spontaneous_file_upload]` in config.toml with a `get_files_directory()` helper that resolves relative paths from APP_ROOT.

PR: #2815

PR #2156, opened 2025-05-07
feat: Make file uploads directory configurable

brooklynbagel:
Fixes https://github.com/Chainlit/chainlit/issues/2155. 

I have tested this locally both running Chainlit using `chainlit run ...` and as a FastAPI sub-application with `uvicorn ...` and when setting a files directory in `config.toml` at `features.spontaneous_file_upload.files_dir` the configured files directory is created and becomes the location where uploaded files go. When it's not set, a default at `.files` is used.


backend/chainlit/config.py
```diff
@@ -50,10 +50,6 @@
 # Get the directory the script is running from
 APP_ROOT = os.getenv("CHAINLIT_APP_ROOT", os.getcwd())
 
-# Create the directory to store the uploaded files
-FILES_DIRECTORY = Path(APP_ROOT) / ".files"
```
hayescode, line 60: This removes APP_ROOT and is not added back anywhere

hayescode (changes requested):

hayescode (approved):

backend/chainlit/config.py
asvishnyakov: @hayescode Did you find out whether it’s added somewhere else?

hayescode (changes requested):
@brooklynbagel please add the files_dir to the config.toml string here https://github.com/Chainlit/chainlit/blob/main/backend%2Fchainlit%2Fconfig.py#L120

Issue #1841, opened 2025-01-31
CustomElement name when display = 'side'


lamylio:
Hello,

My issue is that both the source and the replacement text for a CustomElement are defined by the property `name`, therefore I cannot use a CustomElement with `display='side'`  and an arbitrary name. For example, if I want to have a `CustomPDFElement.jsx` that is displayed to the side, and that would replace the word '_source_' from a `Message()`, this is currently impossible (AFAIK). 

**Describe the solution you'd like**
I'd like to use another field, such as `element`, `source`, `element_name` on the constructor of `CustomElement`, and keep the behavior of the property `name` the same as all the built-in elements. ie, allowing this

```python
await Message(
    content="Blabla, blaba.\n Click here to see the source",
    elements=[
        CustomElement(
            element="CustomPDFElement", # any other field name you'll consider better
            name="source",
            props={...},
        )
    ],
).send()
```

PR #2158, opened 2025-05-07
feat: Add element_name to CustomElement to allow for distinct CustomElements

brunnelu:
Introduce the `element_name` attribute to the `CustomElement` class, allowing for distinct identification of custom elements. Update relevant methods and interfaces to ensure compatibility with this new attribute. For backwards compatibility the element_name equals the name of custom element if not defined.

This solves the issue: https://github.com/Chainlit/chainlit/issues/1841

I'll add the following example to the docs if you like this feature.

hello.py
```
import chainlit as cl
from chainlit import Message, on_chat_start


@on_chat_start
async def main():

    await cl.Message(
        content="Lorem Ipsum, ... source: source_1",
        author="AI Search",
        elements=[
            cl.CustomElement(
                element_name="IFrame",
                name="source_1",
                display="side",
                props={
                    "url": "https://example.com",
                    "title": "Some Title",
                },
            )
        ],
    ).send()
```

public/elements/IFrame.jsx
```
export default function IFrame() {
    return (
        <iframe
            src={props.url}
            title={props.title}
            style={{ width: "100%", height: "100%", border: "none" }}
        />
    );
}
```

### DIFF
--- a/backend/chainlit/element.py
+++ b/backend/chainlit/element.py
@@ -118,6 +118,7 @@ class Element:
                 "display": self.display,
                 "objectKey": getattr(self, "object_key", None),
                 "size": getattr(self, "size", None),
+                "element_name": getattr(self, "element_name", None),
                 "props": getattr(self, "props", None),
                 "page": getattr(self, "page", None),
                 "autoPlay": getattr(self, "auto_play", None),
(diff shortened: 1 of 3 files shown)


brunnelu:
Just to clarify: this change is fully backwards compatible. If element_name is not defined, it simply defaults to the value of name, preserving existing behavior. This update only adds the option to set a different identifier (element_name) when needed, without breaking existing code.

brunnelu:
@willydouhard, could you please take a look? I tink this will feature will be appreciated.

hayescode (changes requested):
You would need to update the data layers to do this.

hayescode:
no response. closing

philipplelidis:
Could we implement this correctly into the data layer, as it would be very helpful. This way, we could reference customElements as other standard elements (e.g., pdfs).

PR #2164, opened 2025-05-12
Fix full pathname for ws/socket.io.

galachad:
In copilot, the path was not generated correctly, when using custom `root_path`.

What I'm getting.
```
https://example.com/pathname//ws/socket.io/?EIO=4&transport=polling&t=PR4r3RN
```

What is expected:
```
https://example.com/pathname/ws/socket.io/?EIO=4&transport=polling&t=PR4r3RN
```

The path name needs to be corrected as the script is not handling redirect correctly.

### DIFF
--- a/libs/react-client/src/useChatSession.ts
+++ b/libs/react-client/src/useChatSession.ts
@@ -105,7 +105,7 @@ const useChatSession = () => {
       const uri = `${protocol}//${host}`;
       const path =
         pathname && pathname !== '/'
-          ? `${pathname}/ws/socket.io`
+          ? `${pathname}ws/socket.io`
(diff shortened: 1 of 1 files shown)


hayescode:
Can you add a check and only remove if the final character is a slash to be safe

asvishnyakov (changes requested):
We shouldn't concatenate URL as strings in first place, there is `URL` for that

hayescode:
@galachad where is root_path coming from? Forgive me I'm not super familiar with the copilot features and don't use them.

hayescode:
@galachad does your `custom_root_path` end in a `/`?

PR #2171, opened 2025-05-14
fix: only update metadata field when not None to prevent overwriting with {}

aitsc:
Only update SQL metadata field when metadata is not None; otherwise, existing metadata will be overwritten as an empty object ({}).


hayescode (approved):

hayescode:
@aitsc reverting this. The chainlit-datalayer repo metadata column to be NOT NULL.

asvishnyakov:
@aitsc I created a #2283 from this

hayescode:
@aitsc you can reopen this and we can merge again now

PR #2175, opened 2025-05-16
add pr scan workflow

yash-infra360:


hayescode (commented):
You should do this in your implementation of chainlit

PR #2205, opened 2025-06-03
fix: prevent constant refresh on the Login page

jfperusse:
When remaining on the Login page, users can notice the page constantly refreshing (approximately every 5 seconds). This is caused by SWR (stale-while-revalidate) which automatically retries the /user route on error (in this case, a 401), which causes the user to be set to undefined in userManagement.ts while fetching the latest status. Having an undefined user results in a full page loading indicator, making it look like Chainlit is reloading.

The main part of the fix is to not set the user to undefined while waiting for the /user route.

In addition to this, I've added an await for the handleAuth call in the Login page so that we properly keep the controls disabled while waiting for the authentication result (e.g. time taken by the auth_callback).

Finally, to further improve the user experience, I've disabled the username and password fields in addition to the login button while logging in.

Here are the updates in action:

chainlit_login_improvements


hayescode (approved):

asvishnyakov:
@hayescode Is there a reason you cancelled this PR merge? Please enable auto merge if everything is ok

libs/react-client/src/api/hooks/auth/userManagement.ts
```diff
@@ -17,8 +17,6 @@ export const useUserManagement = () => {
   useEffect(() => {
     if (userData) {
       setUser(userData);
-    } else if (isLoading) {
-      setUser(undefined);
     }
   }, [userData, isLoading, setUser]);
```
asvishnyakov: If `isLoading` is not used anymore, there is no need in this dependency

frontend/src/pages/Login.tsx
```diff
@@ -59,7 +59,7 @@ export default function Login() {
     formData.append('password', password);
 
     const jsonPromise = apiClient.passwordAuth(formData);
-    handleAuth(jsonPromise);
+    await handleAuth(jsonPromise);
```
asvishnyakov, line 62: `handleAuth` in `handleHeaderAuth` is promise too. May you also check this scenario, please?

hayescode:
@jfperusse is this still an issue in the latest release 2.6.7?

jfperusse:
> @jfperusse is this still an issue in the latest release 2.6.7?

Just tested in 2.6.8 and it is still an issue. I simply installed 2.6.8, pasted code from the documentation including the example under the password authentication, and configured the CHAINLIT_AUTH_SECRET. I then started the app and waited a few seconds to see the page refresh.

hayescode:
@jfperusse ok thanks. Please address @asvishnyakov comments then we're good to merge this.

jfperusse:
@hayescode Comments have been addressed. Thank you!

asvishnyakov (approved):

Issue #2193, opened 2025-05-27
BUG : Different chainlit apps makes the user logout on the other apps


dominpm:
## Description
When a user logs in to multiple Chainlit applications, logging into one app causes the user to be logged out of the others. This disrupts the user experience for anyone using more than one Chainlit app simultaneously. 

The different apps each do have a different CHAINLIT_AUTH_SECRET, is the cookie being shared? I am using Azure Oauth

## Steps to Reproduce
1. Log in to Chainlit App A.
2. In a separate browser tab or window, log in to Chainlit App B (a different Chainlit app instance).
3. Observe that the user is logged out from App A after logging into App B, and vice versa.

## Expected Behavior
Users should remain logged in to each Chainlit app independently. Logging into one Chainlit app should not log the user out of others.

## Actual Behavior
Logging into a second Chainlit app forces a logout on the first app (and vice versa), preventing concurrent use.

## Environment
- Chainlit version: 2.5.5
- Authentication method: Azure OAUTH

## Additional Context
This issue impacts users working across multiple Chainlit applications and can disrupt workflows.

dominpm:
@dosu any info?

dominpm:
@dosu 
What should a proper cookie name / scope /path etc look like?

dominpm:
@dosu 
the cookie path seems like is not configurable with .envs. I have serveral apps on same domain chainlit-app.com, but different sub-paths like chainlit-app.com/app1 chainlit-app.com/app2

dominpm:
#2207

dominpm:
Should close when #2311 merges

PR #2207, opened 2025-06-03
feat: update cookie path handling to use environment variable

dominpm:
### Problem  
The authentication cookie path in Chainlit was previously hardcoded to `/`, limiting flexibility for deployments that require a custom cookie path.

### Solution  
Introduced the `CHAINLIT_COOKIE_PATH` environment variable to allow configuration of the authentication cookie path.

- **Default behavior:** Remains `/` to ensure backward compatibility  
- **Scope:** Used for both setting and deleting authentication cookies  
- **Environment integration:** The variable is read at startup  

### Implementation Details  
- Added the following to `backend/chainlit/auth/cookie.py`:  
 ` _cookie_path = os.environ.get("CHAINLIT_COOKIE_PATH", "/")`

- Updated `set_auth_cookie()` and `clear_auth_cookie()` to use:  
  `path=_cookie_path`

###  Testing  
- Ran all existing tests using:  
  poetry run pytest

###  Backward Compatibility  
- If `CHAINLIT_COOKIE_PATH` is not set, the default path is `/`  
- Existing installations remain unaffected

### Notes  
- This design follows existing environment variable patterns in the codebase


dominpm:
Closes #2193

hayescode (approved):

dominpm:
Hi @hayescode , I have been thinking about the PR I did. I feel like I should update the following:

Instead of directly defining:

```python
_cookie_path = os.environ.get("CHAINLIT_COOKIE_PATH", "/")
```

If the env var `CHAINLIT_ROOT_PATH` is set, we could default the cookie_path to this exact path. I feel like it makes more sense. I will this afternoon commit those changes to the PR.

hayescode:
@dominpm yes good idea. Tag me once you get those changes pushed to this PR.

dominpm:
Since this PR merged and closed automatically I created the following one with the XS change @hayescode :

PR #2311

PR #2224, opened 2025-06-10
fix: properly parse user_env when `config.project.user_env` is empty

Atry:
Currently, when `config.project.user_env` is unset, the `user_env` is not parsed. As a result, `cl.user_session.create_accessor(key="env", default=None).get()` could return either a string of `'{}'` (if `config.project.user_env` is unset) or a non-empty dict (if `config.project.user_env` is non-empty).

I believe the behavior is not right. This PR makes `cl.user_session.create_accessor(key="env", default=None).get()` always return a `dict`.


hayescode:
@Atry please fix merge conflicts then we can merge.

hayescode:
@Atry tracing was removed a few releases ago. Update from main and remove the trace_event() please

Issue #1254, opened 2024-08-22
Why is the loading icon not showing?


frei-x:
**Describe the bug**
![image](https://github.com/user-attachments/assets/cb41d1f4-2e20-4d57-88de-62ccdb4a1609)

**To Reproduce**

1. Ask a question
2. Wait for an answer

chainlit>=1.1.400 has problems, the old version works fine

**expected**
![image](https://github.com/user-attachments/assets/2d9ef1a6-8fff-4eca-bd2c-a6866dad2ca7)

**software**

Python 3.10.12

Linux version 5.15.0-105-generic (buildd@lcy02-amd64-007) (gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0, GNU ld (GNU Binutils for Ubuntu) 2.38) #115-Ubuntu SMP Mon Apr 15 09:52:04 UTC 2024

```python
    ai_msg = cl.Message(content="")
    await ai_msg.send()
   
    if len(message_history) > MAX_ROUNDS:
        message_history = message_history[:MAX_ROUNDS //
                                          2] + message_history[-MAX_ROUNDS//2:]
    stream = llm.stream([system_message] +
                        message_history)
    if (DEBUG):
        print("history", str(message_history))
    full_answer = ""
    for part in stream:
        if content := part.content or "":
            full_answer += content
            if (len(full_answer) == 1):
                gen_time_end = time.time()
                print(gen_time_end - gen_time_start)
            await ai_msg.stream_token(content)
    message_history.append({"role": "assistant", "content": ai_msg.content})
    cl.user_session.set("message_history", message_history)
    await ai_msg.update()
```

I don't know anything else, config.toml is the default,

When streaming a conversation, before the first word appears, chainlit>=1.1.400 will not display the loading icon

dokterbob:
Hi @frei-x, thanks for the feedback? Could you please follow the feedback format for bug reports, it will make it easier for us to assess and replicate your issue.

Specifically, we need to know the expected behaviour. What should be shown in the red box? What does it look like prior to 1.1.400? What platform are you using? Does this problem occur on other platforms etc. etc.?

frei-x:
> When streaming a conversation, before the first word appears, chainlit>=1.1.400 will not display the loading icon

Sorry, I updated the issue and have not tried other environments yet.

When streaming a conversation, before the first word appears, chainlit>=1.1.400 will not display the loading icon

3eeps:
i have the same issue.

[...]

PR #2229, opened 2025-06-18
Fixed issue where blinking indicator is not displayed when all the steps finish

Allaoua9:
This is related to this issue :
https://github.com/Chainlit/chainlit/issues/1254

The fix is that we need to check that all steps have finished execution to display blinking indicator.

### DIFF
--- a/frontend/src/components/chat/Messages/index.tsx
+++ b/frontend/src/components/chat/Messages/index.tsx
@@ -22,10 +22,13 @@ interface Props {
 
 const CL_RUN_NAMES = ['on_chat_start', 'on_message', 'on_audio_end'];
 
-const hasToolStep = (step: IStep): boolean => {
+const hasActiveToolStep = (step: IStep): boolean => {
   return (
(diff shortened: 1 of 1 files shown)


hayescode (changes requested):
@Allaoua9 please remove the "type==tool" since users can modify the tool value. All else looks good and I'll merge once fixed.

asvishnyakov:
@hayescode But `tool` was there before this PR. Is it a separate bug?

hayescode:
> @hayescode But `tool` was there before this PR. Is it a separate bug?

@asvishnyakov step.type doesn't have to be 'tool'. For example I set mine to be 'mcp' or 'code_interpreter' for example

hayescode (approved):

hayescode:
@Allaoua9 @asvishnyakov once the step.type is moved from literalai into Chainlit we can correctly differentiate steps from other messages to handle cases where step.type != tool

PR #2245, opened 2025-06-30
feat: Reduce LiteralAI dependencies

hayescode:
LiteralAI will be going away and is inter-twined with Chainlit. This PR intends to begin reducing our dependencies on LiteralAI and get some PRs merged, with the intention of fully removing in the future. Starting with datetime functions that can just be in Chainlit directly.


asvishnyakov (commented):
I would propose to make LiteralAI opt-in feature until it will be completely shutdown

hayescode:
@asvishnyakov good to review now

PR #2257, opened 2025-07-02
chore: upgrade poetry, node, pnpm & enable corepack support

asvishnyakov:
- Upgraded `poetry` to latest version
- Upgraded `node` version
  - Required because otherwise may cause `Cannot find matching keyid` error due to keys rotation in npm registry
  - Used 24 version as didn't want to downgrade to 22 and 24 will reach LTS status soon
- Upgraded `pnpm` to latest version
- Added `corepack` support
  - Users without it still can use (and contribute) as usual
  - Users with it will get automatic installation of proper version
  - Used it in CI
  - Moved version spec to `package.json` - not only required by `corepack`, but also useful for frontend developers (you can clearly see which version required)
 - Disabled `actionlint` for reused actions - [actionlint support only workflows](https://github.com/rhysd/actionlint/issues/350)


hayescode (approved):

asvishnyakov:
@hayescode I'm not sure if it's something we want to include in the same release where we plan to address important fixes, but if you feel it's okay - feel free to merge this PR.

PR #2265, opened 2025-07-04
Import timestamp_utc from chainlit.utils

joro75:
Due to changes from #2245, the openai and mistralai sub-modules cannot be loaded anymore, as timestamp_utc is not loaded from the chainlit.utils.

These changes correctly import timestamp_utc again.


AbhijithMallya:
@joro75 thank you

hayescode (approved):

Issue #2264, opened 2025-07-04
Chainlit 2.6.0 - Pasting Content Has Several Additional Issues After Newlines Fix


AidanShipperley:
## Describe the bug
The latest Chainlit 2.6.0 release had a PR to fix input spacing issues (#2246), which attempted to address newline issues by switching from `insertHTML` to `insertText` with HTML escaping, but this created a cascade of new problems (see below). This is the fourth major issue (#1890, #2032, #2087) with input handling in the main message composer since v2.0.5 introduced [inline commands](https://docs.chainlit.io/concepts/command), which at the time mimicked how ChatGPT rendered commands in their UI as in line with the text at the beginning of the composed message.

When we look at the Edit Message Composer in the [`UserMessage.tsx`](https://github.com/Chainlit/chainlit/blob/09d7d5e745961176c6f393afedd6ec607129638b/frontend/src/components/chat/Messages/Message/UserMessage.tsx) file, which has had no issues dealing with pasting content at all, we can see it is using [`AutoResizeTextarea`](https://github.com/Chainlit/chainlit/blob/09d7d5e745961176c6f393afedd6ec607129638b/frontend/src/components/chat/Messages/Message/UserMessage.tsx#L88-L95) - which is essentially a `<textarea>` element.

The Main Message Composer ([`Input.tsx`](https://github.com/Chainlit/chainlit/blob/09d7d5e745961176c6f393afedd6ec607129638b/frontend/src/components/chat/MessageComposer/Input.tsx)), however, uses a `contentEditable` div with tons of custom logic for handling escaping html elements, handling pasted content properly, allowing undoing/redoing properly, spacing, etc.

The use of `contentEditable` is solely to support rendering the name of the user's selected command in line with the text - a feature that ChatGPT does not even use anymore. There have been continual issues in each release since v2.0.5 that significantly impact the ability for users to use Chainlit across different platforms for editing and understanding content pasted by them. It seems reasonable at this point for Chainlit to adopt an approach for commands similar to what ChatGPT does today (using buttons in combination with a toolbox under the chat input) as properly working with `contentEditable` across all browsers and for all content will explode the complexity of the [`Input.tsx`](https://github.com/Chainlit/chainlit/blob/09d7d5e745961176c6f393afedd6ec607129638b/frontend/src/components/chat/MessageComposer/Input.tsx) file and it seems unreasonable to expect the code to handle all possible edge cases perfectly.

[...]

PR #2267, opened 2025-07-06
fix: preserve formatting and special characters in pasted content

rajanarahul93:
###  Description:
This PR improves the paste handler in the message composer to:
- Preserve multiline formatting (newlines, indentation)
- Avoid escaping special characters like `<`, `>` and quotes
- Remove deprecated `execCommand` usage
- Use safe DOM insertion via `TextNode` and `<br>` elements
- Retain undo/redo functionality through native input events

###  Fixes:
- #2264 
- Resolves contentEditable paste issues for code formatting and symbols.


AidanShipperley:
Hi @rajanarahul93,

Thank you very much for taking a shot at addressing this issue!

This addresses the problems with the HTML elements like `<`, `>`, and `'`, and it seems to also address the newline duplication issue that is present in 2.5.5.

However, there are unfortunately still a few problems with how input content is handled and how it ends up being sent to the UI.

### Undo/Redo Functionality is Broken

While you mentioned undo/redo is preserved with your changes, upon testing in Chrome and Edge on Windows, the undo/redo functionality (Ctrl+Z / Ctrl+Y) is still non-functional with this change. In version 2.5.5, despite its issues with duplicated newlines in pasted content, users could still undo and redo because it used `document.execCommand('insertHTML', false, htmlToInsert)`. While `execCommand` is deprecated, it actually integrates with the browser's native undo/redo stack, which ended up solving this problem.

Since your approach uses direct DOM manipulation:
```tsx
range.deleteContents();
range.insertNode(fragment);
```

The browser doesn't register these direct DOM manipulations as user actions, so they don't get added to the undo history. This is a fundamental limitation when trying to work around `contentEditable`'s quirks with manual DOM manipulation.

### Consecutive Newlines Are Not Preserved When Sending

While the visual rendering in the composer now correctly shows multiple consecutive newlines when pasting content, they are not preserved when rendering the message that is sent. For example, if you paste:

```python
print("Line 1")

print("Line 3")




print("Line 8")
```

The composer correctly displays all 4 empty lines between line 3 and 8. However, when sent, the rendered message only shows a single newline between each print statement. This could happen from a few places, but I have not had time to track it down.

---

There may be other issues, but I found these through some small testing.

Regardless of whether or not we can get newlines working, I think that unless we can find a way to preserve undo/redo without the deprecated `execCommand`, we may need to just move away from inline command rendering, especially given that ChatGPT does not even have this functionality anymore and this implementation has caused so many different issues.

I still will keep working on a PR to mimic what ChatGPT does today but definitely let me know if you find a way around these issues without relying on `execCommand`!

rajanarahul93:
I've updated the paste handler to use execCommand('insertHTML'), so undo/redo now works as expected. Also, improved content extraction to properly preserve multiple newlines when sending. Let me know if you require any other changes.

hayescode:
@AidanShipperley @asvishnyakov can you help test this? This is working on my setup (windows 11 and both Edge/Chrome) but older issues regarding this suggest Mac and different browsers never had the extra newline issues.

Hopefully we can get this core component of the framework working properly and move past this. It's been whack-a-mole for months so I want to ensure this PR fixes the input issues for all setups without any regressions before merging.

asvishnyakov:
@hayescode @AidanShipperley Yes I'll test it locally

rajanarahul93:
Thanks for testing this when you can. Let me know if anything breaks on Mac or other browsers, and I’ll be happy to tweak further.

asvishnyakov:
@hayescode @rajanarahul93 @AidanShipperley Y’all almost broke my brain with this PR while I was trying to understand why all this code was needed in the first place.

First of all, I don’t understand why a `<div>` with `contentEditable` is used instead of a regular `<textarea>`. From what I gather in the comments, it’s due to some kind of *inline command rendering* - something I’m not familiar with and that even ChatGPT no longer supports.

Regarding the changes made in this PR:

1. HTML escaping probably wasn’t perfect from the beginning, but this PR introduces a new issue with escaping ampersand-followed sequences. Try pasting this:

   ```html
   <h1>My First Heading&amp;</h1>
   ```
   The result:
   ```text
   <h1>My First Heading&</h1>
   ```
   I propose using `insertText` instead, as it automatically escapes HTML and integrates with the undo/redo stack. If that causes any issues during testing, escape using `createTextNode`:
   ```javascript
   document.executeCommand('insertHTML', false, document.createTextNode(textData).textContent)
   ``` 
   There’s also `contentEditable="plaintext-only"`, which already handles all of this and is now supported in Firefox starting from March 2025 (all other browsers have supported it since like 2010s).
2. `if (!textData) return;` prevents files from being pasted. `onPaste` should be called regardless of whether text is found.
https://github.com/Chainlit/chainlit/blob/4e46a42db8200b7a48df9ac2f796396f661b8a29/frontend/src/components/chat/MessageComposer/index.tsx#L61-L75
3. Please explain other changes, especially about the regular expression, as I don't see any difference with current code, except removal of (making the code easier to read) comments. If that's because of linter, then I have no questions about them.

asvishnyakov:
@willydouhard Would be glad to hear your side of the story about the birth of this monstrous code

rajanarahul93:
Hey @asvishnyakov , thanks for taking the time to look through all of this.

Yeah, the regex stuff was my attempt to clean up the pasted HTML and make sure line breaks and special characters showed up right. But I see what you mean about the escaping issue with `&amp;`. That's a problem.

Your idea about using `insertText` (or `createTextNode`) could avoid all that messy parsing and still keep undo/redo working. I’ll give that a try and see if it handles the newlines and escaping better.

I’m also going to look into using `contentEditable="plaintext-only"` if that can simplify everything.

If there’s anything else you think I should try or keep in mind, let me know. I want to make sure to fix this in a way that’s clean and works everywhere.

asvishnyakov:
@rajanarahul93 That's all I could find and think of for now - maybe something will come to mind with a fresh head tomorrow. Thank you for your contribution & time you spent on this issue!

As I said, I don't quite understand why we're using a `div` with `contentEditable` for plain text editing and copy-pasting. I would be really very grateful if someone could explain it to me, because the easiest and most bug-proof approach here would be to just use a `textarea` and remove all the extra logic on top of it.

AidanShipperley:
> As I said, I don't quite understand why we're using a `div` with `contentEditable` for plain text editing and copy-pasting. I would be really very grateful if someone could explain it to me, because the easiest and most bug-proof approach here would be to just use a `textarea` and remove all the extra logic on top of it.

Hi @asvishnyakov, based on my testing and looking at the commit history, the only reason Chainlit is not using [`<AutoResizeTextarea>`](https://github.com/Chainlit/chainlit/blob/160222cc2f90953776229cbad0f9d2778e37ceff/frontend/src/components/AutoResizeTextarea.tsx) for the input message composer (*which is what we are using when a user edits a message and it has had no issues*) is because in [Chainlit v2.0.5](https://github.com/Chainlit/chainlit/releases/tag/2.0.5), Willy [added chat commands](https://github.com/Chainlit/chainlit/commit/0642b1114c38e9b7f970db3e649e5fa246358f3d#diff-f603b49a978423aba6e77034f65c84aa6a0d868c61c0a267c5dceb44d2dfb819) which are ***rendered inline*** with the user's message, which need a custom blue bold look and the selected command is supposed to be fixed to the front of the user's message. Here is an example on the latest Chainlit 2.6.0 release. You can replicate this using the code on the wiki [here](https://docs.chainlit.io/concepts/command#set-available-commands):

> https://github.com/user-attachments/assets/3d2e78f8-7641-4d4a-8182-d65adefbd108


---

Since `<textarea>` doesn't support rich text, Willy [created the `Input.tsx` file](https://github.com/Chainlit/chainlit/commit/0642b1114c38e9b7f970db3e649e5fa246358f3d#diff-f603b49a978423aba6e77034f65c84aa6a0d868c61c0a267c5dceb44d2dfb819) and removed the [`<AutoResizeTextarea>`](https://github.com/Chainlit/chainlit/blob/160222cc2f90953776229cbad0f9d2778e37ceff/frontend/src/components/AutoResizeTextarea.tsx) to use a `contentEditable` div, which has its own markup that needs to be handled and requires escaping for raw HTML. This creates a fragile conversion pipeline where fixing one issue (like newlines) breaks another (like HTML entities) and there have been issues with it in every single release since v2.0.5. However, we can't just switch to a `<textarea>` alone here, as the user needs some way to tell what command they have selected (see the video above).

@rajanarahul93 I mentioned this in my issue (#2264), but I have a PR almost finished that replaces the command UX with a more modern ChatGPT-like look, and it ***switches back to using an [`<AutoResizeTextarea>`](https://github.com/Chainlit/chainlit/blob/160222cc2f90953776229cbad0f9d2778e37ceff/frontend/src/components/AutoResizeTextarea.tsx)*** as a more future-proof solution to avoid all of the workarounds. It is however a bigger change across a few files to support a more modern look, and I don't want to prevent the fix here from going through though in the short-term, as if this works it will be less changes to the codebase.

If it becomes too complicated to get this fix merged with no bugs, I will definitely try to add you as a co-author to my PR in case they can move forward with it in a future release, as I definitely appreciate you taking a stab at this complicated issue.

asvishnyakov:
@AidanShipperley Thank you for the good explanation! Rich text support was my first guess for why the code became so complex, but I noticed that we currently only support plain text (`event.clipboardData.getData('text/plain')`), and that’s been the case even before the recent changes.

For now, I think the best approach is to try fixing the current code without breaking anything else (I’ve suggested a few possible solutions), get that released, and then revisit this part of the implementation with the goal of simplifying it. I’d be happy to review or help with that refactor when the time comes.

AidanShipperley:
> @AidanShipperley Thank you for the good explanation! Rich text support was my first guess for why the code became so complex, but I noticed that we currently only support plain text (`event.clipboardData.getData('text/plain')`), and that’s been the case even before the recent changes.
> 

Yeah, you are correct that only plain text is supported from the user; as far as I can tell, the only reason rich text support was added was for [this code here](https://github.com/Chainlit/chainlit/blob/30a82b50fae55bf9eaff645d8611c953bceddd2c/frontend/src/components/chat/MessageComposer/Input.tsx#L173-L178):

```javascript
// Create new command block
const newCommandBlock = document.createElement('div');
newCommandBlock.className =
  'command-span font-bold inline-flex text-[#08f] items-center mr-1';
newCommandBlock.contentEditable = 'false';
newCommandBlock.innerHTML = `<span>${selectedCommand.id}</span>`;
```

This is the code that applies `font-bold` and `text-[#08f]` to the selected chat command, rendering it as bold and blue so it looks like ChatGPT's commands used to look (as seen in the video I posted above). The code uses `contentEditable` specifically to allow this inline rich text formatting of the command within the user’s regular text input.​​​​​​​​​​​​​​​​

---

Based on your feedback, I will hold off on submitting a separate PR until we get this fix into a release, but let me know if anything changes. I can help test this in the meantime. I do think long term switching away from `contentEditable` to a `<textarea>` is the right approach.

hayescode:
@rajanarahul93 @AidanShipperley @asvishnyakov how are we looking on this? I'd like to include this in the next release if it's ready.

asvishnyakov:
@hayescode The PR in its current state is not ready, as it doesn't fully resolve the issue and introduces another one (related to pasting files). I’ve proposed a few solutions - if the @rajanarahul93 doesn’t have the time or ability to update the PR, I can take care of it myself. Personally, I believe using `contentEditable="plaintext-only"` is the best solution, as it eliminates most of the problematic code and works in all major browsers since around 2015, except for Firefox versions earlier than 135, which was released in March 2025. I consider this acceptable.

rajanarahul93:
@asvishnyakov  Sorry for the delay ,I haven’t had the time yet to try the changes you suggested. I’ll give it a shot and let you know if I run into any issues.

rajanarahul93:
**Updated based on feedback:**

- Replaced manual escaping with `insertText`
- Added `contentEditable="plaintext-only"` for safer HTML handling
- Ensured file pasting works consistently via `onPaste(event)`
- Preserved undo/redo behavior

@AidanShipperley , open to any further changes, or happy to support your PR as a co-author if that’s the better direction.

bdavj:
The fix for this doesn't appear to have worked as far as I can see. Using Chainlit 2.7.2 
Screenshot 2025-10-01 at 15 32 25

PR #2273, opened 2025-07-08
change width of element view in copilot

FeizNouri:
the width of the element view in copilot is fixed and a bit small.

this changes the max width to be 80% which is convenient for many types of element like md.

this is photo screenshot comparing the change in width :

before : 
Screenshot from 2025-07-08 18-14-54

after: 
Screenshot from 2025-07-08 18-09-53


Jimmy-Newtron:
Hello @hayescode ,
can you review this Copilot PR for the next release?
Thanks in advance

hayescode (approved):

Issue #2258, opened 2025-07-02
update from 2.5.5 to 2.6.0


GodgyQing:
2025-07-02 15:31:22,506 - asyncio[base_events.py:1758] - ERROR - Task exception was never retrieved
future: <Task finished name='Task-1281' coro=<ChainlitDataLayer.create_step() done, defined at f:\myproject\ai-cctc-autogen\venv\lib\site-packages\chainlit\data\utils.py:10> exception=PostgresSyntaxError('syntax error at or near ";"')>
Traceback (most recent call last):
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.3056.0_x64__qbz5n2kfra8p0\lib\asyncio\tasks.py", line 232, in __step
    result = coro.send(None)
  File "f:\myproject\ai-cctc-autogen\venv\lib\site-packages\chainlit\data\utils.py", line 25, in wrapper
    return await method(self, *args, **kwargs)
  File "f:\myproject\ai-cctc-autogen\venv\lib\site-packages\chainlit\data\chainlit_data_layer.py", line 296, in create_step 
    await self.update_thread(thread_id=step_dict["threadId"])
  File "f:\myproject\ai-cctc-autogen\venv\lib\site-packages\chainlit\data\chainlit_data_layer.py", line 571, in update_thread
    await self.execute_query(query, {str(i + 1): v for i, v in enumerate(values)})
  File "f:\myproject\ai-cctc-autogen\venv\lib\site-packages\chainlit\data\chainlit_data_layer.py", line 61, in execute_query
    records = await connection.fetch(query, *params.values())
  File "f:\myproject\ai-cctc-autogen\venv\lib\site-packages\asyncpg\connection.py", line 690, in fetch
    return await self._execute(
  File "f:\myproject\ai-cctc-autogen\venv\lib\site-packages\asyncpg\connection.py", line 1864, in _execute
    result, _ = await self.__execute(
  File "f:\myproject\ai-cctc-autogen\venv\lib\site-packages\asyncpg\connection.py", line 1961, in __execute
    result, stmt = await self._do_execute(
  File "f:\myproject\ai-cctc-autogen\venv\lib\site-packages\asyncpg\connection.py", line 2004, in _do_execute
    stmt = await self._get_statement(
  File "f:\myproject\ai-cctc-autogen\venv\lib\site-packages\asyncpg\connection.py", line 432, in _get_statement
    statement = await self._protocol.prepare(
[...]

PR #2277, opened 2025-07-10
fix: revert #2171 to temporarely fix issue with null metadata on thread creation (#2270)

### LINKED ISSUE #2270: Release 2.6.0 gives error while trying to write to PGSQL datalayer
**Describe the bug**
Release 2.6.0 gives error while trying to write to PGSQL datalayer

**To Reproduce**
Simple Chainlit 2.6.0 with authentication and official data layer ( https://github.com/Chainlit/chainlit-datalayer )

DETAIL: Failing row contains (6d64a174-609d-4acb-b369-aaaaaaaaa, 2025-07-08 09:33:03.933, 2025-07-08 09:33:03.933, null, any news?, null, 3e6088d8-1ef1-4eb9-8c66-ccccccccccc, {}).
2025-07-08 09:33:03 - Database error: null value in column "metadata" of relation "Thread" violates not-null constraint

hayescode:
Fixed #2270 by reverting #2171 

The chainlit data layer schema requires the metadata column to be NOT NULL.

### DIFF
--- a/.github/workflows/close_stale.yml
+++ b/.github/workflows/close_stale.yml
@@ -16,13 +16,13 @@ jobs:
           days-before-issue-stale: 14
           days-before-issue-close: 7
           stale-issue-label: "stale"
-          exempt-issue-labels: "enhancement,dev-tooling,e2e-tests,unit-tests"
+          exempt-issue-labels: "enhancement,dev-tooling,e2e-tests,unit-tests,keep-for-a-while"
           stale-issue-message: "This issue is stale because it has been open for 14 days with no activity."
           close-issue-message: "This issue was closed because it has been inactive for 7 days since being marked as stale."
           days-before-pr-stale: 14
           days-before-pr-close: 7
           stale-pr-label: "stale"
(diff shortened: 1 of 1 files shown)


asvishnyakov:
@hayescode I belive it rather should be
```python
if metadata:
  data["metadata"] = metadata
```
Let me check this with real db, and may be I'll be able to add tests

asvishnyakov:
@hayescode Ah, ok, I see

https://github.com/Chainlit/chainlit-datalayer/blob/8844b1b318112961f8589b0614474a85530995df/prisma/migrations/20250103173917_init_data_layer/migration.sql#L80

and `update_thread` is also used to create threads (without metadata). So the only ways to fix both bugs are either to make `metadata` nullable in the DB or to create a separate `create_thread` method, which some users have also suggested on Discord. Let's create an issue and merge this PR as a temporary solution.

asvishnyakov:
@hayescode Can you remove workflow from this PR?

asvishnyakov:
Created #2283 and #2284

asvishnyakov:
It will also fix #2258

Issue #2284, opened 2025-07-11
Create new create_thread method for data layers


asvishnyakov:
**Is your feature request related to a problem? Please describe.**
- Current approach (when you need to call `update_thread` to create thread) is not obvious
- This is possible fix for #2283

**Describe the solution you'd like**
Create `create_thread` function in data layer

**Describe alternatives you've considered**
-

**Additional context**
-

PR #2293, opened 2025-07-15
fix: postgres query syntax error in data_layer

royp-nrd:
Original behavior: 
`asyncpg.exceptions.PostgresSyntaxError: syntax error at or near ";"`

Caused by this query:
```
postgres-1    | 2025-07-15 12:26:38.419 UTC [544] STATEMENT:  
postgres-1    |                     INSERT INTO "Thread" ("id")
postgres-1    |                     VALUES ($1)
postgres-1    |                     ON CONFLICT (id) DO UPDATE
postgres-1    |                     SET ;
```

which happens when `update_sets` is empty list. 

After suggested fix there is no query error. This is the new query:
```
postgres-1    | 2025-07-15 12:28:44.355 UTC [562] STATEMENT:  
postgres-1    |                     INSERT INTO "Thread" ("id")
postgres-1    |                     VALUES ($1)
postgres-1    |                     ON CONFLICT (id) DO NOTHING
```

### DIFF
--- a/backend/chainlit/data/chainlit_data_layer.py
+++ b/backend/chainlit/data/chainlit_data_layer.py
@@ -561,12 +561,19 @@ class ChainlitDataLayer(BaseDataLayer):
 
         update_sets = [f'"{k}" = EXCLUDED."{k}"' for k in data.keys() if k != "id"]
 
-        query = f"""
-            INSERT INTO "Thread" ({", ".join(columns)})
-            VALUES ({", ".join(placeholders)})
-            ON CONFLICT (id) DO UPDATE
-            SET {", ".join(update_sets)};
-        """
+        if update_sets:
+            query = f"""
+                INSERT INTO "Thread" ({", ".join(columns)})
+                VALUES ({", ".join(placeholders)})
+                ON CONFLICT (id) DO UPDATE
+                SET {", ".join(update_sets)};
+            """
+        else:
+            query = f"""
+                INSERT INTO "Thread" ({", ".join(columns)})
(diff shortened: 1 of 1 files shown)


asvishnyakov:
@hayescode Seems like in our investigation we missed last crucial step 😂:
https://github.com/Chainlit/chainlit/blob/160222cc2f90953776229cbad0f9d2778e37ceff/backend/chainlit/data/chainlit_data_layer.py#L554-L555
#2171 will work with this change, but to be fair I would still like to implement separate `create_thread` (#2284)

asvishnyakov:
@royp-nrd Thank you for the contribution! We've already temporarily reverted #2171, which caused this issue, but I’d be happy to fix it properly without removing the original fix for the other bug.

hexart:
This seems overly complex. The `update_sets` list is unlikely to be empty in practice since we're always updating at least the metadata field. Simply removing the trailing semicolon would be sufficient to fix the syntax error.

royp-nrd:
> This seems overly complex. The `update_sets` list is unlikely to be empty in practice since we're always updating at least the metadata field. Simply removing the trailing semicolon would be sufficient to fix the syntax error.

I tried this at first and it still raise syntax error.

asvishnyakov:
@hexart `metadata or {}` causes another bug with metadata overwriting and needs to be fixed. Our previous attempt to address it led to that error. I’m considering a proper fix by introducing a separate `create_thread` method, but until then, this fix looks fine.

hayescode (approved):

PR #2308, opened 2025-07-22
add french (fr-FR) translation

Starland9:


hayescode:
@Starland9 please fix CI

hayescode (commented):
@Starland9 since you added the audioop-lts you'll need to run `poetry lock` and push the updated lock file.

asvishnyakov:
@Starland9 Can you explain why we need directly depend in this and how it related to translations? I don't see any direct usage of it in Chainlit source code @hayescode We already have it in dependencies via `discord` package, which belongs to `tests` dependencies group

hayescode:
These 2 things are totally unrelated. Should be separate PRs but it's such a small change I'll allow it this time

hayescode:
@Starland9 update the poetry.lock file to pass the CI

hayescode:
@Starland9 i added audioop-lts to main and updated the poetry lock, so this PR will only be for the FR translation.

PR #2311, opened 2025-07-23
refactor: default cookie path to CHAINLIT_ROOT_PATH if set

dominpm:
### Description

This PR follows up on [#2207](https://github.com/Chainlit/chainlit/pull/2207) to refine how the authentication cookie path is determined.

### What's Changed

- If `CHAINLIT_ROOT_PATH` is defined, it is now used as the default value for the cookie path.
- This takes precedence over the `CHAINLIT_COOKIE_PATH` env var (which still remains configurable).
- Maintains backward compatibility: if neither is set, the path defaults to `/`.

### Why

In deployments where `CHAINLIT_ROOT_PATH` is used to scope all routes, it makes more sense for cookies to follow the same base path by default — ensuring consistent scoping without requiring extra configuration.

### Notes

@hayescode as discussed in #2207 — this follows up. (The other PR was automatically merged and closed) Let me know if you'd like to adjust anything further!


backend/chainlit/auth/cookie.py
```diff
@@ -23,7 +23,10 @@
     "Invalid value for CHAINLIT_COOKIE_SAMESITE. Must be one of 'lax', 'strict' or 'none'."
 )
 _cookie_secure = _cookie_samesite == "none"
-_cookie_path = os.environ.get("CHAINLIT_COOKIE_PATH", "/")
```
hayescode, line 26: @dominpm why not just change this to CHAINLIT_ROOT_PATH

dominpm: To ensure backwards compatibility if anyone is using the other env var. Not just delete the old one. I can do it though if you feel it’s best.

asvishnyakov:
@hayescode @dominpm Before you merge and release this, keep in mind Chainlit set multiple cookies (I know at least about the one for affinity sessions) which DOES NOT respect auth cookie settings (which is a separate issue). So I think the proper name for variable is `CHAINLIT_AUTH_COOKIE_PATH`

backend/chainlit/auth/cookie.py
hayescode: Makes sense, disregard

dominpm:
Hi @hayescode , just edited the change for the env var name in the PR

PR #2331, opened 2025-07-28
fix: copy text as rendered

jdb78:
When copy-pasting text via the copy-button and the next is in markdown, it will paste the raw markdown while most users will want the rendered html version. This PR fixes this by adding the html version to the clipboard

Before in Word, when you copy text:
Screenshot 2025-08-01 at 10 30 53

Now, it formats properly
Screenshot 2025-08-01 at 10 33 22

While still pasting raw markdown into editors:
Screenshot 2025-08-01 at 10 34 28


hayescode:
@jdb78 can you add before/after screenshots to your PR description?

jdb78:
updated the description. Put in the how the copied text now looks @hayescode

hayescode:
@jdb78 on the latest version 2.6.3 this already works. what version are you on?

jdb78:
I am on the latest version and it certainly doesn't currently paste formatted text on Mac nor iPhone. When pressing the "copy" button, currently `navigator.clipboard.writeText(textToCopy)` writes to the clipboard with mime = test/plain. Therefore text is pasted unformatted into applications.

What the PR does, it writes different formats/mime to the clipboard, so the application that you paste into can make its pick. It uses the exact same conversion from markdown to HTML logic as the main application to ensure consistency. Currently, if you copy not using the copy button, you will also copy background, etc. and this is also a suboptimal experience on mobile.

asvishnyakov:
@hayescode It doesn’t work on Linux either, and I’m not sure how it can work anywhere including Windows. Most rich-text editors require content to be copied with the `text/html` MIME type to be treated as rich text and rendered correctly.

@jdb78 Thank you for the PR! That said, I’d classify this more as a feature, since Chainlit currently supports both copying and pasting only as plain text, which may be intentional. I’ll review the PR and would be happy to merge it if everything looks good. However, my first thought is this: we already render Markdown to HTML for display in the thread, so there should already be code (possibly in a dependency) handling that. And since the message is already rendered by the time the user can copy it, can’t we just extract the HTML from the page instead of rendering it again?

jdb78:
@asvishnyakov this is very much true. created a new PR with changed code: #2353

PR #2353, opened 2025-08-03
feat: copy both plain text and rich text to clipboard, if available

jdb78:
Replace #2331 and directly access rendered component to copy information to clipboard


asvishnyakov:
@jdb78 Did you tested it with commands?

jdb78:
Commands don't seem to make a difference as they are not relevant when copying the text. Maybe I don't understand the question correctly?

asvishnyakov (approved):
Sorry, I forgot we can copy only AI assistant messages :) LGTM

Issue #2384, opened 2025-08-08
Commands Menu disappears after unmount and remounting the plugin


maoguy:
**Describe the bug**
After calling window.mountChainlitWidget, window.unmountChainlitWidget, and window.mountChainlitWidget in sequence, the menu is lost.

**To Reproduce**

1. use window.mountChainlitWidget to mount a chainlit plugin;
2. use window.unmountChainlitWidget to unmount the chainlit plugin of step 1;
3. use window.mountChainlitWidget to mount a chainlit plugin again;

**Expected behavior**

Commands Menu disappears after unmount and remounting the plugin

**Screenshots**

When we use window.mountChainlitWidget success,we can see the commands menu is showing corrected.👇

<img width="449" height="242" alt="Image" src="https://github.com/user-attachments/assets/c258c21d-7a75-43e3-9cd4-1254a98b8841" />

However,if we use window.unmountChainlitWidget to unmount.After that we use window.mountChainlitWidget again,we will can not see the commands menu.👇

<img width="433" height="249" alt="Image" src="https://github.com/user-attachments/assets/65b05a1b-0441-4684-969e-40219a146110" />

I believe the root cause is that after the plug-in was uninstalled and remounted, no new session was initiated.Therefore the websocket of backend don`t know when should push the new config again.👇

<img width="1280" height="613" alt="Image" src="https://github.com/user-attachments/assets/4cb81759-9985-4024-87b4-4dfa3e1d3954" />

maoguy:
I believe the plugin should expose configurable lifecycle hooks—such as a "afterMounted" that fires once the plugin is successfully mounted,or a "beforeUnmounted" hook that is invoked right before the plugin is torn down.

With either of these two lifecycle hooks available, we could call window.clearChainlitCopilotThreadId inside mountedCallback (to start a fresh session after remount) or inside beforeUnmounted (to clean up the current session before the plugin is torn down), ensuring the thread ID is properly reset.

_**If needed, I’m happy to submit a PR that adds support for both mountedCallback and beforeUnmounted.☺**_

maoguy:
[...]

PR #2385, opened 2025-08-08
Update widget.tsx

maoguy:
Before Widget unmounted and clear the method of window.clearChainlitCopilotThredId, call the window.clearChainlitCopilotThreadId?.();

which can fix the bug : https://github.com/Chainlit/chainlit/issues/2384

### DIFF
--- a/libs/copilot/src/widget.tsx
+++ b/libs/copilot/src/widget.tsx
@@ -39,6 +39,7 @@ const Widget = ({ config, error }: Props) => {
       window.toggleChainlitCopilot = () => console.error('Widget not mounted.');
       window.getChainlitCopilotThreadId = () => null;
 
+      window.clearChainlitCopilotThreadId?.();
       window.clearChainlitCopilotThreadId = () =>
         console.error('Widget not mounted.');
     };


xjose97x:
@maoguy I don't quite understand this issue. Could you please provide more details?

maoguy:
> @maoguy I don't quite understand this issue. Could you please provide more details?

# Reproduce
To Reproduce,you can try to do that 👇:
1.【Backend】Set the commands and setting in just like👇【docs ref 👉 https://docs.chainlit.io/concepts/command#command】
image
image
2.【Frontend】Use the method of window.mountChainlitWidget to mount the widget firstly,so we can see the corrected view 👇
image
3.【Frontend】Use the method of window.unmountChainlitWidget() to unmount the widget,and then use the method of window.mountChainlitWidget to remount the widget again.At this point, the error occurs👇we can not see the commands menu.
image

# Possible cause
I believe the root cause is that after the plug-in was uninstalled and remounted, no new session was initiated.Therefore the websocket of backend don`t know when should push the new config again.👇
image

xjose97x:
@maoguy Oh, understood. I'll take a look at this case during this week.

hayescode:
@maoguy does #2393 fix this for you?

maoguy:
> @maoguy does #2393 fix this for you?

I try to pull the pr to test，the new commands UI is very nice！👍
But Unfortunately, it still doesn’t solve my problem.😂


https://github.com/user-attachments/assets/7c7ba1ef-949b-47c1-943c-9ca1ec4aa033

xjose97x:
The proposed fix in this PR changes the threadId on every remount. I don't believe this is the right fix.
I'm currently doing some research on this and will provide a PR once I find the root cause.

maoguy:
> The proposed fix in this PR changes the threadId on every remount. I don't believe this is the right fix. I'm currently doing some research on this and will provide a PR once I find the root cause.

I believe the plugin should expose configurable lifecycle hooks—such as a "afterMounted" that fires once the plugin is successfully mounted,or a "beforeUnmounted" hook that is invoked right before the plugin is torn down.

With either of these two lifecycle hooks available, we could call window.clearChainlitCopilotThreadId inside mountedCallback (to start a fresh session after remount) or inside beforeUnmounted (to clean up the current session before the plugin is torn down), ensuring the thread ID is properly reset.

maoguy:
> The proposed fix in this PR changes the threadId on every remount. I don't believe this is the right fix. I'm currently doing some research on this and will provide a PR once I find the root cause.

I think the root cause is that after the browser-side plugin is unmounted and remounted, the server is unaware of the change, so the WebSocket-based set_commands event is never pushed.

xjose97x:
> > The proposed fix in this PR changes the threadId on every remount. I don't believe this is the right fix. I'm currently doing some research on this and will provide a PR once I find the root cause.
> 
> I think the root cause is that after the browser-side plugin is unmounted and remounted, the server is unaware of the change, so the WebSocket-based set_commands event is never pushed.

I believe the correct fix would be to store the set_commands event in the persistence layer—likely by converting it into a Step.
That way, when the widget is remounted and the message history and steps are reloaded, the commands will be restored automatically.

maoguy:
> > > The proposed fix in this PR changes the threadId on every remount. I don't believe this is the right fix. I'm currently doing some research on this and will provide a PR once I find the root cause.
> > 
> > 
> > I think the root cause is that after the browser-side plugin is unmounted and remounted, the server is unaware of the change, so the WebSocket-based set_commands event is never pushed.
> 
> I believe the correct fix would be to store the set_commands event in the persistence layer—likely by converting it into a Step. That way, when the widget is remounted and the message history and steps are reloaded, the commands will be restored automatically.



> > > The proposed fix in this PR changes the threadId on every remount. I don't believe this is the right fix. I'm currently doing some research on this and will provide a PR once I find the root cause.
> > 
> > 
> > I think the root cause is that after the browser-side plugin is unmounted and remounted, the server is unaware of the change, so the WebSocket-based set_commands event is never pushed.
> 
> I believe the correct fix would be to store the set_commands event in the persistence layer—likely by converting it into a Step. That way, when the widget is remounted and the message history and steps are reloaded, the commands will be restored automatically.

Could this approach be interpreted as modifying the react-client layer to persist a copy of the commandsState—a global state managed by Recoil—into localStorage?☺

Issue #895, opened 2024-04-10
Welcome message on_chat_start not appearing in UI


JamesLloydPartingtonSerendipityai:
**Describe the bug**
When using Safari (even on private window), when sending a message in the `on_chat_start` callback the message frequently does not appear (over 90% of the time). Works fine in Chrome. Messages sent in the `on_message` work fine too.

Issue started occurring when we updated from Chainlit 1.0.301 to Chainlit 1.0.502, downgrading fixed the issue.

This issue does not happen when running locally, but does happen when deployed on Kubernetes.

**Desktop:**
- Apple M2 Pro
- Version 13.5, Ventura
- Safari Version 16.6

constantinidan:
Hi @JamesLloydPartingtonSerendipityai can you share a screenshot of the error message?

JamesLloydPartingtonSerendipityai:
Hi @constantinidan no errors appear on the UI, console or backend. Also noticed the settings do not appear when this happens. We configure the settings in `on_chat_start`.

tpatel:
Hi, it might be linked to an Azure issue: https://github.com/microsoft/azure-container-apps/issues/812.

JamesLloydPartingtonSerendipityai:
Hi @tpatel , If that was the issue, then downgrading to a different version of Chainlit would not have fixed the issue, also why would this only occur in Safari? (Using AWS btw)
Looking more at the sockets though I do notice one thing. When running successfully, the `Switching Protocols` occurs only once. However, it always happens twice when the message and settings do not appear.

jimmarshall87:
Hi guys, just to report that @JamesLloydPartingtonSerendipityai and I tracked this down and managed to workaround it, so I thought you would like to hear the details @tpatel @willydouhard @constantinidan.

First, the facts:

- We found we had this issue in Safari on Mac or iPhone but not on Chrome. Safari on Mac we updated to the latest version with no improvement.
- This issue did not occur locally, only when we are deployed on our two pod AWS setup
- We went through version by version and found that the issue was introduced in chainlit 1.0.500 and was not present in prior versions. The issue persists to the latest version of chainlit.
- The behaviour you see is that sometimes your initial opening message in Chainlit is not displayed as James describes above.
- We also occasionally saw a console message that looks like this:

`WebSocket connection to 'wss://<our url>/ws/socket.io/?EIO=4&transport=websocket&sid=-P8qtONaTloNTEknAABY' failed: WebSocket is closed before the connection is established.`

[...]

PR #2390, opened 2025-08-10
fix: disable GZIP compression due to a safari issue

jdb78:
This reverts commit 239a17a0df2802fb4a93f8bbbcb8660c69097c44.

Adding GZIP compression plays badly with websockets. This PR removes it to enable Safari again (including all Safari-based browsers such as all browsers on iPhones).

Citing from release of version 1.0.506

> - disable gzip middleware to prevent a compression issue on safari

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -25,7 +25,6 @@ from fastapi import (
     UploadFile,
     status,
 )
-from fastapi.middleware.gzip import GZipMiddleware
 from fastapi.responses import FileResponse, HTMLResponse, JSONResponse, RedirectResponse
 from fastapi.security import OAuth2PasswordRequestForm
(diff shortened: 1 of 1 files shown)


asvishnyakov:
@jdb78 This sounds more like an issue, than PR, so please create it next time. GZip compression significantly (3.5 times) reduces frontend size and frontend downloading time. Instead of reverting it totally, I would like to include WebSockets (which for me sounds strange, as Starlette already apply GZip compression only to HTTP requests) and `Switching Protocols` requests to exclusions. I'll do it in separate PR a bit later and until then keep this open to highlight the issue for anyone interested.

asvishnyakov:
Related to #895 and #952

asvishnyakov:
Superseded by #2394

Issue #2013, opened 2025-03-17
Third-Party Cookies Blocked in Safari, Preventing Copilot Usage


haidersultancrewlogix:
#### **Description**  
Safari blocks third-party cookies by default, which prevents Copilot from functioning properly. The issue also occurs in Chrome's Private mode and when the "Block Third-Party Cookies" option is enabled in any browser.  

#### **Steps to Reproduce**  

1. Enable authentication on the Chainlit backend.  
2. Embed Copilot on an HTML page.  
3. Open the page in **Safari**, a **Private tab in Chrome**, or any browser with "Block Third-Party Cookies" enabled.  
4. Check the `/jwt` API response.  
5. **Error Message:**  
   _"This attempt to set cookie via Set-Cookie header was blocked due to user preferences."_

#### **Expected Behavior**  
The UI should display a clear message prompting users to **allow third-party cookies** instead of showing a blank screen.  

#### **Actual Behavior**  
- **Blank screen** with no error message or instructions.  
- Error occurs while setting a cookie via the `/jwt` API.  

#### **Screenshots**  
**Blank Screen (No Error Message or Guide):**  
![Image](https://github.com/user-attachments/assets/78db8f00-cb19-440a-9cfd-24d7f58c9ce5)  

**Error while setting cookie:**  
![Image](https://github.com/user-attachments/assets/8d2304b9-709f-4f73-9fd4-fbbe83286f87)  

#### **Environment Details**  

**Desktop:**  
- **OS:** Ubuntu / Windows / macOS  
- **Browser:** Chrome / Firefox / Safari  
- **Version:** All versions  

**Smartphone:**  
- **Device:** Android / iPhone  
- **Browser:** Chrome / Firefox / Safari  
- **Version:** All versions

willydouhard:
Currently all authentication is handled by cookies. Changing this would be a significant change

haidersultancrewlogix:
I think we can still use Cookies for Authentication. When we allow third-party cookies from the browser settings the Copilot starts working. Maybe an error message in UI that can prompt users to allow third-party cookies if they have them blocked?

willydouhard:
That would be great, open to a contribution!

asvishnyakov:
[...]

PR #2394, opened 2025-08-11
fix: disable gzip for websocket-related http endpoints due to bug in Safari

asvishnyakov:
Supersedes #2390

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -226,7 +226,7 @@ class SafariWebSocketsCompatibleGZipMiddleware(GZipMiddleware):
             return await self.app(scope, receive, send)
 
         # Prevent gzip compression for HTTP requests to socket.io path due to a bug in Safari
-        if SOCKET_IO_PATH in scope["path"]:
+        if URL(scope=scope).path.startswith(SOCKET_IO_PATH):
             await self.app(scope, receive, send)
         else:
             await super().__call__(scope, receive, send)
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -222,6 +222,9 @@ app.add_middleware(
 
 class SafariWebSocketsCompatibleGZipMiddleware(GZipMiddleware):
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http":
+            return await self.app(scope, receive, send)
+
         # Prevent gzip compression for HTTP requests to socket.io path due to a bug in Safari
         if SOCKET_IO_PATH in scope["path"]:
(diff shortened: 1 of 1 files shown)


asvishnyakov:
@jdb78 This should work, can you test it in Safari, please?

hayescode (approved):

asvishnyakov:
@hayescode Thank you, but please don't merge until someone with Safari confirms it's working

xjose97x:
> @hayescode Thank you, but please don't merge until someone with Safari confirms it's working

I'll try to test it this week.

backend/chainlit/server.py
```diff
@@ -217,7 +219,17 @@ def get_build_dir(local_target: str, packaged_target: str) -> str:
     allow_headers=["*"],
 )
 
-app.add_middleware(GZipMiddleware)
+
+class SafariWebSocketsCompatibleGZipMiddleware(GZipMiddleware):
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        # Prevent gzip compression for HTTP requests to socket.io path due to a bug in Safari
+        if SOCKET_IO_PATH in scope["path"]:
```
jdb78: You'll need to use `scope.get("path", "")` to make this reliably work. Scope doesn't always have "path".

jdb78 (changes requested):
Thank you! Tiny changes and this should work.

backend/chainlit/server.py
asvishnyakov: @jdb78 I missed `scope["type"] == "http"` check. Also found [this gem](https://www.starlette.io/middleware/#sending-eager-responses) in depths of Starlette documentation

jdb78:
can confirm this fixes the Safari issue

xjose97x:
> can confirm this fixes the Safari issue

Can you send and receive messages?
I'm still having issues on Safari with this PR.

EDIT:
Just checked and for an unknown reason on safari my copilot authentication token is not working. But only on Safari.
Not sure yet the reason.

asvishnyakov:
@xjose97x I think you experience this issue #2013
Unfortunately there is no good solution for it

asvishnyakov:
@xjose97x Is it working without authentication or if copilot is placed on subdomain?

xjose97x:
> @xjose97x I think you experience this issue #2013 Unfortunately there is no good solution for it

You are right. Seems to be that issue.
That's very interesting. Is anyone working on that issue? If not I can take a look at it.

> @xjose97x Is it working without authentication or if copilot is placed on subdomain?

Without auth seems to be working, yes

PR #2400, opened 2025-08-12
fix(ui): prevent login page flicker by limiting full-screen loader to…

### LINKED ISSUE #2399: Login Page continuously blinks/flickers
**Describe the bug**

With password authentication enabled, the login page continuously blinks/flickers until the user submits credentials and is authenticated.

This did not happen in v2.4.3. The behavior appears in v2.4.4+, introduced by the loader overlay added in PR #2068.

**To Reproduce**
1. Create a minimal Chainlit app with password auth, e.g.:
````
import chainlit as cl

@cl.password_auth_callback
async def auth(username: str, password: str):
    if username == "admin" and password == "pass":
        return cl.User(identifier="admin", metadata={"role": "user"})
    return None

@cl.on_chat_start
async def start():
    await cl.Message("Hello").send()
````
2. Ensure CHAINLIT_AUTH_SECRET is set in .env.
3. Run: `chainlit run app.py`
4. Open the app in a browser without being logged in.

**Expected behavior**
The login page should appear stable (without flicker/blink)

**Observed behavior**
The login page repeatedly flashes

**Desktop**

- OS: Windows
- Browser: FireFox, Chrome

Before fix:
[Login page before fix](https://drive.google.com/file/d/1kC24OdW91j4Icz5JDIOT7_MtC7n63MeQ/view?usp=sharing)

After fix:
[Login Page After fix](https://drive.google.com/file/d/1MASzyOlDCpjj6Q3IJPeRrMXYtz7R5pJf/view?usp=sharing)

What actually happens (root cause):
In `frontend/src/App.tsx`, a full-screen overlay is shown whenever `isReady === false:`
```

<div className={cn(
  'bg-[hsl(var(--background))] ... fixed size-full top-0',
  isReady && 'hidden'
)}>
  <Loader className="!size-6" />
</div>
```
- The auth hook (SWR) sets` user = undefined` while `/user` is (re)loading (initial load and revalidations, e.g. on focus).
- Because` isReady` depends on user being defined, `isReady` flips to false during every revalidation, which makes the overlay reappear briefly.
- Result: overlay toggles on/off over the login page → visible flicker.
This is a regression from PR #2068 where the overlay was introduced.

Environment

    Chainlit: v2.4.4+ (not present in v2.4.3)

    OS: Windows 

    Browsers: Chrome, Firefox

    Backend auth: password auth with CHAINLIT_AUTH_SECRET configured

Proposed fix (minimal & safe)
Show the overlay only during initial bootstrap, not during later revalidations:
```
diff --git a/frontend/src/App.tsx b/frontend/src/App.tsx
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from 'react';
+import { useEffect, useState } from 'react';
 // ...imports omitted for brevity

 function App() {
   const { config } = useConfig();
   const { isAuthenticated, data, isReady } = useAuth();
   // ...

+  // Show the overlay only before the app becomes ready the first time
+  const [bootstrapped, setBootstrapped] = useState(false);
+  useEffect(() => {
+    if (isReady) setBootstrapped(true);
+  }, [isReady]);
+
   // ...
-  return (
+  const showBootOverlay = !bootstrapped && !isReady;
+  return (
     <ThemeProvider storageKey="vite-ui-theme" defaultTheme={data?.default_theme}>
       {/* ... */}
       <RouterProvider router={router} />
 
       <div
         className={cn(
           'bg-[hsl(var(--background))] flex items-center justify-center fixed size-full p-2 top-0',
-          isReady && 'hidden'
+          !showBootOverlay && 'hidden'
         )}
       >
         <Loader className="!size-6" />
       </div>
     </ThemeProvider>
   );
 }

```

pk-incospro:
Fixes #2399  (login page flickers until user logs in)

This is a regression introduced by the loader overlay (see PR #2068).

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,9 +1,8 @@
 import { cn } from '@/lib/utils';
-import { useEffect } from 'react';
 import { RouterProvider } from 'react-router-dom';
 import { useRecoilValue } from 'recoil';
 import { router } from 'router';
-
+import { useEffect, useState } from 'react';
 import { useAuth, useChatSession, useConfig } from '@chainlit/react-client';
 
 import ChatSettingsModal from './components/ChatSettings';
@@ -32,6 +31,11 @@ function App() {
   const { connect, chatProfile, setChatProfile } = useChatSession();
 
   const configLoaded = !!config;
+
+  const [bootstrapped, setBootstrapped] = useState(false);
+  useEffect(() => {
+    if (isReady) setBootstrapped(true); // once ready, never show overlay again
+  }, [isReady]);
 
   const chatProfileOk = configLoaded
     ? config.chatProfiles.length
@@ -73,6 +77,8 @@ function App() {
 
   if (!configLoaded && isAuthenticated) return null;
 
+  const showBootOverlay = !bootstrapped && !isReady;
+
   return (
     <ThemeProvider
       storageKey="vite-ui-theme"
@@ -86,7 +92,7 @@ function App() {
       <div
         className={cn(
           'bg-[hsl(var(--background))] flex items-center justify-center fixed size-full p-2 top-0',
-          isReady && 'hidden'
+          !showBootOverlay && 'hidden'
         )}
       >
         <Loader className="!size-6" />


asvishnyakov:
@pk-incospro Is this related to #2205?

actValora-Pk:
> @pk-incospro Is this related to #2205?
Yes related,

 #2205 fixes the auth-state behavior: it stops setting user to undefined during `/user` revalidation (especially on 401 status), which prevents the periodic refresh/spinner.

 #2399 scopes the full-screen loader in App.tsx to initial bootstrap only, so even if isReady flips later, the overlay won’t flash over the login page.

asvishnyakov (dismissed):

Issue #1502, opened 2024-11-06
Feature request: dynamic configuration management


brucechou1983:
**Is your feature request related to a problem? Please describe.**

Currently, chainlit uses a single config.toml file to manage settings across all ChatProfiles. This creates limitations in providing customized user experiences for different workflows. Specific pain points include:

- All ChatProfiles must share the same UI configuration, regardless of their specific requirements
- Cannot disable/enable features based on the selected profile's capabilities (e.g., image upload for non-vision models)
- No way to dynamically adjust settings when switching between profiles

**Describe the solution you'd like**

Programmatic configuration settings. Something like

```
@cl.on_profile_switch
async def handle_profile_switch(profile: ChatProfile):
    # Update configuration for specific profile
    await profile.update_config({
        "sector": {
            "key1": value1,
            "key2": value2
        }
    })
```

**Describe alternatives you've considered**
Separate Config Files: Create individual config files for each profile.
- Pros: Simple to implement, clear separation.
- Cons: Duplication of common settings, harder to maintain, less control than api.

**Additional context**
I found this issue when building [chainlit langgraph integration](https://github.com/brucechou1983/chainlit_langgraph) for creating agentic chatbot.

dokterbob:
Great point!

Having dynamic configuration management would greatly help with testing as well.

To move forward, I would suggest splitting this issue though:
1. Ensuring that he configuration object is dynamic (e.g. all logic relying on it is stateless).
2. Allow overriding of app init, including loading configuration explicitly -- allowing programmatic initialisation of the configuration.
3. Allowing updating of the configuration from within chat profiles.

If you (or others!) would like to help us move forward, feel free to create separate issues for all of these, linking them to each other (e.g. 2 and 3 both rely on 1) and here, perhaps making a go at spec'ing them. Once these issues are created, we can close this one.

All of this should have at least ~80% unittest coverage and will definitely be required to increase the level of production-readiness of chainlit.

dokterbob:
@dosu Perhaps you could help us make a start scoping those issues based on our issue/feature template?

brucechou1983:
@dokterbob I will find time to review the current implementation details, provide a reasonable implementation plan, and create sub-issue tickets.

aharth:
+1 to "2. Allow overriding of app init".

Having something like `@cl.on_app_startup` would be great to be able to load models or initialise database connections.

I tried a bunch of different approaches already, including to override the FastAPI lifespan() function.

[...]

Issue #810, opened 2024-03-14
Allow dynamic adjustment of chainlit ChatSettings


tituslhy:
Hello!

**Is your feature request related to a problem? Please describe.**
I think that most people use the chainlit ChatSettings (https://docs.chainlit.io/api-reference/chat-settings) tool to set LLM model related settings such as temperature. 

But I've been using the ChatSettings tool as a channel directory for the user to choose which data source they'd like to talk to. Some users might want to refer to administrative manuals while some users want to talk to their own uploaded documents. I have different llamaindex query engines over each different data source. I deliberately designed my chainlit app this way so that there is no data contamination. 

**Describe the solution you'd like**
Users now have to toggle between each chat setting manually to choose their data source. But actually we can adjust the data source on the backend using the cl.user_session variable. So when I detect that the user has uploaded a document with the message, I'll automatically change the document source to the "Uploaded Documents" channel (i.e. cl.user_session.set("document_source") = "uploaded documents"). 

However there is currently no way to automatically port this change to the Chainlit ChatSettings tool - i.e. the ChatSettings does not automatically update itself to show that it's talking to uploaded documents. This is purely optics of course, the users can still just manually adjust it themselves.

**Describe alternatives you've considered**
At the start of every LLM response, I get Chainlit to inform the user via a chainlit Message object on which document source it's constructing the answer from. So that the end user can manually toggle the app to point at a different data source.

I can also prevent any reading of uploaded documents if the user does not toggle to the "Uploaded Documents" channel but I decided against this because Chainlit has an upload button with the message bar by default.

Valdanitooooo:
I found that resending `ChatSettings` when switching `ChatProfiles` can achieve the effect of dynamically switching configurations
```   
    if profile == "test01":
        settings = await cl.ChatSettings(
            [
                Select(
                    id="xxx",
                    label="xxx",
                    values=[
                        "xx01",
                        "xx02"
                    ],
                    initial_index=0,
                )
        ).send()
    else:
        await cl.ChatSettings([]).send()
```

PR #2401, opened 2025-08-12
feat: Add Dynamic Configuration to ChatProfiles

hayescode:
## Changes:
- Audio.enabled and MCP.enabled now must be explicitly enabled. Before it was inferred if on_audio_chunk and on_mcp_connect existed. This enables selective enablement by ChatProfile now.
- Documentation Update PR: https://github.com/Chainlit/docs/pull/256

## ✅ Backend Implementation Complete
- ChainlitConfigOverrides class in config.py - allows specifying UI, features, and project overrides
- ChatProfile.config_overrides field - stores profile-specific configuration overrides
- /project/settings endpoint enhancement - applies config overrides based on the chat_profile parameter
- Proper serialization - removes config_overrides from profiles sent to frontend (they're applied server-side)
- Pydantic dataclass rebuild - fixes forward reference issues

## ✅ Frontend Implementation Complete
- Enhanced useConfig hook - includes chat_profile parameter in API calls
- Automatic re-fetching - clears config when chat profile changes to trigger re-fetch with new profile-specific settings

## 🎯 How It Works
- Default state: Frontend calls /project/settings?language=en-US → gets base configuration
- Profile selection: User selects "vision-assistant" → chatProfileState updates
- Config refresh: Frontend clears config and calls /project/settings?language=en-US&chat_profile=vision-assistant
- Server merges: Backend finds vision-assistant profile, applies its config_overrides to base config
- Frontend receives: Merged configuration with vision-assistant's UI name, file upload settings, etc.

## Other:
- Added "UP007" to ruff ignore because it was flagging these when using Optional[str] which is valid.
image

Fixes https://github.com/Chainlit/chainlit/issues/1502

### DIFF
--- a/.github/workflows/copilot-setup-steps.yaml
+++ b/.github/workflows/copilot-setup-steps.yaml
@@ -1,0 +1,49 @@
+name: "Copilot Setup Steps"
+
+# Automatically run the setup steps when they are changed to allow for easy validation, and
+# allow manual testing through the repository's "Actions" tab
+# This workflow optimizes the GitHub Copilot coding agent's ephemeral development environment
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - .github/workflows/copilot-setup-steps.yaml
+  pull_request:
+    paths:
+      - .github/workflows/copilot-setup-steps.yaml
+
+jobs:
+  # The job MUST be called `copilot-setup-steps` or it will not be picked up by Copilot.
+  copilot-setup-steps:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    # Set the permissions to the lowest permissions possible needed for your steps.
+    # Copilot will be given its own token for its operations.
+    permissions:
+      # If you want to clone the repository as part of your setup steps, for example to install dependencies, you'll need the `contents: read` permission. If you don't clone the repository in your setup steps, Copilot will do this for you automatically after the steps complete.
+      contents: read
+
+    # You can define any steps you want, and they will run before the agent starts.
+    # If you do not check out your code, Copilot will do this for you.
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install Node.js, pnpm and dependencies
+        uses: ./.github/actions/pnpm-node-install
+        with:
+          node-version: "24.3.0"
+          pnpm-install-args: "--frozen-lockfile"
+
+      - name: Install Python, Poetry and dependencies
+        uses: ./.github/actions/poetry-python-install
+        with:
+          python-version: "3.10"
+          poetry-version: "2.1.3"
+          poetry-working-directory: "./backend"
+          poetry-install-args: "--with tests --with mypy --with dev --with custom-data --no-root"
+
+      - name: Build UI components
+        run: pnpm run buildUi
+        timeout-minutes: 5


hayescode:
Docs updated in this PR: https://github.com/Chainlit/docs/pull/256

backend/chainlit/types.py
```diff
@@ -293,6 +295,7 @@ class ChatProfile(DataClassJsonMixin):
     icon: Optional[str] = None
     default: bool = False
     starters: Optional[List[Starter]] = None
+    config_overrides: Any = None
```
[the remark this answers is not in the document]
hayescode: before of future annotations doing this throws an error.

backend/tests/test_server.py
```diff
@@ -687,3 +686,158 @@ def test_project_translations_file_path_traversal(
...
+        McpFeature,
+        UISettings,
+    )
+    from chainlit.types import ChatProfile
+
+    # Mock chat profiles with different config overrides
+    mock_profiles = [
+        ChatProfile(
+            name="basic",
+            markdown_description="Basic profile without overrides",
```
sandangel, line 709: I found an issue. This markdown description does not take into account the "language" param in `/project/settings` right? 

Currently, `/project/settings` endpoint will return the description as is, and the application does not know which language user is using until user send a GET request to `/project/settings` with `language` query params. Is there a way to pass `language` information to the `set_chat_profile` ?

hayescode: I didn't touch that part. Whatever the behavior was before is the same after.

asvishnyakov: @sandangel Only if you introduce `language` parameter to `config.code.set_chat_profiles` which is a great catch, but out-of-scope for this PR

sandangel: Yeah that is true.

backend/chainlit/types.py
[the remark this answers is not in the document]
asvishnyakov: Created #2410

cypress/e2e/config_overrides/spec.cy.ts
```diff
@@ -0,0 +1,140 @@
...
+describe('Config overrides with chat profiles', () => {
+  it('should be able to select a chat profile and test MCP button visibility', () => {
+    cy.visit('/');
+    cy.get("input[name='email']").type('admin');
+    cy.get("input[name='password']").type('admin');
+    cy.get("button[type='submit']").click();
+    cy.get('#chat-input').should('exist');
+
+    // Wait for the interface to be ready
+    cy.get('#starter-default-chat', { timeout: 10000 }).should('exist').click();
```
asvishnyakov: I believe timeouts and `cy.wait` are the same and may cause same issues. Check `data_layer` to see example how to handle situations like this: [check for `location` instead](https://github.com/Chainlit/chainlit/blob/e0bf2a51d9df6b0bd7c61c4417b609bac6283b57/cypress/e2e/data_layer/spec.cy.ts#L41) and then check AI message / element / etc., if you use it

backend/chainlit/server.py
```diff
@@ -777,42 +777,90 @@ async def project_settings(
...
+                            else:
+                                base[key] = value
+
+                    deep_merge(features_config, features_overrides)
+
+                # Apply project overrides
+                if overrides.project:
+                    project_overrides = overrides.project.model_dump(exclude_none=True)
+                    if "user_env" in project_overrides:
+                        project_config["userEnv"] = project_overrides["user_env"]
```
asvishnyakov: I believe [`mode_copy(update: here_your_updated_data, deep: True)`](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_copy) can replace all of this

hayescode: @asvishnyakov good idea! It can replace a lot of this. I pushed an update for this and also added some try/except to be safe. Great suggestion!

cypress/e2e/config_overrides/spec.cy.ts
hayescode: good idea i switched them to cy.location

Valdanitooooo:
This might be the greatest PR in this repository’s history !!!

So, is it possible to add Dynamic ChatSettings Configuration to ChatProfiles ?

https://github.com/Chainlit/chainlit/issues/810

hayescode:
@Valdanitooooo https://docs.chainlit.io/api-reference/chat-profiles#dynamic-configuration

Issue #2020, opened 2025-03-19
Sort threads by last updated rather than creation date


mihidumh:
**Describe the bug**
After resuming and chatting with a resumed thread, the date in the thread list does not change.

**To Reproduce**
Resume old thread, chat with old thread.

**Expected behavior**
The resumed thread with new activity should go to the top of the threads list with a more recent last update time, rather than creation time.

PR #2404, opened 2025-08-14
fix: changed thread sorting to use updated time instead of creation time

jasparm:
fix for issue #2020.
- changed list_threads to sort by `updatedAt` instead of `createdAt`
- changed update_thread to update `updatedAt` field to datetime.now() for every update


hayescode:
@jasparm can you update all of the other data layers too, so this change is consistent for all users?

asvishnyakov:
@hayescode I believe not all data layers has this field in first place, and that's why it wasn't done before

hayescode:
@asvishnyakov SQLAlchemyDataLayer has it, at least. Good to know though. All the more reason to consolidate!

jasparm:
@hayescode I have updated the SQLAlchemyDataLayer, and it seems that DynamoDB is using "createdAt" as the latest updated timestamp. It gets overwritten to current ts at every call of update_thread.

hayescode (approved):

Issue #2178, opened 2025-05-19
Azure Storage - Resumed Thread Missing Filename


mihidumh:
**Describe the bug**
For resumed threads, any file attachments do not download with a file name but just the blob name with no extension.

**To Reproduce**
Start a thread with a file element.
Resume this thread and try to download the file.

**Expected behavior**
The file should download with the correct file name and file extension.

**Screenshots**

Chainlit: 2.5.5

mihidumh:
This is still an issue,

PR #2405, opened 2025-08-14
fix: added content-disposition metadata to azure blob uploads to persist download file name

jasparm:
fix for #2178 

- added content_disposition var for AzureBlobStorageClient upload_file function

This means that downloading files stored in a thread after a reload, when they are not in local storage *should* keep the filename.


hayescode:
@jasparm like with #2404 you're only updating 1/4 data layers. Please make the changes to all data layers so we can maintain consistent behavior.

We aim to consolidate data layers in the future, but that's a large effort and will take some time.

jasparm:
@hayescode Both other data layers are using the file name as object key, so will persist correct file name on reload. This just aligns behaviour of azure to the same.

jasparm:
I can add content-disposition metadata to the other Azure layer (I forgot about it), and AWS, but GCP doesn't support it AFAIK from the docs.

backend/chainlit/data/chainlit_data_layer.py
```diff
@@ -9,6 +9,7 @@
 import aiofiles
 import asyncpg  # type: ignore
 
+from backend.chainlit.data.storage_clients.azure_blob import AzureBlobStorageClient
```
hayescode: @jasparm This should just be `from Chainlit...` no need for backend because the project is built from the backend directory.

jasparm: done

backend/chainlit/data/storage_clients/azure.py
```diff
@@ -61,12 +61,13 @@ async def upload_file(
         data: Union[bytes, str],
         mime: str = "application/octet-stream",
         overwrite: bool = True,
+        content_disposition: str = "",
```
hayescode: Should be optional and default to None, right?

backend/chainlit/data/sql_alchemy.py
```diff
@@ -560,16 +560,26 @@ async def get_all_user_threads(
...
+            WHERE t."userId" = :user_id OR t."id" = :thread_id
+            GROUP BY
+                t."id",
+                t."createdAt",
+                t."name",
+                t."userId",
+                t."userIdentifier",
+                t."tags",
+                t."metadata"
+            ORDER BY updatedAt DESC NULLS LAST
```
hayescode: I think you accidentally copied your other PRs code change here. Remove?

jasparm: yes. sorry not sure how to remove the commit from github.

backend/chainlit/data/storage_clients/azure.py
jasparm: yes, default to empty string which will behave same as currently behaves. otherwise if filename set correctly that will override the filename when downloading.

hayescode: But passing in an empty string may set content-disposition to an empty string instead of unset like before. I'm worried about backwards compatibility.

backend/chainlit/data/sql_alchemy.py
hayescode: Just undo this in your branch and submit a new commit

backend/chainlit/data/storage_clients/azure.py
jasparm: ok fixed. defaults to None type now.

backend/chainlit/data/sql_alchemy.py
```diff
@@ -570,7 +570,6 @@ async def get_all_user_threads(
             FROM threads
             WHERE "userId" = :user_id OR "id" = :thread_id
             ORDER BY "createdAt" DESC
-            LIMIT :limit
```
hayescode: Add this back please

backend/chainlit/data/chainlit_data_layer.py
```diff
@@ -199,12 +200,21 @@ async def create_element(self, element: "Element"):
...
+                    overwrite=True,
+                    content_disposition=f"attachment; filename=\"{element.name}\"",
+                )
+            else:
+                await self.storage_client.upload_file(
+                    object_key=path,
+                    data=content,
+                    mime=element.mime or "application/octet-stream",
+                    overwrite=True,
+                )
```
hayescode:
```suggestion
            content_disposition = f"attachment; filename=\"{element.name}\"" if not isinstance(self.storage_client, GCSStorageClient) else None
            await self.storage_client.upload_file(
                object_key=path,
                data=content,
                mime=element.mime or "application/octet-stream",
                overwrite=True,
                content_disposition=content_disposition,
                )
```

hayescode (commented):
@jasparm everything else looks good!

backend/chainlit/data/sql_alchemy.py
jasparm: oops. done

hayescode:
@jasparm you have some mypy errors

hayescode (approved):

PR #2417, opened 2025-08-19
Add missing headers when connecting Streamable HTTP MCP

ggirou:
Headers are missing in the payload sent to the server

### DIFF
--- a/frontend/src/api/index.ts
+++ b/frontend/src/api/index.ts
@@ -23,7 +23,12 @@ const onError = (error: ClientError) => {
 };
 
 class ExtendedChainlitAPI extends ChainlitAPI {
-  connectStreamableHttpMCP(sessionId: string, name: string, url: string, headers: any) {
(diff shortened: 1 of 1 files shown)


hayescode:
@ggirou can you please add tests, especially e2e, for this

ggirou:
Hi @hayescode,

Thank you for your feedback! I intended to provide a quick fix while debugging an issue I encountered when adding an MCP with headers. However, I may not have much time to dive into writing tests, especially e2e, since this is my first contribution to Chainlit.

If you could provide some guidance or pointers on how to get started with the testing framework, I would be more than happy to give it a try.

Thanks for your understanding!

hayescode:
I use the existing tests as a baseline and customize. I just checked and we don't have good coverage on MCP which isn't great but that's another issue. Given that and the optionality here I think we're good, but for next time that's how I add tests.

hayescode (approved):

PR #2427, opened 2025-08-21
fix: revert breaking change of custom auth

sandangel:


xjose97x:
@sandangel Hello San! Did that change cause any issues in your solution?

sandangel:
we are using custom auth with that chainlit custom auth env var without requiring auth secret. the previous change broke that set up

hayescode (approved):

xjose97x:
> we are using custom auth with that chainlit custom auth env var without requiring auth secret. the previous change broke that set up

oh that's quite interesting. but are you using the auth callbacks?
could you please provide more detail on how your authentication implementation works?

thanks San!

Issue #2328, opened 2025-07-28
Support for shadcn/ui Charts in Chainlit CustomElements


anandwilbert:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I want to create visually appealing data visualizations in Chainlit apps but am limited by the lack of native support for modern, customizable chart components. Specifically, integrating interactive and aesthetically pleasing area charts, like those provided by shadcn/ui (built with Recharts), is challenging without direct support in Chainlit's CustomElements. This makes it harder to build engaging, data-driven applications that meet modern UI/UX standards.

**Describe the solution you'd like**
I’d like Chainlit to add support for shadcn/ui’s Chart component within the CustomElements framework. This would allow developers to easily integrate and configure area charts with features like grids, axes, tooltips, and legends directly in Chainlit apps. The implementation should enable seamless copy-pasting of shadcn/ui’s Area Chart code, with Chainlit handling the React/Recharts dependencies and providing a simple API to pass data and customize chart properties (e.g., colors, animations, responsiveness) through CustomElements. This would streamline the creation of visually appealing, interactive data visualizations in conversational AI applications.

**Describe alternatives you've considered**
I’ve considered manually integrating Recharts or other charting libraries into Chainlit apps by building custom components from scratch. However, this is time-consuming, requires managing dependencies, and often results in inconsistent styling with the rest of the app. Another alternative is using existing Chainlit components or external iframe-based solutions, but these lack the flexibility, accessibility, and modern design of shadcn/ui’s Area Chart, making it harder to achieve a polished, cohesive look.

**Additional context**
The shadcn/ui Area Chart, built with Recharts, is highly customizable and supports features like responsive design, tooltips, legends, and grid lines, which are ideal for creating data-rich, visually appealing dashboards in Chainlit apps. Adding this to CustomElements would align with Chainlit’s goal of enabling rapid development of conversational AI interfaces while enhancing the ability to present data insights attractively. For example, developers could use these charts to visualize trends or metrics in real-time within a Chainlit UI, improving user engagement. Screenshots of shadcn/ui’s Area Chart examples can be found at https://ui.shadcn.com/charts/area for reference

josca42:
As far as I can tell this can be achieved by simple doing updating frontend/src/components/Elements/CustomElement/Imports.ts to include:

import * as Recharts from 'recharts';
import { ChartContainer, ChartTooltip, ... } from '@/components/ui/chart';

[...]

PR #2428, opened 2025-08-21
Make custom elements a first class citizen

josca42:
Quickly creating beautiful UI elements in claude and adding them as custom elements in chainlit enables creating beautiful UX flows that differentiate from the just another chat bot experience. A quick example of this is shown in this [video](https://www.loom.com/share/ebb2e8da387140a8973cc5a5f24e6a0f?sid=8f3d679b-c09e-4d29-a2c1-70a9fb3858d0) .

Claude natively uses shadcn and recharts so everything almost work out of the box but a few key changes are needed. More tailwind styling needs to be made available to the custom elements.
And since a lot of chainlit users will likely be using it related to analysis workflows - preferring chainlit to streamlit etc. - then the charting functionality used by claude should be available. This means recharts and the shadcn chart component.

This also addresses #2328

### DIFF
--- a/frontend/src/components/ui/chart.tsx
+++ b/frontend/src/components/ui/chart.tsx
@@ -1,0 +1,362 @@
+import { cn } from '@/lib/utils';
+import * as React from 'react';
+import * as RechartsPrimitive from 'recharts';
+
+// Format: { THEME_NAME: CSS_SELECTOR }
+const THEMES = { light: '', dark: '.dark' } as const;
+
+export type ChartConfig = {
+  [k in string]: {
+    label?: React.ReactNode;
+    icon?: React.ComponentType;
+  } & (
+    | { color?: string; theme?: never }
+    | { color?: never; theme: Record<keyof typeof THEMES, string> }
+  );
+};
+
+type ChartContextProps = {
+  config: ChartConfig;
+};
+
+const ChartContext = React.createContext<ChartContextProps | null>(null);
+
+function useChart() {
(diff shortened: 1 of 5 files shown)


josca42:
As promised @asvishnyakov

josca42:
If these changes are something you would like to add then I'd be happy to create some good docs on how to use this kind of workflow to quickly create custom UI elements that you can integrate in your chatbot :-).

hayescode:
Thanks for this contribution @josca42 this looks very promising! Do you have some examples code to run and see how this works?

Docs/Cookbook will be very helpful for the community once this is released for sure.

frontend/src/components/ui/chart.tsx
```diff
@@ -0,0 +1,362 @@
...
+            ([theme, prefix]) => `
+${prefix} [data-chart=${id}] {
+${colorConfig
+  .map(([key, itemConfig]) => {
+    const color =
+      itemConfig.theme?.[theme as keyof typeof itemConfig.theme] ||
+      itemConfig.color;
+    return color ? `  --color-${key}: ${color};` : null;
+  })
+  .join('\n')}
```
sandangel, line 90: hi, why do we need dangerouslySetInnerHTML? can it be set from style attribute instead of dangerouslySetInnerHTML

sandangel:
some example usage would be great. currently I'm not sure how to use this component.

sergiossm:
any update on this?

PR #2415, opened 2025-08-18
feat: Migrate from poetry to uv

hayescode:
supersedes #1995  

This pull request migrates the project from Poetry to uv for Python dependency management and updates the build system to use hatchling instead of poetry-core. All related documentation, scripts, and GitHub workflows are updated to reflect this change, ensuring consistency across development, testing, and deployment processes.

**Dependency Management Migration**

* Migrated from Poetry to uv for Python dependency management, including updating `backend/pyproject.toml` to use PEP 621 `[project]` metadata and hatchling as the build backend. All dependencies and scripts are now specified in uv/hatchling format. [[1]](diffhunk://#diff-9f2cf60079756e94b31dbaeb145297215236d3c0135647163c2e51b80fd81551L1-R23) [[2]](diffhunk://#diff-9f2cf60079756e94b31dbaeb145297215236d3c0135647163c2e51b80fd81551L30-L109) [[3]](diffhunk://#diff-9f2cf60079756e94b31dbaeb145297215236d3c0135647163c2e51b80fd81551L131-L148) [[4]](diffhunk://#diff-8225802a8ab008873adfd5038f50faed02d85494d19a14e1706639cb2dd51866L1-R1) [[5]](diffhunk://#diff-8225802a8ab008873adfd5038f50faed02d85494d19a14e1706639cb2dd51866L73-R73)
* Updated build system from poetry-core to hatchling, including build configuration and package inclusion rules. [[1]](diffhunk://#diff-9f2cf60079756e94b31dbaeb145297215236d3c0135647163c2e51b80fd81551L1-R23) [[2]](diffhunk://#diff-9f2cf60079756e94b31dbaeb145297215236d3c0135647163c2e51b80fd81551L30-L109) [[3]](diffhunk://#diff-9f2cf60079756e94b31dbaeb145297215236d3c0135647163c2e51b80fd81551L131-L148)

**Workflow and Automation Updates**

[...]

PR #2430, opened 2025-08-21
fix: include frontend assets as artifacts

sandangel:
`include` does not include files that is ignored by vcs
should use `artifacts` instead: https://hatch.pypa.io/latest/config/build/#artifacts


asvishnyakov:
Related to #2415

PR #2435, opened 2025-08-22
feat: Dynamic Config Improvements

hayescode:


backend/chainlit/config.py
```diff
@@ -407,16 +411,23 @@ class ChainlitConfigOverrides(BaseModel):
...
 
-class ChainlitConfig(BaseModel):
-    model_config = ConfigDict(
-        arbitrary_types_allowed=True, revalidate_instances="always"
+class ChainlitConfig(BaseSettings):
+    model_config = SettingsConfigDict(
+        env_prefix="CHAINLIT_",
+        env_file=(".env",),
+        env_file_encoding="utf-8",
+        env_nested_delimiter="__",
```
sandangel: Will this create breaking changes?

backend/chainlit/config.py
```diff
@@ -407,16 +411,23 @@ class ChainlitConfigOverrides(BaseModel):
...
+        extra="ignore",
+        arbitrary_types_allowed=True,
     )
 
-    # Directory where the Chainlit project is located
     root: str = APP_ROOT
-    # Chainlit server URL. Used only for cloud features
-    chainlit_server: str
-    run: RunSettings
+    chainlit_server: str = Field(
```
sandangel: If we no longer have cloud feautures, do we still need this?

backend/chainlit/config.py
hayescode: @sandangel I got this from the [docs](https://docs.pydantic.dev/latest/concepts/pydantic_settings/). It's moreso for us to know we can configure and leverage pydantic's validation _in the future_. You're right though, i'll remove this since we aren't using this.

backend/chainlit/config.py
hayescode: yep good catch, removed!

Issue #2441, opened 2025-08-25
When accessing the Chainlit web UI over an insecure context (e.g., HTTP), content copying fails.


maoguy:
**Describe the bug**
When accessing the Chainlit web UI over an insecure context (e.g., HTTP), content copying fails.

**To Reproduce**

1. use the http mode to view chainlit;
2. use the copyButton to copy content.

**Expected behavior**
We want the copy button to remain fully functional even when accessed over an insecure context (e.g., HTTP).

**Screenshots**
<img width="708" height="521" alt="Image" src="https://github.com/user-attachments/assets/c99f2a08-1bef-4874-80cd-942dd549b475" />

**Desktop (please complete the following information):**

- OS: [windows11]
- Browser [chrome]
- Version [137.0.7151.120]

maoguy:
# Here is a feasible solution.

1. setup the "custom_js" config of ".chainlit\config.toml" (ref: https://docs.chainlit.io/customisation/custom-js#js);
2. Insert the following polyfill code snippet into the JavaScript file corresponding to the `custom_js` configuration item.

```javascript
/* clipboard-polyfill */
(function (global) {
  console.log("polyfill")
  /* ---------- ClipboardItemPolyfill ---------- */
  function ClipboardItemPolyfill(items, options) {
    options = options || {};

    if (typeof items !== 'object' || items === null) {
      throw new TypeError('Failed to construct "ClipboardItem": parameter 1 is not an object');
    }

    this._items = {};
    this._types = [];

    for (var mimeType in items) {
      if (!items.hasOwnProperty(mimeType)) continue;
      var blob = items[mimeType];
      if (!(blob instanceof Blob)) {
        throw new TypeError(
          'Failed to construct "ClipboardItem": Failed to read the \'' +
          mimeType +
          '\' property from ClipboardItemData: The provided value is not of type \'Blob\''
        );
      }
      this._items[mimeType] = blob;
      this._types.push(mimeType);
    }

    this._options = options;
  }

  Object.defineProperty(ClipboardItemPolyfill.prototype, 'types', {
    get: function () {
      return Object.freeze(this._types.slice());
    }
  });

  ClipboardItemPolyfill.prototype.getType = function (type) {
    var self = this;
    return new Promise(function (resolve, reject) {
      if (!self._items[type]) {
        reject(new DOMException('The type \'' + type + '\' is not supported', 'NotFoundError'));
        return;
      }
      resolve(self._items[type]);
    });
  };

[...]

PR #2443, opened 2025-08-25
Use clipboard-polyfill in the CopyButton component.

maoguy:
Which can solve the https://github.com/Chainlit/chainlit/issues/2441

image


sandangel:
can you share why this polyfill would help?

maoguy:
> can you share why this polyfill would help?

when you use chainlit web client in http mode,the polyfill will work.(more detail see the issue)

sandangel:
hi @maoguy  thanks for the explanation. Could you please help add the reference source to the polyfill or is this something you have written yourself? usually for such polyfill I believe there should be some existing packages that we can just import?

maoguy:
> hi @maoguy  thanks for the explanation. Could you please help add the reference source to the polyfill or is this something you have written yourself? usually for such polyfill I believe there should be some existing packages that we can just import? 

Previously, I looked for polyfills related to ClipboardItem but couldn't find any npm packages offering full support for its functionality. So, I implemented my own version.

sandangel:
oh right, we have custom.js so I think we do not need this as part of chainlit

maoguy:
ok know that

maoguy:
> oh right, we have custom.js so I think we do not need this as part of chainlit

Here is the corresponding proposal document I’ve prepared; please review it.👇

https://github.com/Chainlit/chainlit/issues/2441#issuecomment-3252547274

PR #2456, opened 2025-08-26
Add a second optional language parameter to set_starters

jfperusse:
This changes adds an optional language parameter for set_starters callbacks, in addition to the currently optional User parameter.

This allows for programmatically localizing the label and prompts associated with the starters.

Includes a test to confirm the feature works as expected.


sandangel:
Hi, Thanks for the contribution. I wonder if it's okay for you to do the same for set_profile in this PR or would you prefer having it in a separate PR?

backend/chainlit/callbacks.py
```diff
@@ -240,17 +240,27 @@ def set_chat_profiles(
...
 def set_starters(
     func: Callable[[Optional["User"]], Awaitable[List["Starter"]]],
-) -> Callable:
+) -> Callable[[Optional["User"]], Awaitable[List["Starter"]]]: ...
+
+
+@overload
+def set_starters(
+    func: Callable[[Optional["User"], Optional["str"]], Awaitable[List["Starter"]]],
+) -> Callable: ...
```
sandangel:
```suggestion
) -> Callable[[Optional["User"], Optional["str"]], Awaitable[List["Starter"]]]: ...
```

jfperusse:
> Hi, Thanks for the contribution. I wonder if it's okay for you to do the same for set_profile in this PR or would you prefer having it in a separate PR?

@sandangel Had a quick look at it. Most of the change would be trivial, except a critical part where I'm not exactly sure.

In `backend/chainlit/session.py`, for WebsocketSession.get_config(), we don't have the language in the same way as through the routes, but set_chat_profiles is called to see if the profile defines a config_overrides. Without a way to make sure we use the same language as /project/settings, that functionnality would break with localized profile names.

We could parse the first part of the HTTP_ACCEPT_LANGUAGE we have in the environ variable using a regex, but I'm not sure this would guarantee it's the same language as the one passed in the routes?

jfperusse:
@sandangel For better clarity, changes for set_chat_profiles are in a separate PR at https://github.com/Chainlit/chainlit/pull/2461.

sandangel (approved):
lgtm

PR #2460, opened 2025-08-27
fix: data persistence bug introduced in #2433

mihidumh:
The Type Checking fix from #2433 introduced a new issue that causes an error while persisting to the database.
This PR should fix that while passing mypy checks. 

    if not isinstance(self.storage_client, GCSStorageClient)
                                           ^^^^^^^^^^^^^^^^
NameError: name 'GCSStorageClient' is not defined. Did you mean: 'BaseStorageClient'?


sandangel:
question: if we use client storage protocol, why do we still have to import the concrete class here?

mihidumh:
@sandangel I believe you'll find your answer in #2405

hayescode (approved):

PR #2463, opened 2025-08-28
Threadsorting fix

jasparm:
My previous PR did not fully fix the thread sorting to go off updatedAt for the default data layer. This is a basic change to fix this.


sandangel (approved):
lgtm

PR #2480, opened 2025-09-02
fix: missing / in url with base path when connecting Streamable HTTP MCP

ggirou:
In example, with `CHAINLIT_ROOT_PATH=/mypath`

### DIFF
--- a/frontend/src/api/index.ts
+++ b/frontend/src/api/index.ts
@@ -30,7 +30,7 @@ class ExtendedChainlitAPI extends ChainlitAPI {
     headers?: Record<string, string>
   ) {
     // Assumes the backend expects { clientType, name, url }
-    return fetch(`${this.httpEndpoint}/mcp`, {
+    return fetch(new URL("mcp", this.httpEndpoint.endsWith("/") ? this.httpEndpoint : `${this.httpEndpoint}/`), {
       method: 'POST',
       headers: {
         'Content-Type': 'application/json',


asvishnyakov:
I'm begging everyone to stop manually build urls and start using `URL` for that

ggirou:
Hi @asvishnyakov ,

Thank you for your feedback. I understand the importance of using the `URL` instead of manually building URLs. However, as I'm still getting familiar with the codebase, I would appreciate any guidance or examples you could provide on how to implement this properly.

Your help would be invaluable as I work on making these changes.

Thanks in advance!

asvishnyakov:
@ggirou

```
new URL(path, base)
```
or, if base itself may already contain path:
```
new URL(path, base.endsWith("/") ? base : `${base}/`)
```
if you need to keep path or
```
new URL(path, base)
```
if you need to replace path

Chainlit already has `buildEndpoint` [in API class](https://github.com/Chainlit/chainlit/blob/0e863aee98c030b85d67e0d9ed8ad6f532967820/libs/react-client/src/api/index.tsx#L54) which we may use, but to be honest I would like to refactor it a bit

ggirou:
Hi @asvishnyakov

I updated the PR to use `URL`

cornmail:
Hi @asvishnyakov, @asvishnyakov, @sandangel

Do you have any suggestions or feedback for merging my pull request?

Thanks!

sandangel:
Hi, I believe when the client miss the end `/` , server will redirect it to that `/`, is it correct? Either case should work without issue.

mukul1em:
is there any update on this PR?

asvishnyakov:
I'll merge it after next release

hayescode (approved):

PR #2481, opened 2025-09-02
fix(hot-reload): preserve run.module_name on config reload and safely…

DmitriyAlergant:
**Summary**

- Fixes an issue with chat profiles (and other decorator-registered callbacks) disappearing after hot reload (--watch)
- Root cause: reload_config() replaced config.run with a fresh RunSettings() which has module_name=None. The watcher then skipped load_module(), so decorators were never re-executed.
- Change: Preserve and restore run.module_name during reload_config(). Also safely iterate sys.modules while force-refreshing to avoid mutation errors.

**Reproduction of the original issue (fixed with this CR)**

1. Start dev server with watch (e.g., chainlit run app.py --watch), where an app implements @cl.set_chat_profiles decorator

2. UI shows chat profiles selector initially

3. Change anything in app.py and observe hot reload

3. UI refreshes but chat profiles selector no longer visible; GET /project/settings did not return chat profiles (now fixed)


DmitriyAlergant:
@hayescode thanks for looking into this.  I've added a commit to fix ruff formatting.

PR #2489, opened 2025-09-04
[WIP] - Add Thread Sharing

hayescode:
Work in progress...


backend/chainlit/server.py
```diff
@@ -1075,6 +1124,62 @@ async def rename_thread(
...
+    else:
+        metadata.pop("shared_at", None)
+    try:
+        await data_layer.update_thread(thread_id=thread_id, metadata=metadata)
+        logger.debug(
+            "[share_thread] updated metadata for thread=%s to %s",
+            thread_id,
+            metadata,
+        )
+        print("updated metadata for thread=%s to %s", thread_id, metadata)
```
sandangel: should not use print. should add ruff rule to lint error

sandangel (approved):
As I understand:
- viewer only allow view shared thread, can not add more chat messages
- the owner is still able to add more messages and those are automatically shared when viewer reload
- viewer list can be controlled by implementing callback to filter

please let me know if otherwise

backend/chainlit/server.py
```diff
@@ -951,6 +958,48 @@ async def get_thread(
...
+        try:
+            metadata = json.loads(metadata)
+        except Exception:
+            metadata = {}
+    if not isinstance(metadata, dict):
+        metadata = {}
+
+    if not bool(metadata.get("is_shared")):
+        # Pretend it doesn't exist if not shared
+        raise HTTPException(status_code=404, detail="Thread is not shared")
```
lamylio: Small detail, but if the goal is to avoid leaking existence of the thread, then the detail `Thread is not shared` goes against that.

PS: check to remove the prints above

frontend/src/components/share/ShareDialog.tsx
```diff
@@ -0,0 +1,186 @@
...
+          const next = { ...prev, threads: [...prev.threads] };
+          const idx = next.threads.findIndex((t) => t.id === threadId);
+          if (idx !== -1) {
+            const md = { ...(next.threads[idx].metadata || {}) };
+            md.is_shared = true;
+            next.threads[idx] = { ...next.threads[idx], metadata: md } as any;
+          }
+          return next;
+        });
+        toast.success('Share link created!');
```
lamylio: Hardcoded english toasts ? :(

frontend/src/pages/Thread.tsx
```diff
@@ -32,13 +33,16 @@ export default function ThreadPage() {
     });
   }, [id]);
 
+  const isSharedRoute = location.pathname.startsWith('/share/');
+
   return (
     <Page>
       <>
-        {config?.threadResumable && !isCurrentThread ? (
+  {isSharedRoute ? <ReadOnlyThread id={id!} /> : null}
```
lamylio, line 41: This poses a duplication issue with the lines 54-60, as if we are in a `/share` path while the `config.threadResumable` is false, we ends up with 2 components of `ReadOnlyThread`.

image

Adding `&& !isSharedRoute` to line 54 fixes it.

backend/chainlit/server.py
lamylio: Also, we should (in the same way as in socket.py) pop some of the metadatas, i.e. `env` and profiles to avoid leaking credentials/infos. :)

lamylio:
Thanks for the PR, that's great already ! :) 

I've reviewed a few things. As for `AutoOpenSharedThread`, I haven't found any reference to it, so the `open_shared_thread` event handler in `socket.py` doesn't seems to be called, leading to the `@on_shared_thread_view` callback to also not be called ?

More globally, I'm wondering wether the socket communication is needed at all, and if the `on_shared_thread_view` callback shouldn't be called in the `get_shared_thread()` endpoint instead. I've tested it locally, and this has the behavior I expected. 

```py
async def get_shared_thread(...):
# [...]
on_shared = getattr(config.code, "on_shared_thread_view", None)
    if on_shared:
        allowed = await on_shared(thread, current_user, "token") # either add it as params, or remove this functionality?
        if not allowed:
            raise HTTPException(status_code=404)
return JSONResponse(content=thread)
```

backend/chainlit/server.py
hayescode: Good suggestions! fixed.

backend/chainlit/server.py
hayescode: removed

frontend/src/pages/Thread.tsx
hayescode: nice catch! fixed

frontend/src/components/share/ShareDialog.tsx
hayescode: added

hayescode:
> Thanks for the PR, that's great already ! :)
> 
> I've reviewed a few things. As for `AutoOpenSharedThread`, I haven't found any reference to it, so the `open_shared_thread` event handler in `socket.py` doesn't seems to be called, leading to the `@on_shared_thread_view` callback to also not be called ?
> 
> More globally, I'm wondering wether the socket communication is needed at all, and if the `on_shared_thread_view` callback shouldn't be called in the `get_shared_thread()` endpoint instead. I've tested it locally, and this has the behavior I expected.
> 
> ```python
> async def get_shared_thread(...):
> # [...]
> on_shared = getattr(config.code, "on_shared_thread_view", None)
>     if on_shared:
>         allowed = await on_shared(thread, current_user, "token") # either add it as params, or remove this functionality?
>         if not allowed:
>             raise HTTPException(status_code=404)
> return JSONResponse(content=thread)
> ```

@lamylio good call there was some baggage still in here from an earlier approach i tried. I implemented this.


> As I understand:
> 
> * viewer only allow view shared thread, can not add more chat messages
> * the owner is still able to add more messages and those are automatically shared when viewer reload
> * viewer list can be controlled by implementing callback to filter
> * owner can not choose who to share with.
> 
> please let me know if otherwise

@sandangel I've seen many different behaviors users want, including my own. So when a user shares a thread we add `is_shared` and `shared_at` to the thread's metadata then inside of the @cl.on_thread_share_view callback it's up to the developers to define this. This requires a return of `True` to succeed. So devs could specify however they wish including granting admins access to all threads if they want to, or limit it even further if desired.

backend/chainlit/server.py
```diff
@@ -951,6 +958,60 @@ async def get_thread(
...
+            user_can_view = await config.code.on_shared_thread_view(
+                thread, current_user
+            )
+        except Exception:
+            user_can_view = False
+
+    is_shared = bool(metadata.get("is_shared"))
+
+    # Proceed only if both conditions are True.
+    if not (user_can_view and is_shared):
```
sandangel, line 1005: Should throw permission denied when user can not view

hayescode: An earlier comment suggested to not throw that error in order to not let the user know the thread exists.

celeriev:
Thank you @hayescode 🙏 🙏 🙏

sammydeprez:
Awesome thanks for this update, due to caching the solution might not always work bit the same how the images are stored somewhere. But once it works, it works!

PR #2495, opened 2025-09-04
Added LocalStorageClient and a new /storage/file API route

DmitriyAlergant:
This PR adds LocalStorageClient supporting local filesystem storage, and a new /storage/file API route to allow files serving via ChainLit backend itself, since local storage cannot provide a pre-signed URL for direct files downloads.

Problem statement: persistence requires a storage provider. But a dependency on Cloud Storage introduces complexity for fully local deployments (e.g. docker) that is not already in AWS/Azure/GCS.  Yes, MinIO exists and can be co-deployed and used with S3 storage provider, but it adds substantial complexity to the deployment; Not only for MinIO itself but also for reverse proxying scenarios (ingress), etc.  It's not the same level of complexity as having a Postgre database that's trivial.

Basic files upload and download capability can be well handled by ChainLit itself.

### DIFF
--- a/backend/chainlit/data/__init__.py
+++ b/backend/chainlit/data/__init__.py
@@ -57,9 +57,13 @@ def get_data_layer():
                 azure_storage_key = os.getenv("APP_AZURE_STORAGE_ACCESS_KEY")
                 is_using_azure = bool(azure_storage_account and azure_storage_key)
 
+                # Local Storage
+                local_storage_path = os.getenv("APP_LOCAL_STORAGE_PATH")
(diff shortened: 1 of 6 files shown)


DmitriyAlergant:
Now passes all tests and linting. Commits re-squashed into a single commit.

DmitriyAlergant:
@asvishnyakov @hayescode thoughts?

asvishnyakov:
Leave it for a while

dokterbob:
'A while' has passed. ;)

PR #2501, opened 2025-09-08
Rename author before sending token

zrquan:
We should rename author before sending the token in `stream_token`; otherwise, the avatar will not display correctly until all tokens have been sent.

### DIFF
--- a/cypress/support/run.ts
+++ b/cypress/support/run.ts
@@ -15,7 +15,7 @@ export const runChainlit = async (
     const testDir = spec ? dirname(spec.absolute) : CHAILIT_DIR;
     const entryPointFileName = spec
       ? spec.name.startsWith('async')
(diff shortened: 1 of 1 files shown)


hayescode:
@zrquan the send() method already has this can you try sending the message before streaming tokens into it?

zrquan:
This does solve the problem, but it's not clear to the user whether they should perform send() before stream_token(), either from the [documentation](https://docs.chainlit.io/api-reference/message#stream-a-message) or intuitively.

hayescode:
@zrquan you're right the docs had a bad example. Fixed here https://github.com/Chainlit/docs/pull/263

@asvishnyakov @sandangel

hayescode:
closing in favor of #2509 2509

PR #2509, opened 2025-09-11
Feat(front): separate author and avatar; Add easy backend function

lamylio:
Motivation: #2501 

Currently, the message avatar is either displayed from the author name or the miscellaneous and undocumented `metadata.avatarName` and requires the user to know he needs to `.send()` the message for it to have an effect.

Also, if `metadata.avatarName` is set, then the frontend ToolTip falsely displays the avatarName instead of the author. 

This PR aims to :

## Backend:

Add new `set_author_and_avatar`method to the message API, allowing dynamic updates to both the author and avatar, demystifying the process.

## Frontend:

Refactor the avatar display logic to separate author and avatar handling, with a fallback to the author's name if the avatar is not defined (as this is currently the case).

### DIFF
--- a/backend/chainlit/message.py
+++ b/backend/chainlit/message.py
@@ -1,5 +1,6 @@
 import asyncio
 import json
+import os
 import time
 import uuid
 from abc import ABC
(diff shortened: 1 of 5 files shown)


hayescode:
@lamylio `send()` send the message to the UI, `update()` updates the message in the UI, `stream_token()` send a chunk of text to the message content, and `remove()` removes the message from the UI. This seems like one of the least mystifying parts of this project.

I'm assuming you're doing something like this:
```py
msg = cl.Message(author="GPT")
await msg.stream_token(chunk)
```

But your issue here and in #2501 can be solved like this:
```py
msg = await cl.Message(author="GPT").send()
await msg.stream_token(chunk)
```

Maybe the docs just need to be updated to make this clearer? If not please share your use case and code.

hayescode:
I updated the documentation to fix the example here.

https://github.com/Chainlit/docs/pull/263

@asvishnyakov @sandangel review/merge?

lamylio:
Hey, thanks for the reply. Yeah I know and I agree with you, actually never had the issue in my case. 
However this PR addresses another issues:
1. The fact that the author's name and the avatar are linked
2. The not-easy-to-know "avatarName" metadata

and by "coincidence" (😉) could "solve" the #2501 issue, by offering another possibility : 

```py
msg = await cl.Message()
await msg.set_author_and_avatar(author="GPT")
```

I admit the `set_author_and_avatar` could be dropped, but I'd like the have the frontend modification. @hayescode

hayescode:
@lamylio I'd be ok with the frontend modification only, although I don't see how this will change any behavior or what the benefit is outside of making it clearer what's happening. Can you share why you want this?

lamylio:
> @lamylio I'd be ok with the frontend modification only, although I don't see how this will change any behavior or what the benefit is outside of making it clearer what's happening. Can you share why you want this?

Currently doing this : 

```py
await Message(
    content="Hello :)",
    author="Personal Assistant",
    metadata={"avatarName": "robot"},
).send()
```
Results in this : 
image

While with this PR it shows : 
image

Which is a useful use case, and better match the behavior it should have, imo.

hayescode:
@lamylio gotcha. Yes please remove the other changes and also please make a PR in https://github.com/Chainlit/docs with this great example code. Thanks!

PR #2512, opened 2025-09-12
fix: take overridden config into account in audio handlers

hayescode:
The audio handlers were using the global config instead of the overrides from the chat_profile

### DIFF
--- a/backend/chainlit/socket.py
+++ b/backend/chainlit/socket.py
@@ -12,7 +12,7 @@ from chainlit.auth import (
     require_login,
 )
 from chainlit.chat_context import chat_context
-from chainlit.config import config, ChainlitConfig
+from chainlit.config import ChainlitConfig, config
 from chainlit.context import init_ws_context
 from chainlit.data import get_data_layer
 from chainlit.logger import logger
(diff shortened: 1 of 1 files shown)


hayescode:
@asvishnyakov fixed

backend/chainlit/socket.py
```diff
@@ -330,10 +330,13 @@ async def window_message(sid, data):
 @sio.on("audio_start")  # pyright: ignore [reportOptionalCall]
 async def audio_start(sid):
     """Handle audio init."""
-    session = WebsocketSession.require(sid)
+    WebsocketSession.require(sid)
+    session = WebsocketSession.get_by_id(sid)
```
asvishnyakov: Why do you do

```python
WebsocketSession.require(sid)
session = WebsocketSession.get_by_id(sid)
```

if `require` already returns session?

hayescode: We need the session_id to get the users configs which includes overrides the require() is the socket_id

asvishnyakov: https://github.com/Chainlit/chainlit/blob/66ec36407721dd83af04363a68d333f8230b1414/backend/chainlit/session.py#L339-L354
`sid` here is Socket ID, isn't it? And `get_by_id` require Session ID, which is not the same

You're talking about global config vs config from the session, I get it. That's why 
```python
config: ChainlitConfig = session.get_config()
```
But why this change with `require` → `get_by_id`?

asvishnyakov: @hayescode

hayescode: ok changed back

asvishnyakov (approved):

Issue #2199, opened 2025-05-29
How to reliably check Copilot widget readiness and send "silent" frontend messages?


fracerma:
Hi Chainlit Team,

We're working on integrating the Chainlit Copilot widget into our frontend application and have encountered a couple of challenges where we'd appreciate some guidance or clarification.

**1. Reliably Checking Copilot Widget Readiness:**

We need to send an initial system message (e.g., with user context) from our frontend to the Chainlit backend as soon as the Copilot widget is fully mounted and ready to process messages via window.sendChainlitMessage.

We've observed that window.sendChainlitMessage might exist as a placeholder function even before the widget is fully operational, logging a message like "Copilot is not active. Please check if the widget is mounted." This makes simply checking for typeof window.sendChainlitMessage === 'function' unreliable for determining true readiness.

We've attempted a few strategies:
* Polling for window.sendChainlitMessage to change from its placeholder (difficult to distinguish).
* Listening for a specific event like chainlit-ready (couldn't find documentation or observe this event firing reliably).
* Polling for the existence of a specific DOM element that the widget creates (e.g., document.getElementById("chainlit-copilot")). This seems like the most promising approach so far, but we're unsure if there's a specific, guaranteed DOM element ID or class, or a more official "ready" event/callback we should be using.

Question: What is the recommended, most reliable method to programmatically determine from the frontend JavaScript that the Chainlit Copilot widget has finished its initialization and is ready to successfully process messages sent via window.sendChainlitMessage? Is there a specific event, callback, or state we can check?

**2. Sending "Silent" Messages from Frontend to Backend:**

We have use cases where we'd like to send contextual information from the frontend to the Chainlit backend (e.g., user's current page, specific UI state) that should be processed by the agent but not displayed as a message in the chat interface visible to the end-user. This is similar to how a "system message" might work, but initiated from the client-side after initial load.

Currently, messages sent via window.sendChainlitMessage({ type: "system_message", output: "..." }) seem to be handled by the @cl.on_message decorator, but we're unsure if the system_message type (or another type) inherently prevents rendering in the chat UI, or if there's a different mechanism for this.

Question: Is there a way to use window.sendChainlitMessage (or another frontend utility) to send data/messages to the Chainlit backend that are processed by the agent logic but are not rendered in the chat UI? If so, how should these messages be structured (e.g., specific type, output vs. content, metadata)?

Any insights or examples on these two points would be greatly appreciated!

Thank you for the great work on Chainlit!

[...]

PR #2533, opened 2025-09-18
Copilot: Collect metadata and add to message before sending

profpylons:
Allow copilot clients to send contextual information with a message about the state of the client.

eg.
```js
        (window as any).setChainlitMetadata({
          questionnaireState: {
            questionId,
            questionnaire,
          }
        });
```

The function `setChainlitMetadata()` is available before the widget is mounted in order that the first question asked has the state/metadata available immediately. This bypasses the difficulty of listening for when the widget is mounted.

In the backend this can read out and used to format queries to the downstream LLM

```python
@cl.on_message

    if cl.context.session.client_type == "copilot":
        print("Copilot detected.")

    if message.metadata:
        print(f"Message type: '{message.type}', sent with metadata: '{message.metadata}'")
        # Do something with different prompts based on the question passed
```


profpylons:
FYI previously we'd been using `system_message` as recommended by the docs.

```python
@cl.on_message
    if cl.context.session.client_type == "copilot":
        print("Copilot detected.")

    # Check for received system message that tells us to use guidance
    # from the chainlit widget copilot. This just sets the session state
    # and awaits a user message to process the query.
    if message.type == "system_message":
        # Store guidance metadata in the user session
        cl.user_session.set(SESS_CLIENT_STATE, message.metadata)
        print(f"Setting Client App State via system message: {message.metadata}")

        # Don't show the message to the user, this is a system message
        await message.remove()
        return

    # Determine guidance source and extract questionnaire information
    print(f"Fetching Client App State from user session: {cl.user_session.get(SESS_CLIENT_STATE)}")
    copilot_guidance = cl.user_session.get(SESS_CLIENT_STATE)

    # Update the prompts for different questions
```

But as we could not listen for amount event this meant the first question asked by a user could not be preceded by setting this metadata.

asvishnyakov:
@profpylons Could you elaborate a bit? I think the docs are talking about `window_message`, which is a way to send data through an `iframe`, not a system message.

---

https://docs.chainlit.io/api-reference/window-message

profpylons:
With an iFrame, yes, that seems to be the way. Using the copilot. the docs recommend [system_message](https://docs.chainlit.io/deploy/copilot#send-a-message). 

There are two drawbacks to this message sending.

1. The `sendChainlitMessage()` method is [not responsive](https://github.com/Chainlit/chainlit/blob/173797a6e66832db2c102175d60097bd90e05c88/libs/copilot/index.tsx#L66) until the popover has been loaded. So the copilot cannot be updated with the state of the client UI whilst the widget is collapsed. Opening the widget does not fire an event to preload the state from the client UI, so the first message sent to the chainlit backend is always unaware of the state. 
2. In any case, the Copilot UI [does not distinguish](https://github.com/Chainlit/chainlit/issues/2199) `system_message` from `user_message` and renders the content before firing the backend event. So when sending state information, it is necessary to update the `system_message` with `message.remove()` after receipt causing noticeable flicker in the widget. (see example code above).

I considered adding an event to fire when the copilot is mounted. But this just requires extra plumbing and tracking in the client app. 

In any case, I think it is still a good idea to send the state in metadata rather than the with `system_message` to prevent the UI flicker and reduce the need for strange session management hula-hoops to track the value in the backend.

This led me to here: Keep the necessary state locally in the widget and pass it with each `user_message` from the Copilot. It leads to:

1. Simpler & more stateless backend (not tracking client state in session variables)
2. No additional event listener in the client to pre-seed the widget after mounting the popover. Just send every state change.
3. Removal of the UI flicker

PR #2536, opened 2025-09-20
Add support for custom_meta_url

jstsch:
This PR introduces a new optional field `custom_meta_url` in `UISettings` and the default config.

* Adds `custom_meta_url` to `config.py` and default TOML.
* Updates `server.py` `get_html_template` to use: `default_url = config.ui.custom_meta_url or "https://github.com/Chainlit/chainlit"`

Currently the Chainlit GitHub link is hard-coded into meta tags. This means it always shows up in link previews (e.g. iMessage/Twitter). This patch makes it configurable, in line with other UI metadata (title, description, image).


jstsch:
PS: thank you for creating Chainlit, it has been super fun to work with

PR #2537, opened 2025-09-22
Fix/combined timezone and metadata

DmitriyAlergant:


DmitriyAlergant:
created by mistake, closing

PR #2538, opened 2025-09-22
Chainlit Data Layer - preserve metadata in update_thread()

DmitriyAlergant:


DmitriyAlergant:
I'll take a look at the failed test, a contract has changed after the test was written, I think I missed that commit to fix the test.

E2E tests failed not at my fault - may need to restart later
```
Downloading uv from "https://github.com/astral-sh/uv/releases/download/0.8.19/uv-x86_64-unknown-linux-gnu.tar.gz" ...
read ECONNRESET
Waiting 10 seconds before trying again
read ECONNRESET
Waiting 20 seconds before trying again
Error: read ECONNRESET
```

akotlar:
thank you for making this!

DmitriyAlergant:
@akotlar sorry I never got a chance to properly finish this PR (tests, etc). While eventually i think I fixed it in our fork, this change got intermingled with other things we needed to fix and at this time I don't have a clean atomic PR for this fix only which would have passed CI.   The original change still works though, but tests need to be fixed.

akotlar:
No man don't apologize, this was great. It helped me resolve an issue that I was having, and for that I'm grateful. I don't care about the tests, I can fix those :)

PR #2549, opened 2025-09-25
fix: on_chat_start not always firing

### LINKED ISSUE #2535: on_chat_start not firing on page load/login since 2.8
Hello,

After upgrading to Chainlit 2.8.0, on_chat_start no longer runs when I log in or simply refresh the page. My init flow doesn’t appear, and related logs are missing.

If I manually click “New chat”, on_chat_start fires and everything works.

Expected (per docs): on_chat_start runs on a new session after login/refresh.
Actual: Seems to require a user-initiated “New chat” to create a thread first.

Is there a new setting or recommended way in 2.8 to auto-create a chat/thread on load (so on_chat_start runs) without saving empty conversations? Any breaking change around when on_chat_start is triggered?

Thanks !

hayescode:
fixes #2535

This bug only happens **some of the time** and causes chat_settings sent from on_chat_start and LLM configuration to not fire. Clicking 'New Chat' is the only sure-fire way to trigger on_chat_start.

After lots of debugging trying to reproduce this bug I noticed that `context.session.restored` conditional was hit which skips on_chat_start. I was able to fix this issue by adding an additional conditional `context.session.has_first_interaction` which can only be True when the user has actually interacted.

Like mentioned it's difficult to reliably reproduce this bug, and the root cause is still alluding me, but for the meantime this doesn't appear to fix the issue.

### DIFF
--- a/backend/chainlit/socket.py
+++ b/backend/chainlit/socket.py
@@ -174,7 +174,11 @@ async def connection_successful(sid):
     await context.emitter.clear("clear_call_fn")
 
     if context.session.restored:
-        return
+        if config.code.on_chat_start and not context.session.has_first_interaction:
+            task = asyncio.create_task(config.code.on_chat_start())
+            context.session.current_task = task
+        else:
+            return
 
     if context.session.thread_id_to_resume and config.code.on_chat_resume:
         thread = await resume_thread(context.session)


fgalind1:
I can highly reproduce the issue on my setup and with this PR it fixed the issue 👍

Elienvalleau:
This PR fixes my issue too

SeaRedm:
I had been able to reproduce this by setting the browser performance to 3G in dev tools and sometimes also the CPU lower, if I was able to then switch profiles before it loaded it would then start a new chat and almost 100% of the time never load the settings, we had found if latency on device or location distance i.e. Italy to Australia was present then this issue would be more pronounced. The issue we would get was the model not found due to settings not loading. Only since 2.8

Issue #2525, opened 2025-09-16
"New Chat"-confirmation popup behaves counter-intuitively (bug?)


cjayb:
**Describe the bug**

The "New Chat" button (top-left) launches a confirmation popup with two options
 
- Cancel (secondary colour)
- Confirm (primary colour)

The UI is confusing/unintuitive: hitting enter closes the popup without starting a new chat. Enter must be mapped to Cancel, despite clear colour cue to the opposite.

**To Reproduce**, try the above

**Expected behavior**

Given that the Confirm-button is highlighted with the primary colour, it is natural to expect that hitting Enter will be equivalent to clicking on Confirm.

**Development environment:**

- chainlit-2.8.0
- python-3.13.5

**Desktop (please complete the following information):**

- OS: macOS
- Browser: tested on Chrome and Safari
- Version Chrom 139.0.7258.155

Ms3314:
Heyy @cjayb I have tried to fix this on my end please check my pr here https://github.com/Chainlit/chainlit/pull/2556

PR #2556, opened 2025-09-27
fix: confirm when enter

Ms3314:
before #2525  
After 

https://github.com/user-attachments/assets/2c4d22d1-326d-4ba4-87bc-22ee6fa2c507

mapped the enter key to confirm a new chat

### DIFF
--- a/frontend/src/components/header/NewChat.tsx
+++ b/frontend/src/components/header/NewChat.tsx
@@ -32,9 +32,20 @@ export const NewChatDialog = ({
   handleClose,
   handleConfirm
 }: NewChatDialogProps) => {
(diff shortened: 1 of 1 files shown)


cjayb:
Hi @Ms3314 ! Thanks for looking into this! I can confirm that your code fixes the "complaint" I have about the UX of the New Chat modal popup. It now defaults to Confirm when I hit enter.

However, I also notice that Tab-cycling now longer works. Hitting Tab once seems to highlght Cancel, but hitting Enter still confirms the modal and creates a new chat.

To be clear: I prefer your change over the current default! I feel it's much more intuitive that enter Confirms, given how prominent the button is. Hope you get a positive review here 👍

PR #2561, opened 2025-10-01
Added ability to display a Lucid icon in the Step, instead of the default avatar

Allaoua9:
image

### DIFF
--- a/backend/chainlit/step.py
+++ b/backend/chainlit/step.py
@@ -63,6 +63,7 @@ class StepDict(TypedDict, total=False):
     showInput: Optional[Union[bool, str]]
     defaultOpen: Optional[bool]
     language: Optional[str]
+    icon: Optional[str]
     feedback: Optional[FeedbackDict]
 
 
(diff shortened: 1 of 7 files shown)


hayescode:
@Allaoua9 if you're adding a property to the step class you'll need to add it to all of the data layers.

Allaoua9:
@hayescode i've updated the SQL datalayers to include the icon on INSERT.

That also means people using SQL databases would need to create a migration to add new icon field...

Allaoua9:
UP @hayescode @asvishnyakov @sandangel

PhenSG:
Hi, when will this nice enhancement be added to the Chainlit framework? What's blocking?

sandangel:
sr for the delay, this LGTM

sandangel:
@Allaoua9 can we use step metadata instead? This will avoid DB migration

Allaoua9:
> @Allaoua9 can we use step metadata instead? This will avoid DB migration

Hi @sandangel Let me check that and get back to you.

Allaoua9:
@hayescode @sandangel 
I have updated the code so that we store icon in metadata to avoid applying a database migration.

Should we move forward with this PR ?

Thanks

dokterbob:
Is this good to go? @hayescode @sandangel

Allaoua9:
I have rebased the PR and added support for the new **avatar_size** config.

@hayescode @sandangel is there any particular reason this PR is stalled ?

PR #2563, opened 2025-10-01
None

jasperpolaxinter:


asvishnyakov (changes requested):
This is nice idea, however we can't accept this PR as is, as it introduces breaking change

asvishnyakov:
@jasperpolaxinter Tests are failing rightfully. We need to support both URLs **and** icon names; otherwise, this is a breaking change and cannot be merged.

jasperpolaxinter:
There are actually 3 different cases here:

- An image/icon in /public
- A URL
- A named lucide icon.

Maybe it's worthwhile to think about what we'd like. Do we want to support all of these use cases?

PR #2568, opened 2025-10-02
Configurable header_links target attribute

mwmdev:
Allow header links to specify their target behavior:
- Add optional `target` field to `HeaderLink` model
- Support standard HTML target values: `_blank`, `_self`, `_parent`, `_top`
- Maintain backward compatibility (default is still `_blank`)

## Configuration Examples
  ```toml
  [[UI.header_links]]
      name = "Help"
      url = "/help"
      target = "_self"  # Open link in same tab
  ```
  
## Use cases
  - Useful when Chainlit is mounted as a FastAPI app, allows for accessing other routes in the same browser tab


frontend/src/components/ButtonLink.tsx
```diff
@@ -35,7 +38,8 @@ export default function ButtonLink({
           >
             <a
               href={url}
-              target="_blank"
+              target={target ?? '_blank'}
+              rel="noopener noreferrer"
```
sandangel, line 42: can you share why do we need `rel` ?

mwmdev: It's a security precaution that prevents potential reverse tabnabbing.

sandangel: https://www.elegantthemes.com/blog/wordpress/rel-noopener-noreferrer-nofollow

sandangel (approved):

PR #2393, opened 2025-08-11
Rewrite chat input to use textarea and ChatGPT-style command selection (fixes inline command bugs)

AidanShipperley:
This PR aims to resolve #2357.

---

This PR rewrites the chat input to use a standard `<textarea>`, removing all `contentEditable` and inline command rendering logic. The input now only handles user text, which fixes a long-standing set of bugs around pasted content (see #2326, #2264, #1890, #2032, #2087). From my testing, I believe this solves all issues with pasted content and maintains compatibility with the existing [command APIs](https://docs.chainlit.io/concepts/command), so it should be non-breaking.

**Key changes:**
- **No more inline commands:** All selected commands are now shown as animated pill buttons below the input instead of being rendered inline as rich text along with user input.
- **ChatGPT-style `/` command menu:** Typing `/` brings up a floating, animated command menu with full keyboard and mouse support.
- **"Tools" popover:** Non-button commands are still accessible via a dropdown popover now clearly labeled as "Tools" similar to ChatGPT with smooth navigation and tooltips.
- **Robust paste handling:** Pasting text or files works as expected, with no interference from command rendering.

**Video demo:**
> https://github.com/user-attachments/assets/29573ee2-2eda-4522-8a3b-ec8507ecf7da

This should resolve all the major issues with the previous inline command implementation and make the input UX much more reliable.

Let me know if you have any feedback or want any tweaks, as I am not as experienced with frontend development.

PR #2575, opened 2025-10-07
fix: propagate IME composition events in AutoResizeTextarea

servak:
## Summary

Fixes a bug where Japanese (and other IME-based) input would incorrectly send messages when pressing Enter during text conversion.

## Problem

After PR #2393 rewrote the chat input to use `<textarea>`, the `AutoResizeTextarea` component was not propagating `onCompositionStart` and `onCompositionEnd` events to its parent component (`Input.tsx`). This caused the parent's `isComposing` state to always remain `false`, resulting in Enter key presses during IME text conversion being incorrectly interpreted as message send commands.

## Solution

- Added `onCompositionStart` and `onCompositionEnd` to `AutoResizeTextarea` component props
- Created handler functions that:
  1. Update the component's internal `isComposing` state
  2. Propagate the events to parent handlers if provided
- This ensures both the `AutoResizeTextarea` and parent components correctly track IME composition state

## Test Plan

Tested with Japanese IME input:
1. ✅ Type "english" and press Enter → message sends correctly
2. ✅ Type "こんにちは" (Japanese) and press Enter during conversion → text is confirmed, message does NOT send
3. ✅ After confirming Japanese text, press Enter → message sends correctly

## Related Issues

This bug was introduced in PR #2393 when migrating from `contentEditable` to `<textarea>`.

### DIFF
--- a/frontend/src/components/AutoResizeTextarea.tsx
+++ b/frontend/src/components/AutoResizeTextarea.tsx
@@ -8,6 +8,12 @@ interface Props extends Omit<React.ComponentProps<'textarea'>, 'onPaste'> {
   placeholder?: string;
   onPaste?: (event: ClipboardEvent) => void;
   onEnter?: (event: React.KeyboardEvent<HTMLTextAreaElement>) => void;
+  onCompositionStart?: (
+    event: React.CompositionEvent<HTMLTextAreaElement>
+  ) => void;
+  onCompositionEnd?: (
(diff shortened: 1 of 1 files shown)


servak:
I tested the fix locally, and here is a short demo video.

https://github.com/user-attachments/assets/99db3e83-6289-4661-961d-b58f1c426188

Issue #2580, opened 2025-10-08
Add "@" parsing to invoke agents


jpoullet2000:
Instead of having to go into the settings (settings icon) and select a chat profile from a list it would be convenient to just use the "@" symbol for invoking an agent in the prompt. 

Something like:
"@wheather_agent is it sunny on Friday in London?"

PR #269, opened 2023-08-10
refactor: Remove actions from element and put it into a menu

alimtunc:
This pull request aims to enhance user-friendliness for actions. I have removed these actions from the elements, as they are no longer relevant, and instead, I have organized them within a menu for improved readability.

Capture d’écran 2023-08-11 à 10 35 01

PR #2585, opened 2025-10-11
Feature/chat settings invoke

Barani05072002:
**Title:** Add Chat Settings Invocation via `@` Keyword (#2850)

Hello friends. I saw one of the feature enhancement issue in chainlit repo [#2580](https://github.com/Chainlit/chainlit/issues/2580). PR introduces a new way for users to invoke chat settings using the '@' keyword. The feature enhances input interaction by supporting both '@' and '/' triggers:

- Typing `@` now displays the chat settings input prompt.
- Typing `/` continues to show input values as before.
- This dual-trigger approach improves accessibility and flexibility for users interacting with chat settings.

All relevant UI and backend logic have been updated to support this behavior.

Let me know if you'd like to add screenshots to related issues or discussions from the [[#2580](https://github.com/Chainlit/chainlit/issues/2580)].

Thanks.

### DIFF
--- a/.github/actions/pnpm-node-install/action.yaml
+++ b/.github/actions/pnpm-node-install/action.yaml
@@ -18,6 +18,7 @@ runs:
       uses: actions/setup-node@v4
       with:
         node-version: ${{ inputs.node-version }}
+        registry-url: 'https://registry.npmjs.org'
         cache: 'pnpm'
         cache-dependency-path: '**/pnpm-lock.yaml'
     - name: Install JS dependencies
--- a/.github/workflows/publish-libs.yaml
+++ b/.github/workflows/publish-libs.yaml
@@ -48,4 +48,7 @@ jobs:
         run: pnpm build:libs
 
       - name: Publish packages to npm
-        run: pnpm publish --recursive ${{ inputs.dry_run && '--dry-run --no-git-checks' || '' }} # --no-git-checks allows testing from non-main branches
+        # --no-git-checks allows testing from non-main branches and publishing from release tags
+        run: pnpm publish --recursive --no-git-checks ${{ inputs.dry_run && '--dry-run' || '' }}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_REACT_CLIENT }}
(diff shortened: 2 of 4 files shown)


hayescode:
@Barani05072002 you want to invoke *chat settings* with the @ sign, not commands? Are you using a multi-agent framework? 

Chainlit is behind the curve on multi-agent support. It's one of the top major features requests. I'd love to see a right-hand side pane to do this because you're not alone. Sub agents can have way too many settings to fit in chat_settings.

Barani05072002:
> @Barani05072002 you want to invoke _chat settings_ with the @ sign, not commands? Are you using a multi-agent framework?
> 
> Chainlit is behind the curve on multi-agent support. It's one of the top major features requests. I'd love to see a right-hand side pane to do this because you're not alone. Sub agents can have way too many settings to fit in chat_settings.

Thanks for the feedback! Yes, in this pull request, using the @ sign allows us to invoke chat settings, not commands. We already have the option to invoke commands using the / sign. In Chainlit, the chat settings can include multiple select-options, so I added a feature that lets you invoke chat settings using a combination of @ and /. When you click @, it shows the chat settings input, and after choosing the required option, you can use / to select the appropriate value.

Choosing Chat_settings Input
image

Choosing Value  of the Input
image

after choosed the Value
image

That’s how it works for now — currently, I can only invoke the Select type options from the chat settings, not sliders or radio buttons. Let me know if you’d like to see the code I used in this image, or if we want to enhance it further to support other types.

hayescode:
I think this would be a temporary workaround and should be configurable in config.toml. Please add example and explanation to the docs along with the Select-only limitation.

Barani05072002:
Hello @hayescode, I’ve added documentation for invoking Chat Settings using the combination of @ and / symbols. You can find it in docs [docs #269](https://github.com/Chainlit/docs/pull/269). 

Could you please review and merge it? Let me know if any changes are needed. Thanks!

PhenSG:
For me, this feature's implementation is too "developer-centric" and not enough "end user-centric". Don't forget that some companies use Chainlit as a front-end for end/business users and not only for IT teams !
So we must be very careful with what we propose as a user experience on the frontend.

dokterbob:
Marking this as stale.

PR #2593, opened 2025-10-17
:bug: Dump thread tags in order to store them in sqlalchemy datalayer sqlite

b3j0f:
Sqlite does not support arrays.
A solution is to  dump/load arrays in/from json.


hayescode:
@b3j0f please add a description and reason why this change is needed, and why just sqlite.

b3j0f:
> @b3j0f please add a description and reason why this change is needed, and why just sqlite.

Because sqlite does not support arrays. A solution is to format arrays in json

Issue #2581, opened 2025-10-08
Add option to access raw OAuth token response


ADWasHere:
**Is your feature request related to a problem? Please describe.**
At the moment, Chainlit's OAuth integration only retrieves the provider's access token. When using providers like Keycloak, where the refresh token is required to renew the access token without re-authentication, this is restrictive. It is difficult to implement automatic token renewal or access other fields (such as expires_in, scope) without access to the complete OAuth response.

**Describe the solution you'd like**
Give devs the choice to receive the complete raw JSON response from the OAuth provider token endpoint rather than just the access token. A new function (get_token_raw_json, for example) would return the complete JSON, including access_token, refresh_token, and other fields, while the existing get_token function could be kept for backward compatibility. The access token or any other information required can then be extracted by devs.

**Describe alternatives you've considered**
- Storing only the access token and forcing users to re-authenticate when it expires (JWT should be short-lived)
- increase the exp of the JWT but this would lead to security risks

**Additional context**

ADWasHere:
@hayescode Can you mark this issue as Closed? Fix was merged

PR #2608, opened 2025-10-27
Added raw response from OAuthProviders

ADWasHere:
Suggested fix for #2581 

**Changes to oauth_providers.py**

- Added get_raw_token_response() to every provider
- get_token() still remains and uses the get_raw_token_response()
- Generalized the exception when token is missing


cjayb:
+1 on this (pinging to keep the PR alive so dosu, the eager bugger, doesn't stale it!)

Issue #2611, opened 2025-10-29
Messages sent during on_chat_resume disappear after on_chat_resume


lupingsse:
**Describe the bug**
The chat resume feature seems to have an issue with resuming the thread correctly, most notably messages sent during chat resume disappear immediately afterwards.

When a function decorated using `@chainlit.on_chat_resume` is called, the following happens:
1. The `welcome view` is shown (as if the user started a new chat)
2. (optional) 
    2.a. A message is sent using chainlit.Message().send()
    2.b. The the `chat view` (displaying previous messages) is loaded
    2.c. The new message appears in `chat view`
    2.d. After `on_chat_resume` returns, the new message disappears immediately
3. The chat view is (re)loaded displaying previous messages (not including a message sent during step 2)

We are guessing the issue is, that `context.emitter.resume_thread` is called after `on_chat_resume` instead of before, and will post a merge request fixing this issue shortly.

**To Reproduce**

1. Enable authentication and data persistence (we used the example `header_auth_callback` and the official data layer as shown in chainlit docs).
2. Run an application that sends a message `msg` during `on_chat_resume` (example code down below).
3. Start a new chat & send a message `prev_msg`.
4. Start another new chat.
5. Switch back to the first chat, triggering `on_chat_resume`. The `welcome view` is shown. When `msg` is sent, the `chat view` is loaded and previous messages (including `prev_msg`) and `msg` are displayed. When `on_chat_resume` returns, `msg` disappears.

Example code for reproduction:
```
import chainlit as cl
from chainlit.types import ThreadDict
from starlette.datastructures import Headers

@cl.on_message
async def on_message(message: cl.Message) -> None:
    await cl.Message("in cl_on_message").send()

@cl.on_chat_resume
async def on_chat_resume(thread: ThreadDict) -> None:
    # on_chat_resume, 'welcome view' is shown (as if user started a new chat) instead of 'chat view' with previous messages
    await cl.sleep(2)
    # sending a new message triggers loading the 'chat view'
    # the new message itself appears in chat immediately
    await cl.Message("in on_chat_resume").send()
    # showcase the new message indeed appears in chat
    await cl.sleep(2)
    # new message disappears
    return

@cl.on_chat_start
async def on_chat_start():
    app_user = cl.user_session.get("user")
    await cl.Message(f"Hello {app_user.identifier}").send()

@cl.header_auth_callback
async def header_auth_callback(headers: Headers) -> cl.User:
    return cl.User(
[...]

Issue #2338, opened 2025-07-31
@on_chat_resume Callback Does Not Refresh Browser for Message.send() and CustomElement.send()


i-w-a:
---
## Bug: `@on_chat_resume` Callback Does Not Refresh Browser for `Message.send()` and `CustomElement.send()`

**Describe the bug**
When using the `@on_chat_resume` decorator in Chainlit, any `Message.send()` or `CustomElement.send()` calls made within the decorated function don't update the browser interface until the page is manually refreshed. This means you won't see real-time updates for messages or custom elements after a chat session resumes.

**To Reproduce**

1.  Set up a Chainlit application with a function decorated by `@on_chat_resume`.

    ```python
    import chainlit as cl

    @cl.on_chat_start
    async def start():
        await cl.Message(content="Chat started!").send()

    @cl.on_chat_resume
    async def resume():
        await cl.Message(content="Chat resumed! This message often doesn't appear until refresh.").send()
        # You can also try sending a custom element
        # await cl.CustomElement(name="MyElement", content="Resumed element", display="inline").send()
    ```

2.  Inside that decorated function, include calls like `Message.send()` and `CustomElement.send()`. (See example code above.)
3.  Start your Chainlit application.

    ```bash
    chainlit run your_app.py -w
    ```

4.  Begin a chat session, then close the browser tab or navigate away.
5.  Reopen the Chainlit application URL to resume the chat.
6.  You'll notice that messages and custom elements sent via `Message.send()` and `CustomElement.send()` within the `@on_chat_resume` function aren't visible until you manually refresh your browser.

**Expected behavior**
When a chat session resumes and the `@on_chat_resume` function runs, `Message.send()` and `CustomElement.send()` should immediately display their content in the browser without needing a manual refresh. The browser should automatically update to show the new messages and custom elements.

**Screenshots**
(If you have screenshots showing the missing messages/elements before and after a refresh, add them here.)

**Desktop (please complete the following information):**

* OS: [e.g., macOS Sonoma 14.5]
* Browser: [e.g., Google Chrome]
* Version: [e.g., 126.0.6478.127]

**Smartphone (please complete the following information):**

* Device: [e.g., iPhone 15 Pro]
* OS: [e.g., iOS 17.5.1]
* Browser: [e.g., Safari]
* Version: [e.g., latest]

**Additional context**
[...]

Issue #2337, opened 2025-07-31
AskActionMessage in @on_chat_resume hides previous messages


i-w-a:
### Describe the bug

When an `AskActionMessage` is used within a function decorated with `@on_chat_resume`, the original message that initiated the chat session disappears from the UI. Only the `AskActionMessage` and its actions are visible, which causes a loss of context for the user, as they can no longer see their initial query.

### To Reproduce

1.  Create a Python file (e.g., `app.py`) with the following code:
    ```python
    import chainlit as cl

    @cl.on_chat_start
    async def start():
        cl.user_session.set("message_history", [])
        await cl.Message(content="Session started. Please provide your initial message.").send()

    @cl.on_message
    async def main(message: cl.Message):
        # Store message in history and "end" the chat to trigger resume
        messages = cl.user_session.get("message_history")
        messages.append(message.content)
        cl.user_session.set("message_history", messages)

        # This message will trigger the @on_chat_resume logic
        await cl.Message(content="Processing your request... please refresh the page to see the resume behavior.").send()


    @cl.on_chat_resume
    async def on_chat_resume(author: str):
        messages = cl.user_session.get("message_history")
        
        # This is the original message, which should be visible
        # but currently disappears.
        original_message = messages[0]
        
        # The AskActionMessage that causes the issue
        actions = [
            cl.Action(name="continue", value="continue", label="Continue"),
            cl.Action(name="cancel", value="cancel", label="Cancel")
        ]
        await cl.AskActionMessage(
            content=f"Resuming chat for '{author}'. Do you want to continue?",
            actions=actions
        ).send()
    ```
2.  Run the app with the command: `chainlit run app.py -w`
3.  In the browser, send an initial message like "Hello, this is my first message".
4.  After seeing the "Processing your request..." message, refresh the browser page.
5.  See that the original message has disappeared.

### Expected behavior

When the page is refreshed, I expect to see my original message ("Hello, this is my first message") preserved in the chat history, followed by the new `AskActionMessage` ("Resuming chat for 'user'. Do you want to continue?").

### Screenshots

[Attach a screenshot here showing the original message is missing after refresh]

### Desktop (please complete the following information):

* **OS:** [e.g., Windows 11, macOS Sonoma]
* **Browser:** [e.g., Chrome, Firefox]
* **Version:** [e.g., 1.0.0 (Chainlit version)]

### Additional context

[...]

PR #2612, opened 2025-10-29
attempt to fix issue #2611

lupingsse:
See issue [#2611](https://github.com/Chainlit/chainlit/issues/2611) for a detailled explanation of cause for this request.

Calling `contex.emitter.resume_thread(thread)` after executing the hook to `on_chat_resume`, restores the saved thread and, thereby, overwrites any potential UI (e.g., `chainlit.Message().send()`) triggered during `on_chat_resume`.

If we didn't overlook any specific reasons against calling `resume_thead` before `on_chat_resume`, this would fix issue #2611.


lupingsse:
As dosubot mentionend, this also relates to (and as far as i can tell also fixes) issue #[2338 ](https://github.com/Chainlit/chainlit/issues/2338) and #[2337](https://github.com/Chainlit/chainlit/issues/2337).

PR #2639, opened 2025-11-08
Include raw response in callback

ADWasHere:
This allows developers to access provider-specific fields, such as `expires_in` or custom metadata in the `oauth_callback`


ADWasHere:
@hayescode Hey, what happend to this pr? 
It's still relevant to get the raw response of the token. Especially if you need the refresh-token instead of only the token

Issue #2599, opened 2025-10-21
Page Blinking with two auth methods


Alex2251:
**Describe the bug**
The application starts blinking if two authentication methods are used by login and password, and by the header in the request.

**To Reproduce**

1. Add in your app two methods **header_auth_callback** and **password_auth_callback**
2. add to  **header_auth_callback** code: return None #to simulate that you didn't send header
3. Run app
4. Log in by login and password
5. Start chat, you will see blinking. 
6. **header_auth_callback** is still colling but you are already loged in by password
7. If you refresh page the problem will disappear

**Expected behavior**
You can login using one of the methods without blinking.

**Desktop (please complete the following information):**

- OS: Windows 11
- Browser chrome
- Version 141.0.7390.108

**Additional context**
I think refresh helps becuse cookies start to work.

Alex2251:
@dosu your solution works only if you use one auth method.

PR #2640, opened 2025-11-09
Bug: Fix page blinking after login

ADWasHere:
When using the `header_auth_callback ` and another auth callback for login, the page starts to "blink" as soon as a user logged in. This is because `header_auth_callback ` always gets called when routing to the root. 

Fix: Now when a user is already authenticated, the `header_auth_callback `won't be called
Issue: #2599


constantinidan:
@cubic-dev-ai

constantinidan:
@cubic-dev-ai review this

Issue #2283, opened 2025-07-11
Metadata is overwritten on thread update if not explicitly specified


asvishnyakov:
**Describe the bug**

If `metadata` is `None` (which is default value) on `update_thread`, it will overwrite saved metadata with empty object

https://github.com/aitsc/chainlit/blob/31fe84a212f5a68fd0816f1e259d012c6cfa99e7/backend/chainlit/data/chainlit_data_layer.py#L551

asvishnyakov:
Still relevant

DmitriyAlergant:
Same issue: https://github.com/Chainlit/chainlit/issues/2497

@asvishnyakov can you comment on the proposed fix? https://github.com/Chainlit/chainlit/pull/2538
Note, it overlaps an conflicts with Timezones fix we were discussing recently - so this branch is based on chainlit's main and does not depend on the other fix, although ideally they would be merged.

PR #2642, opened 2025-11-10
Fix thread sharing not working with persistency enabled

jfperusse:
The core of the issue is that the user_session parameters are saved to the persistent thread metadata, overwriting any other custom field in the thread's metadata.

For thread sharing, the following steps would reproduce the issue if you have a simple on_shared_thread_view implementation returning the value of the is_shared thread metadata field:
* Send a message in a new thread.
* Share the conversation.
* Open the share URL in a new tab, refresh that new tab -> everything is fine.
* Refresh the original tab, or create a new conversation.
* Refresh the share URL tab -> the conversation is no longer accessible.

The fix is to store the user_session in a new "user_session" thread metadata field, keeping the rest of the thread metadata intact. We also made sure this is backwards compatible, loading the user_session from the metadata directly if the user_session field does not exist.

### DIFF
--- a/backend/chainlit/socket.py
+++ b/backend/chainlit/socket.py
@@ -48,9 +48,15 @@ def restore_existing_session(sid, session_id, emit_fn, emit_call_fn):
     return False
 
 
-async def persist_user_session(thread_id: str, metadata: Dict):
+async def persist_user_session(thread_id: str, user_session: Dict):
     if data_layer := get_data_layer():
-        await data_layer.update_thread(thread_id=thread_id, metadata=metadata)
+        thread = await data_layer.get_thread(thread_id=thread_id)
+        current_meta = thread.get("metadata") or {}
+        if isinstance(current_meta, str):
+            current_meta = json.loads(current_meta)
+        current_meta["user_session"] = user_session
+
+        await data_layer.update_thread(thread_id=thread_id, metadata=current_meta)
 
 
 async def resume_thread(session: WebsocketSession):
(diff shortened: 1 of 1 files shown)


asvishnyakov:
I think the reason is [this bug](https://github.com/Chainlit/chainlit/issues/2283) and I would rather fix it, than merge this workaround. I think we already had multiple attempts to fix it, I'll review one of them and merge

jfperusse:
> I think the reason is [this bug](https://github.com/Chainlit/chainlit/issues/2283) and I would rather fix it, than merge this workaround. I think we already had multiple attempts to fix it, I'll review one of them and merge

I don't consider this to be a workaround. It's not only if metadata is empty. The way user session is persisted is to **always** overwrite the thread metadata with the user_session (see `persist_user_session`).

We've been running with this fix for a month now without issues, and our thread sharing is working fine.

asvishnyakov:
@jfperusse `update_thread` shouldn't overwrite metadata, that's bug and it only occurs in Chainlit data layer, other data layers update thread metadata (merge old and new one) instead

jfperusse:
> @jfperusse `update_thread` shouldn't overwrite metadata, that's bug and it only occurs in Chainlit data layer, other data layers update thread metadata (merge old and new one) instead

Right, I see how some other data layers merge the metadata instead of update it.

I still think custom thread metadata and user session persistence should not be stored directly in the thread metadata since their is a risk of name conflicts. Users of thread metadata have to be aware of user_session fields to avoid naming their metadata keys the same.

This PR avoid this risk by clearly separating custom thread metadata from the serialized session context, under a "user_session" metadata key. It also solves the thread sharing problem in a data-layer agnostic way.

jfperusse:
@asvishnyakov Actually, I just found out that there are hardcoded places where we access user_session parameters from the thread metadata directly, such as inside tsx files (e.g.thread.metadata?.chat_profile), so this does introduce issues where the chat_profile dropdown is not showing the right profile on resume. I'll close this for now and look into merging metadata instead.

PR #2683, opened 2025-11-21
fix: fix chainlit sets logging while act as a library

### LINKED ISSUE #2682: Chainlit sets logging format before application code
Describe the bug

chainlit sets up the logging for the application it is in. A library should not do this.

Reproduces:

1. Setup a chainlit app at chainlit_app.py.
2. mount to fastapi at app.py

at app.py you configure logging, it would not word due to chainlit

Root cause: https://github.com/Chainlit/chainlit/blob/1aa9ad1b218acf3617dabdbaa1919a0809d01a66/backend/chainlit/logger.py#L4

Chainlit should setup logging only if it is run as a cli, not while as a library.

https://github.com/Chainlit/chainlit/issues/1676

goatwu1993:
fix #2682

---
## Summary by cubic
Stop Chainlit from changing global logging when imported as a library. Move logging.basicConfig to the CLI entrypoint and remove it from chainlit.logger, while keeping socketio/engineio loggers at ERROR.

<sup>Written for commit aff41d1176b97794233362b6425c66e12985d7eb. Summary will update automatically on new commits.</sup>


asvishnyakov:
@goatwu1993 While this PR fixes issue, I don't thin that's right solution, as it will change how logging is working for existing users. To make `logging.basicConfig` working in you app, you just run it before chainlit, in `__init__.py`, for example or call it with `force=True` parameter

goatwu1993:
> logging.basicConfig working in you app, you just run it before chainlit, in __init__.py, for example

Tried it no luck. Also library package should almost NEVER config logging themself, making importer have to configure logging before/after chainlit import.

nzjrs:
Thanks @goatwu1993  - I also recently noticed that chainlit was doing this. You are correct, libraries should never configure global logging.

PR #2689, opened 2025-11-25
fix(socket): set environ when restoring websocket session

daniel-oronsi:
Hi,

I encountered this behavior when using `cl.context.session.environ` to get the HTTP headers.
It seemed like the headers were outdated when I used them - and after going through the `connect` function it seemed like `environ` is not passed to the `restore_existing_session` function, which means that the updated socket won't necessarily have the updated environment variables.

If there's an intentional reason for this behavior that I'm missing, I'd appreciate some context :)
And one more question - I didn't dive deeply into the authentication flow, but do you think I should also update the user when restoring the session? for now I left it unchanged.

Thanks for the help!

---
## Summary by cubic
Restored websocket sessions now receive the current request environ so headers and context stay up to date after reconnects. This prevents stale header usage in session logic.

- **Bug Fixes**
  - Added an environ parameter to restore_existing_session and set session.environ during restore.
  - Passed environ when restoring a session and updated tests to assert environ propagation and not-found behavior.

<sup>Written for commit 0b840fe55f7ae3828fb303ea0fef74632da3780d. Summary will update automatically on new commits.</sup>


hayescode:
@daniel-oronsi thanks for the contribution! I don't see why we would include the environ so looks good to me but please address the CI issues then we can merge this

daniel-oronsi:
Hey, thanks for the quick response :)

I see that `main` is also failing with the same `header_auth` test, so I think the recent change to `main` might've introduced some flakiness...
I modified the test a bit, let me know if it looks good to you

PR #2697, opened 2025-11-28
Fix for Tasklist send method

VirtualNonsense:
Hey everybody.
The current version of the library breaks when trying to add a Tasklist element to a message.
The root of the issue seems to lie in the implementation of the [send method](https://github.com/Chainlit/chainlit/blob/9a9be68be6f4c2e438105a8f23d90e390f95e51b/backend/chainlit/element.py#L351) within tasklist because it does not have the required signature that is [defined in the baseclass](https://github.com/Chainlit/chainlit/blob/9a9be68be6f4c2e438105a8f23d90e390f95e51b/backend/chainlit/element.py#L234).
 

---
## Summary by cubic
Fix Tasklist sending by aligning the send method with the base Element signature. This resolves crashes when adding Tasklist elements to messages.

- **Bug Fixes**
  - Tasklist.send now accepts for_id and persist and forwards them to super().send.
  - update() generates a new message id and calls send(for_id=uuid...).

<sup>Written for commit 2300abf231a175cd18ab6f00db80533fb88f667b. Summary will update automatically on new commits.</sup>


hayescode:
@VirtualNonsense please fix CI

PR #2701, opened 2025-11-28
docs: Move hello.py to avoid import shadowing

RoyBA:
## Summary
Moves `hello.py` from `backend/chainlit/` to `backend/` to prevent Python import shadowing issues when testing with integrations.

## Problem
When contributors place test files inside `backend/chainlit/` (e.g., `backend/chainlit/hello.py`), Python's import resolution causes subtle but critical issues:

1. Running `chainlit run chainlit/hello.py` from `backend/` adds `backend/chainlit/` to the module search path
2. When the test file (or its dependencies) tries to `import openai` or `import langchain`, Python finds the local folders `backend/chainlit/openai/` and `backend/chainlit/langchain/` first
3. These folders contain Chainlit's integration code, not the actual packages
4. This causes confusing `AttributeError` exceptions that are difficult to debug

**Real error encountered:**
```python
from langchain_openai.chat_models import AzureChatOpenAI, ChatOpenAI
...
AttributeError: module 'openai' has no attribute 'DefaultHttpxClient'
```

## How to Reproduce

1. Create a test file at `backend/chainlit/test_app.py`:
```python
import chainlit as cl
from langchain_openai import ChatOpenAI

@cl.on_message
    llm = ChatOpenAI(model="gpt-4")
    response = await llm.ainvoke(message.content)
    await cl.Message(content=response.content).send()
```
2. From the `backend` directory, run:
```
uv run chainlit run chainlit/test_app.py
```

---
## Summary by cubic
Moved hello.py from backend/chainlit/ to backend/ to prevent Python import shadowing that breaks imports like openai and langchain during local testing. Updated docs, the CLI hello command, and e2e test runner to use the new path.

- **Migration**
  - From backend/, run: uv run chainlit run hello.py
  - Replace references to backend/chainlit/hello.py with backend/hello.py

<sup>Written for commit 4f3ced1c700f23b0d19db9c4845595a70cd3675d. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/chainlit/cli/__init__.py
+++ b/backend/chainlit/cli/__init__.py
@@ -211,7 +211,7 @@ def chainlit_run(
 @cli.command("hello")
 @click.argument("args", nargs=-1)
 def chainlit_hello(args=None, **kwargs):
-    hello_path = os.path.join(BACKEND_ROOT, "hello.py")
+    hello_path = os.path.join(os.path.dirname(BACKEND_ROOT), "hello.py")
     run_chainlit(hello_path)
 
 
--- a/cypress/support/run.ts
+++ b/cypress/support/run.ts
@@ -10,9 +10,10 @@ export const runChainlit = async (
   spec: Cypress.Spec | null = null
 ): Promise<ChildProcessWithoutNullStreams> => {
   const CHAILIT_DIR = join(process.cwd(), 'backend', 'chainlit');
+  const BACKEND_DIR = join(process.cwd(), 'backend');
 
   return new Promise((resolve, reject) => {
-    const testDir = spec ? dirname(spec.absolute) : CHAILIT_DIR;
+    const testDir = spec ? dirname(spec.absolute) : BACKEND_DIR;
     const entryPointFileName = spec
       ? spec.name.startsWith('async')
         ? 'main_async.py'


asvishnyakov:
1. [I wanted to do it too](https://github.com/Chainlit/chainlit/pull/2634), but unfortunately this is a breaking change. When you move it outside the chainlit folder, it is no longer part of the distribution package (the wheel), so when you call it installed from the wheel, CLI will fail because it cannot find `hello.py`.
2. Tests must be placed outside `backend/chainlit`. Right now we use `backend/tests`, and I think it is totally fine to continue doing so.
3. I would totally agree with the issues, though, so if there is any other solution (except this and renaming files, as it is, again, a breaking change), I would be happy to merge it. My best guess is to move `hello.py` into something like `backend/sample` and include it in the distribution package, but unfortunately I am sick now and will be able to test it only later.

RoyBA:
Thanks for the feedback!

I’ve figured out that moving `hello.py` to `backend/chainlit/sample/hello.py` would solve both issues:

1. It resolves import shadowing. Python adds `backend/chainlit/sample/` to sys.path (not `backend/chainlit/`), so it won’t be able to find the `openai/` and `langchain/` folders.
2. It remains part of the package. Since it’s still under `backend/chainlit/`, it’s automatically included in the wheel distribution.

What do you think of this solution?
Wishing you a quick recovery! 💪🏼

PR #2710, opened 2025-12-03
fix: failing header_auth spec

fanyak:
The header_auth spec fails on windows machines.
To fix it, we need to replace the second cy.wait call for ''@auth' that happens in `shouldBeLoggedIn`, because the request is already waited and consumed in the `setupInterceptors` function that is called before each test.
According to the [documentation](https://docs.cypress.io/api/commands/wait#Wait-automatically-increments-responses) `cy.wait` automatically increments responses, meaning that it waits for the **next nth** request with the alias "@auth".

---
## Summary by cubic
Fix Windows failures in the header_auth Cypress spec by replacing the second cy.wait('@auth') with a shared interception promise to avoid race conditions and reliably assert the auth response.

- **Bug Fixes**
  - Store the '@auth' interception via Promise.withResolvers and use cy.wrap(interceptionPromise) for assertions.
  - Register the '/user' intercept only after auth completes.

<sup>Written for commit e15188adeeee9803550cfdea6ba9e374fa8b124c. Summary will update automatically on new commits.</sup>

### DIFF
--- a/cypress/e2e/header_auth/spec.cy.ts
+++ b/cypress/e2e/header_auth/spec.cy.ts
@@ -1,3 +1,5 @@
+import { Interception } from 'cypress/types/net-stubbing';
+
 describe('Header auth', () => {
   beforeEach(() => {
     cy.visit('/');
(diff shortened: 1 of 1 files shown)


hayescode:
Thanks for your PR @fanyak ! Looks like we were both fixing this at the same time. I merged mine in #2711 only because I also bumped dependencies which we haven't done in a while, so closing this one. Again thanks for tackling this!

fanyak:
> Thanks for your PR @fanyak ! Looks like we were both fixing this at the same time. I merged mine in #2711 only because I also bumped dependencies which we haven't done in a while, so closing this one. Again thanks for tackling this!

thanks!

PR #2711, opened 2025-12-03
Fix: Fix CI issue introduced in #2640

hayescode:
fix issue introduced in #2640

Issue: The Cypress test should have an access_token cookie in /auth/header response was failing in the "after reloading" test suite. This happened because after a page reload, the user is already authenticated via the cookie set during initial login, so the frontend skips calling /auth/header again. The test was waiting for an /auth/header request that never occurs in this scenario.

Fix: Moved the /auth/header cookie verification test out of the shared shouldBeLoggedIn() function so it only runs during initial login. The reload tests now only verify that the user remains logged in (no alert message, "Hello admin" displayed) without expecting /auth/header to be called again.

Also I updated the uv lock file. It has been a while.

---
## Summary by cubic
Fixes the failing Cypress header-auth test in CI by asserting the access_token cookie only on initial login. After reload, tests now confirm the user is still logged in and no longer wait for /auth/header.

<sup>Written for commit d72e0ab8457ec58676bb6f5d90e483ab7d05eac0. Summary will update automatically on new commits.</sup>


asvishnyakov:
@hayescode I think you and @fanyak in #2710 is trying to fix the same issue, right?

hayescode:
@asvishnyakov yeah looks like we both tried fixing at the same time. I'd prefer to merge this one if we can, including the upgraded uv lock file.

sandangel (approved):
lgtm

Issue #2713, opened 2025-12-05
Thread Deletion Race Condition


Allaoua9:
# Thread Deletion Race Condition & State Management Issues

## Problem Summary

When deleting an active thread, a race condition occurs where `update_thread` is called on a thread that was already deleted, causing errors and potentially recreating incomplete thread data.

## Root Cause

The application maintains **multiple sources of truth** for the active thread ID, leading to synchronization issues:

1. **Route parameter** (`/thread/:id`) - `id` from `useParams()` in `Thread.tsx`
2. **`currentThreadIdState`** (Recoil atom) - Used by `useChatMessages()` to get `threadId`
3. **`threadHistoryState.currentThreadId`** (Recoil atom) - Set by `Thread.tsx` from URL params
4. **`threadIdToResumeState` (`idToResume`)** (Recoil atom) - Used for websocket connection auth

### The Race Condition Flow

1. User deletes active thread → `clear()` is called, clearing `currentThreadIdState` and `idToResume`
2. Socket disconnects and reconnects without thread ID
3. **Race condition**: `Thread.tsx` is still mounted (navigation hasn't completed)
4. `AutoResumeThread` component renders because `!isCurrentThread` is now `true` (since `threadId` was cleared)
5. `AutoResumeThread`'s `useEffect` sets `idToResume` back to the deleted thread ID
6. This triggers `_connect` to be recreated (dependency on `idToResume`)
7. `App.tsx`'s `useEffect` detects `connect` changed and calls `connect()` again
8. Socket connects with deleted thread ID → Backend tries to resume deleted thread
9. Backend session still has `session.thread_id = deleted_thread_id`
10. When user closes window/navigates away, disconnect handler tries to `update_thread` on deleted thread -> This would result in an incomplete thread to be created

## Current Fix (Workaround)

I've implemented a workaround for my client :
1. Clears state first (synchronously) to prevent `Chat` from redirecting
2. Navigates to `/` to unmount `Thread.tsx` and prevent `AutoResumeThread` from running

This fixes the immediate bug but doesn't address the underlying architectural issues.

I will create a PR for this.

## Technical Debt: Recoil is Archived

**Recoil has been archived** and hasn't had a release for over 3 years.

A **major state management refactor** is needed to:

1. **Consolidate thread ID state** into a single source of truth
2. **Migrate from Recoil to a modern, actively maintained library** zustand maybe ?

vquilon:
[...]

PR #2731, opened 2025-12-18
Workaround for thread deletion race condition issue

Allaoua9:
When deleting an active thread, a race condition occurs where update_thread is called on a thread that was already deleted, causing errors and potentially recreating incomplete thread data.

See details here : https://github.com/Chainlit/chainlit/issues/2713

---
## Summary by cubic
Fixes a race condition when deleting the active thread by navigating away immediately after clearing state. This prevents updates on a deleted thread and avoids recreating partial thread data.

- **Bug Fixes**
  - In ThreadList.tsx, call navigate('/') right after clear() when deleting the current thread, and remove the later navigate call.

<sup>Written for commit 24a8b13ac6cc49d5397fc84490ebae6aad37bce7. Summary will update automatically on new commits.</sup>

### DIFF
--- a/frontend/src/components/LeftSidebar/ThreadList.tsx
+++ b/frontend/src/components/LeftSidebar/ThreadList.tsx
@@ -149,6 +149,7 @@ export function ThreadList({
       threadIdToDelete === currentThreadId
     ) {
       clear();
+      navigate('/');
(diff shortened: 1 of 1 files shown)


Allaoua9:
@hayescode 
Can we reopen this as it's a real issue and the workaround is simple ?

PR #2680, opened 2025-11-20
Update watchfiles version constraint in pyproject.toml

shanto12:
Fix Python 3.14 compatibility issue by updating watchfiles dependency.

Updated watchfiles version constraint from >=0.20.0,<1.0.0 to >=0.20.0,<2.0.0 to allow watchfiles 1.1.1 which adds Python 3.14 support.

Fixes #2656

---
## Summary by cubic
Enable Python 3.14 support by widening the watchfiles dependency and adding 3.14 to the CI matrix. Updated watchfiles from >=0.20.0,<1.0.0 to >=0.20.0,<2.0.0 to allow 1.1.1.

<sup>Written for commit e155669aae6830bfe2600d40f557d361d2371a31. Summary will update automatically on new commits.</sup>

PR #2733, opened 2025-12-19
chore: bump watchfiles

hayescode:
We can't upgrade to python 3.14 because the `watchfiles` dependency was pinned below 1.0.0 and support was added in 1.1.0.

This was attempted to be fixed in #2680 but they're unresponsive so this will supersede that PR.

### DIFF
--- a/backend/chainlit/config.py
+++ b/backend/chainlit/config.py
@@ -99,6 +99,9 @@ latex = false
 # Autoscroll new user messages at the top of the window
 user_message_autoscroll = true
 
+# Autoscroll new assistant messages
+assistant_message_autoscroll = true
+
 # Automatically tag threads with the current chat profile (if a chat profile is used)
 auto_tag_thread = true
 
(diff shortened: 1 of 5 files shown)


asvishnyakov:
@hayescode I think Cubic has a point here. If we need this for Python 3.14 and it supported starting from 1.1.1, then what's the reason to update it to just 1.0.0. What was the reason behind CI fail? Something else required lower watchfiles version? Can we upgrade it too?

hayescode:
@asvishnyakov resolved

asvishnyakov:
@hayescode But 1.0.0 doesn't support Python 3.14, only 1.1.1 does. So the change is basically useless

hayescode:
> @hayescode But 1.0.0 doesn't support Python 3.14, only 1.1.1 does. So the change is basically useless

it allows users to upgrade to 3.14 because now the ceiling is 2.0.0 instead of 1.0.0 but OK changed to 1.1.1

PR #2745, opened 2026-01-07
fix: Starters should take selected mode

hayescode:
The default/selected mode is removed/nulled when clicking a starter so we add the mode state to preserve it.

---
## Summary by cubic
Starters now respect the selected mode. When clicking a starter, the current mode selections are included so they aren’t cleared.

- **Bug Fixes**
  - Read modesState and build a modes dict using the default option (or first option) per mode.
  - Attach modes to the IStep sent by Starter and add modes to the useCallback dependencies.

<sup>Written for commit 684d24441153fdd2c81167b453db7c4da94bab3c. Summary will update on new commits.</sup>

### DIFF
--- a/frontend/src/components/chat/Starter.tsx
+++ b/frontend/src/components/chat/Starter.tsx
@@ -6,6 +6,7 @@ import {
   ChainlitContext,
   IStarter,
   IStep,
+  modesState,
   useAuth,
   useChatData,
   useChatInteract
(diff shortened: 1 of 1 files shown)


hayescode:
@asvishnyakov @sandangel please take a look when you can please.

asvishnyakov (approved):

Issue #2758, opened 2026-01-16
Bug: Thread resume breaks with blank screen when CustomElements are persisted


Sanady:
**Describe the bug**
CustomElements cause a fatal frontend crash (`TypeError: Cannot assign to read only property 'Symbol(Symbol.toStringTag)' of object '#<DataView>'`) when resuming threads that contain persisted CustomElements. The application shows only a blank background and becomes completely unusable.

**To Reproduce**

1. Create a CustomElement component (e.g., `Citations.jsx` or `MessageTimestamp.jsx`) in `/public/elements/`
2. Use `cl.CustomElement(name="Citations", props={"citations": [...]}, display="inline")` and attach it to a message with `.send(for_id=message.id)`
3. The element renders correctly during the active session
4. Close the browser or navigate away
5. Return to the application and click on the thread containing the CustomElement in the sidebar
6. **Result:** Application crashes with blank screen. Browser console shows:
   ```
   TypeError: Cannot assign to read only property 'Symbol(Symbol.toStringTag)' of object '#<DataView>'
   ```

**Expected behavior**
CustomElements should persist correctly and render when resuming a thread, just like they did during the original session.

**Root Cause Analysis**
The issue appears to be in how Chainlit's frontend handles serialization/cloning of element data loaded from the database:

1. CustomElement `props` are stored correctly as JSONB in the `ui.elements` table
2. When the frontend loads thread data (via `get_all_user_threads` or `get_thread`), it attempts to clone/serialize the element data
3. The cloning operation fails on certain data structures, causing the DataView serialization crash
4. This crash happens **before** the JSX component even loads, so defensive parsing in the component cannot help

**Workaround**
The only working solution is to:
1. Override `get_thread` AND `get_all_user_threads` in a custom data layer to filter out all `type="custom"` elements
2. Delete existing CustomElements from the database: `DELETE FROM ui.elements WHERE type = 'custom';`
3. Use plain markdown instead of CustomElements for content that needs to persist

```python
class CustomSQLAlchemyDataLayer(SQLAlchemyDataLayer):
    async def get_all_user_threads(self, user_id, thread_id):
        threads = await super().get_all_user_threads(user_id, thread_id)
        if threads:
            for thread in threads:
                thread["elements"] = []  # Remove all elements
        return threads
```

**Desktop (please complete the following information):**
- OS: macOS
- Browser: Chrome (also reproducible in Safari)
- Chainlit Version: 2.9.5

**Additional context**
[...]

PR #2760, opened 2026-01-17
feat: ORM SQLAlchemyDataLayer

hayescode:
### Summary

Refactors [sql_alchemy.py](cci:7://file:///c:/Users/Haze/Desktop/chainlit/backend/chainlit/data/sql_alchemy.py:0:0-0:0) from raw SQL queries to SQLAlchemy ORM, improving maintainability and adding multi-dialect database support.

### Benefits

- **Multi-dialect support**: Now works with PostgreSQL, SQLite, and MySQL/MariaDB out of the box
- **Type safety**: ORM models provide better IDE autocompletion and type checking
- **Reduced code duplication**: Removed ~200 lines of repetitive SQL string construction
- **Automatic schema creation**: New `create_tables=True` parameter auto-creates tables on first use
- **Easier testing**: Test fixtures reduced from 100+ lines of manual SQL to 4 lines

### What's New

| File | Description |
|------|-------------|
| [chainlit/data/models.py](cci:7://file:///c:/Users/Haze/Desktop/chainlit/backend/chainlit/data/models.py:0:0-0:0) | New ORM models with [CrossDialectJSON](cci:2://file:///c:/Users/Haze/Desktop/chainlit/backend/chainlit/data/models.py:10:0-19:46) type for PostgreSQL/SQLite/MySQL compatibility |
| [chainlit/data/sql_alchemy.py](cci:7://file:///c:/Users/Haze/Desktop/chainlit/backend/chainlit/data/sql_alchemy.py:0:0-0:0) | Refactored to use ORM operations instead of raw SQL |

### Migration

**No breaking changes** - the public API remains identical.

**Optional**: Users can now enable automatic table creation:

```python

data_layer = SQLAlchemyDataLayer(
    conninfo="sqlite+aiosqlite:///data.db",
    create_tables=True,  # NEW: Auto-creates tables if they don't exist
)
```

---
## Summary by cubic
Refactors SQLAlchemyDataLayer to use SQLAlchemy ORM, adding multi-dialect support and optional automatic table creation. Improves maintainability and type safety, keeps the public API unchanged, and simplifies tests.

- **New Features**
  - Works with PostgreSQL, SQLite, and MySQL/MariaDB.
  - Optional create_tables=True to auto-create tables.
  - CrossDialectJSON for JSONB on Postgres and JSON elsewhere.

- **Refactors**
  - Replaced raw SQL with ORM models and session-based operations; unified per-dialect upsert.
  - Removed SQL helpers and duplicated query code.
  - Simplified tests by enabling automatic schema creation.

<sup>Written for commit 469274c210d66b2a46129c4f6e6d5de0a514bed0. Summary will update on new commits.</sup>

### DIFF
--- a/backend/chainlit/data/models.py
+++ b/backend/chainlit/data/models.py
@@ -1,0 +1,160 @@
+from __future__ import annotations
+
+from typing import Any, Optional
+
+from sqlalchemy import Boolean, ForeignKey, Integer, String, Text
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column, relationship
+from sqlalchemy.types import JSON, TypeDecorator
+
+
+class CrossDialectJSON(TypeDecorator):
+    """JSON type that uses JSONB on PostgreSQL and JSON on other databases."""
+
+    impl = JSON
+    cache_ok = True
+
+    def load_dialect_impl(self, dialect):
+        if dialect.name == "postgresql":
+            return dialect.type_descriptor(JSONB())
+        return dialect.type_descriptor(JSON())
+
+
+class Base(DeclarativeBase):
+    """Shared base for all ORM models. Required so Base.metadata.create_all() discovers all tables."""
+
+    pass
+
+
+class UserModel(Base):
+    __tablename__ = "users"
+
+    id: Mapped[str] = mapped_column(String(36), primary_key=True)
+    identifier: Mapped[str] = mapped_column(String(255), unique=True, nullable=False)
+    metadata_: Mapped[Optional[dict[str, Any]]] = mapped_column(
+        "metadata", CrossDialectJSON, nullable=True, default=dict
+    )
+    createdAt: Mapped[Optional[str]] = mapped_column(String(50), nullable=True)
+
+    threads: Mapped[list[ThreadModel]] = relationship(
+        back_populates="user", cascade="all, delete-orphan", passive_deletes=True
+    )
+
+
+class ThreadModel(Base):
+    __tablename__ = "threads"
+
+    id: Mapped[str] = mapped_column(String(36), primary_key=True)
+    createdAt: Mapped[Optional[str]] = mapped_column(String(50), nullable=True)
+    name: Mapped[Optional[str]] = mapped_column(String(255), nullable=True)
+    userId: Mapped[Optional[str]] = mapped_column(
+        String(36), ForeignKey("users.id", ondelete="CASCADE"), nullable=True
+    )
+    userIdentifier: Mapped[Optional[str]] = mapped_column(String(255), nullable=True)
+    tags: Mapped[Optional[str]] = mapped_column(Text, nullable=True)
+    metadata_: Mapped[Optional[dict[str, Any]]] = mapped_column(
+        "metadata", CrossDialectJSON, nullable=True
+    )
+
+    user: Mapped[Optional[UserModel]] = relationship(back_populates="threads")
+    steps: Mapped[list[StepModel]] = relationship(
+        back_populates="thread", cascade="all, delete-orphan", passive_deletes=True
+    )
+    elements: Mapped[list[ElementModel]] = relationship(
+        back_populates="thread",
+        cascade="all, delete-orphan",
+        passive_deletes=True,
+        foreign_keys="ElementModel.threadId",
+    )
+
+
+class StepModel(Base):
+    __tablename__ = "steps"
+
+    id: Mapped[str] = mapped_column(String(36), primary_key=True)
+    name: Mapped[str] = mapped_column(String(255), nullable=False)
+    type: Mapped[str] = mapped_column(String(50), nullable=False)
+    threadId: Mapped[str] = mapped_column(
+        String(36), ForeignKey("threads.id", ondelete="CASCADE"), nullable=False
+    )
+    parentId: Mapped[Optional[str]] = mapped_column(String(36), nullable=True)
+    disableFeedback: Mapped[Optional[bool]] = mapped_column(
+        Boolean, nullable=True, default=False
+    )
+    streaming: Mapped[Optional[bool]] = mapped_column(
+        Boolean, nullable=True, default=False
+    )
+    waitForAnswer: Mapped[Optional[bool]] = mapped_column(Boolean, nullable=True)
+    isError: Mapped[Optional[bool]] = mapped_column(Boolean, nullable=True)
+    metadata_: Mapped[Optional[dict[str, Any]]] = mapped_column(
+        "metadata", CrossDialectJSON, nullable=True
+    )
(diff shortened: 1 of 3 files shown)


Sanady:
Hey @hayescode, is this pull request going to fix the issue that I reported? https://github.com/Chainlit/chainlit/issues/2758

dominpm:
If we are currently using the official `ChainlitDataLayer` with a postgresql and a blob container in Azure for `storage_client` i.e.:

```python
@data_layer
def get_data_layer():
    return ChainlitDataLayer(
        database_url=DATABASE_URL,
        storage_client=AzureBlobStorageClient(
            container_name=DATALAYER_AZURE_BLOB_CONTAINER_NAME,
            storage_account=DATALAYER_AZURE_BLOB_STORAGE_ACCOUNT_NAME,
            storage_key=DATALAYER_AZURE_BLOB_STORAGE_ACCOUNT_KEY
        )
    )
```

should we migrate to this SQLAlchemy data layer?

And if so, is there a direct way?

PS: Is this datalayer still maintained https://github.com/Chainlit/chainlit-datalayer?

hayescode:
That data layer is not maintained. The goal here is to consolidate the mess of data layers.

ilkersigirci:
Could you kindly say the status of this PR? It would be very good that we have only one source of truth for database related operations

nzjrs:
@hayescode just sharing some observations - we are currently using the existing dalalchemy data layer and are patching around a few bugs when using it with sqlite. Hopefully these patches can be dropped if going to an ORM approach

Basically running SQLAlchemyDataLayer with SQLite (sqlite+aiosqlite:///) when auto_tag_thread = true with chat profiles active, the emitter passes tags as a Python list (e.g. ['default']) into update_thread(). SQLite complains  sqlite3.ProgrammingError: type 'list' is not supported on every thread update. Fix was

```python
class SQLiteDataLayer(SQLAlchemyDataLayer):
    """SQLAlchemyDataLayer with SQLite-compatible tags serialization.

    Chainlit passes tags as a Python list, which PostgreSQL binds natively
    but SQLite rejects. Serialize to JSON on write, deserialize on read.
    """

    async def update_thread(self, thread_id, name=None, user_id=None,
                            metadata=None, tags=None):
        if isinstance(tags, list):
            tags = json.dumps(tags)
        return await super().update_thread(
            thread_id, name=name, user_id=user_id,
            metadata=metadata, tags=tags,
        )

    async def get_all_user_threads(self, user_id=None, thread_id=None):
        threads = await super().get_all_user_threads(user_id, thread_id)
        if threads:
            for t in threads:
                if isinstance(t.get("tags"), str):
                    try:
                        t["tags"] = json.loads(t["tags"])
                    except (json.JSONDecodeError, TypeError):
                        t["tags"] = []
        return threads
```

We also have another patch get_all_user_threads returns element props as a raw JSON string instead of a parsed dict. The write path (sql_alchemy.py:549) parses it fine, but the read path (~:848) doesn't deserialize. So we manually parse props in our chat resume handler before the thread dict hits the frontend.

Hopefully these both go away once the orm layer knows the types that both underlying dbs handle. Just a note that you test these.

backend/chainlit/data/models.py
```diff
@@ -0,0 +1,160 @@
...
+    cache_ok = True
+
+    def load_dialect_impl(self, dialect):
+        if dialect.name == "postgresql":
+            return dialect.type_descriptor(JSONB())
+        return dialect.type_descriptor(JSON())
+
+
+class Base(DeclarativeBase):
+    """Shared base for all ORM models. Required so Base.metadata.create_all() discovers all tables."""
```
dokterbob, line 24: Inheriting directly from `DeclarativeBase` doesn't work?

dokterbob (commented):
This definitely looks like welcome and much needed cleanup.

I'd say LGTM but I am reluctant to approve without other users taking it for a spin - particularly in non-SQLite environments (PostgreSQL, MySQL/MariaDB), on their own apps.

Could anyone provide feedback?

dokterbob:
Does this mean we can close #1365?

dokterbob:
> That data layer is not maintained. The goal here is to consolidate the mess of data layers.

@hayescode You reckon' theres' incompatibilities migrating from the 'official' data layer?

nzjrs:
@hayescode another similar comment to my current issues post above https://github.com/Chainlit/chainlit/pull/2760#issuecomment-3920991376

get_element() at sql_alchemy.py:549 does json.loads(element_dict.get("props", "{}")). When an element has no props (e.g. a cl.Text with display="page"), the DB column is NULL. .get("props", "{}") returns None because the key exists with a NULL value -- the .get() default only kicks in when the key is absent, not when the value is None. json.loads(None) raises TypeError, returning a 500 on the /element/{id} route when a user clicks a page-display element. Override uses d.get("props") or "{}".                     

```python
  async def get_element(self, thread_id, element_id):       
      from chainlit.element import ElementDict
      query = """SELECT * FROM elements WHERE "threadId" = :thread_id AND "id" = :element_id"""
      parameters = {"thread_id": thread_id, "element_id": element_id}
      element = await self.execute_sql(query=query, parameters=parameters)
      if isinstance(element, list) and element:
          d = element[0]
          return ElementDict(
              id=d["id"],
              threadId=d.get("threadId"),
              type=d["type"],
              chainlitKey=d.get("chainlitKey"),
              url=d.get("url"),
              objectKey=d.get("objectKey"),
              name=d["name"],
              props=json.loads(d.get("props") or "{}"),
              display=d["display"],
              size=d.get("size"),
              language=d.get("language"),
              page=d.get("page"),
              autoPlay=d.get("autoPlay"),
              playerConfig=d.get("playerConfig"),
              forId=d.get("forId"),
              mime=d.get("mime"),
          )
      return None
```

ilkersigirci:
Is there any timeline for merging this into main @hayescode ?

dokterbob:
> Is there any timeline for merging this into main @hayescode ?

Apart from resolving merge conflicts, as this is a much-used component: I think we need feedback from users in the field.

1. Does this work/is it stable (in actual apps)?
2. Is it compatible with the original SQLAlchemy data layer. If not, what's the upgrade path and/or can/should we make it compatible.
3. Can we provide a migration path from the custom PostgresSQL 'official' data layer which ... is not the _most_ elegant code exactly? (IMHO we should, ideally it's 100% smooth.)

backend/chainlit/data/sql_alchemy.py
```diff
@@ -61,6 +73,10 @@ def __init__(
         self.async_session = sessionmaker(
```
Allaoua9, line 73: maybe we should be using async_sessionmaker to align with sql alchemy asyncio doc.

backend/chainlit/data/sql_alchemy.py
Allaoua9, line 70: I think we should allow the user to configure and inject their own engine / session.

Allaoua9: In our case, we already have an engine created by our fastapi app.

PR #2775, opened 2026-02-05
fix: initialize user_can_view before conditional check

### LINKED ISSUE #2766: UnboundLocalError: user_can_view referenced before assignment when on_shared_thread_view hook is not set (2.9.6)
**Describe the bug**
In Chainlit 2.9.6, the shared thread endpoint can crash with an UnboundLocalError because user_can_view is only assigned inside the optional @cl.on_shared_thread_view hook. When the hook is not defined / not executed, the code still evaluates (not user_can_view) and raises:
UnboundLocalError: cannot access local variable 'user_can_view' where it is not associated with a value.

This reproduces on a fresh install using the installed wheel at .venv/lib/python3.12/site-packages/chainlit/server.py (around line 1019).

**To Reproduce**

1. Install Chainlit 2.9.6 (fresh venv).
2. Create a minimal Chainlit app without defining @cl.on_shared_thread_view.
3. Start the server (e.g., chainlit run server.py).
4. Call the shared thread endpoint (example):

Open the share URL in the browser

5. See error.

**Expected behavior**
When @cl.on_shared_thread_view is not implemented, Chainlit should not crash. The request should be handled gracefully (e.g., return a 403/404 or a safe default behavior), and user_can_view should have a default value (e.g. False) before being checked.

**Screenshots**

- OS: wsl Ubuntu 22.04
- Browser chrome
- Version latest

Suggested fix/workaround: set a safe default before the optional hook runs, e.g. user_can_view = False prior to evaluating (not user_can_view).

Stack trace:
ERROR: Exception in ASGI application
  File ".../site-packages/uvicorn/protocols/http/h11_impl.py", line 403, in run_asgi
    result = await app(...)
  File ".../site-packages/fastapi/applications.py", line 1054, in __call__
    await super().__call__(scope, receive, send)
  File ".../site-packages/starlette/routing.py", line 75, in app
    response = await f(request)
  File ".../site-packages/fastapi/routing.py", line 302, in app
    raw_response = await run_endpoint_function(
  File ".../site-packages/fastapi/routing.py", line 213, in run_endpoint_function
    return await dependant.call(**values)
  File ".../site-packages/chainlit/server.py", line 1019, in get_shared_thread
    if (not user_can_view) and (not is_shared):
UnboundLocalError: cannot access local variable 'user_can_view' where it is not associated with a value

**Additional context**

thecaptain789:
## Summary

Fixes `UnboundLocalError` when the `@cl.on_shared_thread_view` hook is not defined.

## Problem

In `backend/chainlit/server.py`, the `user_can_view` variable was only assigned inside the optional `on_shared_thread_view` hook block, but was used unconditionally in the subsequent check at line 1019:

```python
if (not user_can_view) and (not is_shared):
```

When the hook is not defined/registered, this caused:
```
UnboundLocalError: cannot access local variable 'user_can_view' where it is not associated with a value
```

## Solution

Initialize `user_can_view = False` before the conditional block, ensuring the variable is always defined regardless of whether the hook exists.

## Changes

```diff
+    user_can_view = False
     if getattr(config.code, "on_shared_thread_view", None):
```

## Testing

1. Create a minimal Chainlit app without defining `@cl.on_shared_thread_view`
2. Start the server
3. Access a shared thread URL
4. Verify no `UnboundLocalError` is raised

## Related Issue

Fixes #2766

---
## Summary by cubic
Initialize user_can_view to False before the shared-thread access check to prevent UnboundLocalError when @cl.on_shared_thread_view is not defined. This ensures apps without the hook handle shared thread views without crashing.

<sup>Written for commit 5e5e00c6b9fbc813e75bee13f7c39b117087deff. Summary will update on new commits.</sup>

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -1005,6 +1005,7 @@ async def get_shared_thread(
     if not isinstance(metadata, dict):
         metadata = {}
 
+    user_can_view = False
     if getattr(config.code, "on_shared_thread_view", None):
         try:
             user_can_view = await config.code.on_shared_thread_view(


dokterbob:
@thecaptain789 @veeceey This clearly seems a real issue. I don't really have bandwidth (in general) but would love to merge this if accompanied by a little regression test (e.g. a unit test failing on main, passing on your PR). That way I will be 100% sure it fixes an issue and IT WILL NEVER COME BACK. ;)

PR #2776, opened 2026-02-07
fix: initialize user_can_view before conditional block

### LINKED ISSUE #2766: UnboundLocalError: user_can_view referenced before assignment when on_shared_thread_view hook is not set (2.9.6)

thecaptain789:
Fixes #2766

When `@cl.on_shared_thread_view` hook is not defined, `user_can_view` was never initialized, causing an `UnboundLocalError` when the variable was referenced.

This initializes `user_can_view = False` before the conditional block to provide a safe default.

---
## Summary by cubic
Initialize user_can_view = False in get_shared_thread to avoid UnboundLocalError when on_shared_thread_view isn’t defined.

<sup>Written for commit f145ceb3fbc78dfe915df71087f71bd4d8bac6f7. Summary will update on new commits.</sup>

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -1005,6 +1005,7 @@ async def get_shared_thread(
     if not isinstance(metadata, dict):
         metadata = {}
 
+    user_can_view = False
     if getattr(config.code, "on_shared_thread_view", None):
         try:
             user_can_view = await config.code.on_shared_thread_view(


dokterbob:
Dup of #2775

PR #2785, opened 2026-02-18
feat: Add copilot sidebar mode with drag-to-resize

EyalAmitay:
## Summary

Adds a new **sidebar** display mode for the copilot widget. Instead of the default floating popover, the copilot can now render as a full-height side panel anchored to the right edge of the viewport, pushing the host page content to make room. Users can drag the left edge to resize and switch between sidebar and floating modes via a dropdown in the header.

image

## Changes

- **Sidebar display mode**: Right-side sidebar panel as an alternative to the floating popover
- **Drag-to-resize**: Left-edge drag handle (300px min, 50% viewport max) with smooth transitions
- **Display mode toggle**: Header dropdown to switch between "Floating" and "Sidebar"; close button in sidebar mode
- **Persistence**: Sidebar width and display mode saved to localStorage across sessions
- **Config-over-localStorage precedence**: Explicit `displayMode` passed via `mountChainlitWidget()` takes priority over a stale localStorage value; localStorage is only used as a fallback when `displayMode` is omitted
- **`resolveDisplayMode` helper**: Extracted display mode resolution logic into a standalone pure function (`libs/copilot/src/resolveDisplayMode.ts`)
- **`useSidebarResize` hook**: Encapsulates resize logic, body margin management, and drag lifecycle (including `window.blur` handling for edge cases)
- **`DisplayMode` type**: Consolidated into `types.ts`, replacing inline string unions

## Backward Compatibility

Fully backward compatible — no breaking changes to any existing API or behavior.

- The `displayMode` option in `IWidgetConfig` is **additive and optional**; defaults to `'floating'` when not specified, preserving existing behavior
- Existing `mountChainlitWidget()` calls without `displayMode` work unchanged
- `unmountChainlitWidget()` continues to work as before
- Two new localStorage keys are introduced (`chainlit-copilot-displayMode`, `chainlit-copilot-sidebarWidth`) — only read/written when the feature is used, no effect on existing deployments

### Configuration

To default to the sidebar mode:
```js
window.mountChainlitWidget({
  chainlitServer: 'http://localhost:8000',
  displayMode: 'sidebar',
});
```

## Test plan
- [x] E2E: Sidebar opens and pushes body content
- [x] E2E: Sidebar closes and restores body margin
- [x] E2E: Drag handle resizes sidebar and updates body margin
- [x] E2E: Switch between sidebar and floating mode
- [x] E2E: Unmount cleanup restores body margin
- [x] E2E: localStorage persistence survives widget remount
- [x] Unit: config `displayMode` takes precedence over localStorage
- [x] Unit: localStorage fallback when config omits `displayMode`
- [x] Unit: defaults to `floating` when neither config nor localStorage is set
- [x] Build and lint pass
- [x] Manual: Open, drag left edge to resize
- [x] Manual: State persists in localstorage

https://github.com/user-attachments/assets/7085f786-3705-400d-bf4e-27716f9e242a

image


EyalAmitay:
@hayescode

EyalAmitay:
@asvishnyakov

libs/copilot/src/widget.tsx
```diff
@@ -14,21 +14,34 @@ import { useConfig } from '@chainlit/react-client';
...
 }
 
+const LS_DISPLAY_MODE_KEY = 'chainlit-copilot-displayMode';
+
 const Widget = ({ config, error }: Props) => {
   const [expanded, setExpanded] = useState(config?.expanded || false);
   const [isOpen, setIsOpen] = useState(config?.opened || false);
+  const [displayMode, setDisplayMode] = useState<DisplayMode>(
+    () =>
+      (localStorage.getItem(LS_DISPLAY_MODE_KEY) as DisplayMode) ||
```
hayescode: This initialization order lets a previously saved localStorage value override an explicit `mountChainlitWidget({ displayMode })` setting. That means an embed can ask for `sidebar` and still come up `floating` because of an older user preference. Can we flip this so the explicit config wins, and only fall back to localStorage when `displayMode` is omitted?

EyalAmitay: Fixed - flipped the precedence so `config.displayMode` wins over localStorage. Also extracted the resolution logic into `resolveDisplayMode()` and added a focused regression test covering all three cases (config wins over localStorage, localStorage fallback when config omits displayMode, default to floating).

EyalAmitay:
@hayescode Thanks for the review! Both items addressed:

1. **Precedence flipped** -`config.displayMode` now takes priority over localStorage. localStorage is only used as a fallback when no explicit config is passed.
2. **Regression test added** - `frontend/tests/displayModePrecedence.spec.ts` covers config-wins-over-localStorage, localStorage-fallback, and default-to-floating cases.

Extracted the resolution logic into a standalone `resolveDisplayMode()` function (`libs/copilot/src/resolveDisplayMode.ts`) to keep the test focused and free of component dependencies.

libs/copilot/src/widget.tsx
```diff
@@ -108,20 +184,7 @@ const Widget = ({ config, error }: Props) => {
         )}
       >
         <div id="chainlit-copilot-chat" className="flex flex-col h-full w-full">
-          {error ? (
-            <Alert variant="error">{error}</Alert>
```
Cfh-GitHub, line 112: If the widget encountered an error (like a network failure), the original PR logic used error ? -<Alert> : <Header />..., meaning it entirely skipped rendering the <Header> component when an error was present.

In the old "floating" popover mode, this is somewhat okay because the user can click outside the popover to close it. However, this new "sidebar" mode renders as a fixed, full-height element with no click-outside handler. Without the <Header> component (which contains the close button), a user who encountered an error while in sidebar mode would be left permanently stuck with an uncloseable error overlay covering the side of their screen.

So, it may be a good idea to remove this ---error ? (alert variant=""error" ... --- line.

PR #2790, opened 2026-02-18
fix: prevent empty strings from overwriting step content in data layer

### LINKED ISSUE #2789: [Bug] Race condition in ChainlitDataLayer causes Steps to lose content in History
**Describe the bug**
When using a database via `DATABASE_URL` (Postgres), content within `cl.Step` occasionally disappears or appears empty when reloading a thread from history. This is caused by a race condition between the initial `step.send()` and the subsequent `step.update()`.

**To Reproduce**
It is caused because `cl.Step` initializes `input` and `output` as empty strings (`""`). In `ChainlitDataLayer.create_step`, and the SQL uses `COALESCE(EXCLUDED.output, "Step".output)`. Since empty string is not `NULL`, Postgres accepts the empty string as a valid update. If the background task for the initial send finishes after the final update, the content is overwritten by the empty string.

Script to reproduce this issue:
```python
import os
import asyncio
import chainlit as cl
from chainlit.data.chainlit_data_layer import ChainlitDataLayer

# Force the race condition by delaying empty saves
original_create_step = ChainlitDataLayer.create_step
async def laggy_create_step(self, step_dict):
    if step_dict.get("output") == "":
        await asyncio.sleep(2) # Task A waits
    return await original_create_step(self, step_dict)
ChainlitDataLayer.create_step = laggy_create_step

@cl.on_message
async def send_message():
    async with cl.Step(name="VulnerableStep") as step:
        await asyncio.sleep(0.1)
        step.output = "This content will be lost"
    await cl.Message(content="Refresh the page and check history.").send()
```

Set DATABASE_URL and authentication to enable threads. Perform `chainlit run`, send a message, then refresh the page.

**Expected behavior**
Step content should be persistent. The data layer should not allow an empty string to overwrite existing content in the database.

**Additional context**
Current behaviour: If the initial "start" task (Task A) commits to the DB after the "end" task (Task B), the step becomes empty in history.

Pranjalya:
fixes #2789 

**Description**
In the `ChainlitDataLayer`, the `create_step` method uses an `INSERT ... ON CONFLICT DO UPDATE` pattern with `COALESCE`. Because `Step` objects initialize `input` and `output` as `""` (empty string) instead of `None`, the database treats the empty string as a valid value to update, rather than a "missing" value to ignore.  

This PR ensures that `input` and `output` are passed as `None` (SQL `NULL`) if they are empty strings. This allows the existing `COALESCE` logic in the SQL query to correctly preserve existing content if a late-arriving "initial" save task executes after a "final" update task.  

**Changes**
Modified `backend/chainlit/data/chainlit_data_layer.py`:
- Updated the `params` dictionary in `create_step` to convert empty `input` and `output` strings to `None`.  

**Testing**
Verified using the reproduction script provided in the associated issue. With this change, the `COALESCE` function correctly identifies the `NULL` value and maintains the existing content even if the tasks finish out of order.

---
## Summary by cubic
Prevents empty input/output strings from overwriting existing step content by treating them as NULL in the data layer, preserving final step data when saves finish out of order.

- **Bug Fixes**
  - In ChainlitDataLayer.create_step, pass input/output as None when empty so SQL COALESCE keeps existing values; minor Ruff lint cleanup.

<sup>Written for commit 0ffb7f9324cec57dd58d08b30ed221765ef3f84a. Summary will update on new commits.</sup>


dokterbob:
Thanks @Pranjalya! It seems #2796 is a dup of this one, I only notice now.

Considering they're already implementing my feedback, I will close this one for now.

PR #2794, opened 2026-02-20
fix: use error detail for login credential error messages

### LINKED ISSUE #2787: Bug: On Logging-Screen for password authentication, the error message for incorrect credentials is missing or just the default.
**Describe the bug**
On Logging-Screen for password authentication, the error message for incorrect credentials is missing or just the default. Expected would be the error message for the translation key "auth.errors.credentialsSignin"

**To Reproduce**

1. create a password authentication hook.
2. start chainlit => Login-Screen is shown
3. enter incorrect credentials => no error message or error message auth.errors.default is displayed

**Expected behavior**
I would expect the error message for the translation key "auth.errors.credentialsSignin" to be displayed

**Additional context**
I already did some analysis and I'm pretty sure about the root cause.
Here https://github.com/Chainlit/chainlit/blob/bc292f69875f01dd239f1a9c072324c1f3febe66/frontend/src/components/LoginForm.tsx#L67 the LoginForm reads the message of the error, and uses this to lookup the translation.
But when the error is created in the fetch function https://github.com/Chainlit/chainlit/blob/bc292f69875f01dd239f1a9c072324c1f3febe66/libs/react-client/src/api/index.tsx#L143 this message is the statusTxt, while the actual reason (in this case credentialsSignin), is passed as "detail". 
So probably the fix is to read the detail in the LoginForm instead of the error message, in the case where the Error is a ClientError

hztBUAA:
## Summary
Fixes #2787

The login form displayed the wrong error message for incorrect credentials because it read `error.message` (HTTP status text like "Unauthorized") instead of `error.detail` (the actual error reason like "credentialsSignin").

## Changes
- Import `ClientError` from `@chainlit/react-client` in `LoginForm.tsx`
- Updated the catch block to check for `ClientError.detail` first, which contains the meaningful error reason (e.g., "credentialsSignin")
- Falls back to `Error.message` when `detail` is not available, preserving backward compatibility

## How it works

When the `/login` endpoint returns an error, the `fetch` wrapper in `api/index.tsx` creates a `ClientError` with:
- `message` = HTTP status text (e.g., "Unauthorized")
- `detail` = parsed response body detail (e.g., "credentialsSignin")

Previously, `LoginForm` used `err.message` for the translation key lookup (`auth.login.errors.unauthorized`), which didn't match any translation. Now it uses `err.detail` (`auth.login.errors.credentialssignin`), which correctly maps to the intended error message.

## Test Plan
- [x] Enter incorrect credentials on the login screen -> proper translated error message is displayed
- [x] Non-ClientError exceptions still display `error.message` as before
- [x] ClientError without detail falls back to `error.message`

---
## Summary by cubic
Fixes #2787. Use ClientError.detail in LoginForm so invalid credential attempts display the correct translated error, falling back to Error.message when detail is missing.

<sup>Written for commit 18a81871ba08e5faf964b29ca82fc2ba4a0c7a60. Summary will update on new commits.</sup>

### DIFF
--- a/frontend/src/components/LoginForm.tsx
+++ b/frontend/src/components/LoginForm.tsx
@@ -8,6 +8,8 @@ import { Input } from '@/components/ui/input';
 import { Label } from '@/components/ui/label';
 import Translator, { useTranslation } from 'components/i18n/Translator';
 
+import { ClientError } from '@chainlit/react-client';
+
 import Alert from './Alert';
 import { ProviderButton } from './ProviderButton';
 
@@ -63,7 +65,9 @@ export function LoginForm({
     try {
       await onPasswordSignIn(data.email, data.password, callbackUrl);
     } catch (err) {
-      if (err instanceof Error) {
+      if (err instanceof ClientError && err.detail) {
+        setErrorState(err.detail);
+      } else if (err instanceof Error) {
         setErrorState(err.message);
       }
     } finally {


dokterbob:
Does this close/address #2787 ?

PR #2796, opened 2026-02-20
fix: prevent race condition in step content persistence

### LINKED ISSUE #2789: [Bug] Race condition in ChainlitDataLayer causes Steps to lose content in History

hztBUAA:
## Summary
Fixes #2789

Race condition in `ChainlitDataLayer` causes Steps to lose content when the initial `step.send()` (with empty output) finishes after the final `step.update()` (with real output). The `COALESCE` in the upsert SQL treats empty string as a valid value, so it overwrites the actual content instead of preserving it.

### Root cause

1. `cl.Step` initializes `input` and `output` as empty strings (`""`)
2. The `create_step` SQL uses `COALESCE(EXCLUDED.output, "Step".output)` to preserve existing values
3. `COALESCE` only skips `NULL`, not empty strings — so an empty string from the initial `send()` overwrites real content written by `update()`

### Fix

Convert empty string `input`/`output` to `NULL` before passing to the database query. This allows the existing `COALESCE` logic to correctly preserve content when a late-arriving `send()` would otherwise overwrite it.

```python
# Before
"input": step_dict.get("input"),
"output": step_dict.get("output"),

# After
"input": step_dict.get("input") or None,
"output": step_dict.get("output") or None,
```

## Test plan
- [x] All existing step tests pass (45/45)
- [x] All existing data layer tests pass (35/35, excluding unrelated missing optional deps)
- [x] Empty string input/output is now treated as NULL in DB, so COALESCE correctly falls back to existing content
- [x] Non-empty input/output values are persisted normally (truthy strings pass through `or None` unchanged)

---
## Summary by cubic
Prevents step content loss when send() (empty output) finishes after update() (real output). Empty input/output are now stored as NULL so the upsert COALESCE preserves existing content.

- **Bug Fixes**
  - Convert empty input/output to NULL before DB upsert in ChainlitDataLayer.create_step.
  - Stops late send() from overwriting real output; non-empty values persist normally.

<sup>Written for commit 4544aa86f769e46273852480faefc72dc234fb9c. Summary will update on new commits.</sup>


backend/chainlit/data/chainlit_data_layer.py
```diff
@@ -378,10 +378,10 @@ async def create_step(self, step_dict: StepDict):
             "id": step_dict["id"],
             "thread_id": step_dict.get("threadId"),
             "parent_id": step_dict.get("parentId"),
-            "input": step_dict.get("input"),
+            "input": step_dict.get("input") or None,
```
dokterbob, line 381: Seems this would be a more elegant approach.

```suggestion
            "input": step_dict.get("input", None)
```

dokterbob:
@hztBUAA Would be GREAT to have a regression (unit) test demonstrating the issue on main (e.g. the test fails), then passing on your PR!

hztBUAA:
Thanks for the review and feedback. I am following up on this PR now and will either push the requested changes or reply point-by-point shortly.

hztBUAA:
Quick follow-up: I am reviewing the feedback and will update this PR shortly.

PR #2808, opened 2026-03-01
fix: treat empty custom_fonts array as intent to remove default fonts

### LINKED ISSUE #2804: theme.json: custom_fonts: [] should remove default Inter Google Font
## Description

When using `theme.json` with self-hosted fonts loaded via `@font-face` in `custom_css`, the default Inter font from Google Fonts is still loaded — there is no way to disable it.

The built `index.html` contains:
```html
<!-- FONT START -->
<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;700&display=swap" />
<!-- FONT END -->
```

In `server.py`, the replacement logic only triggers when `custom_fonts` is truthy:
```python
if custom_theme and custom_theme.get("custom_fonts"):
    font = "\n".join(...)
```

Setting `custom_fonts: []` in `theme.json` is falsy in Python, so the default Inter `<link>` tags are never removed.

## Use case

We are adopting the French Government Design System (DSFR) which requires the **Marianne** typeface, self-hosted from `/public/dsfr/fonts/`. We load it via `@font-face` in our `custom_css` and set `--font-sans` to `'Marianne', ...` in `theme.json` variables.

The font works, but the browser still makes an unnecessary request to `fonts.googleapis.com` for Inter, which:
1. Wastes bandwidth
2. Breaks in air-gapped / government network environments where Google domains may be blocked
3. Creates a privacy concern (sends user IPs to Google)

## Expected behavior

When `custom_fonts` is present in `theme.json` — even as an empty array `[]` — the default Inter `<link>` tags between `<!-- FONT START -->` and `<!-- FONT END -->` should be **removed**, since the developer is signaling they manage fonts themselves.

## Suggested fix

In `server.py` `get_html_template()`, change the condition from:
```python
if custom_theme and custom_theme.get("custom_fonts"):
```
to:
```python
if custom_theme and "custom_fonts" in custom_theme:
```

Then when `custom_fonts` is `[]`, the `font` variable would be `""`, and `replace_between_tags` would effectively clear the Inter tags. When `custom_fonts` has URLs, they'd replace Inter as before. When `custom_fonts` is absent, the default Inter is preserved (backward-compatible).

## Environment

- Chainlit version: 2.9.6
- OS: macOS
- Python: 3.13

👾 Generated with [Letta Code](https://letta.com)

Br1an67:
## Problem

When `custom_fonts` is set to `[]` in `theme.json`, the user intends to manage fonts themselves (e.g., via `@font-face` in `custom_css`). However, the default Inter Google Font `<link>` tags are still injected because `[]` is falsy in Python, so the condition `custom_theme.get("custom_fonts")` evaluates to `False`.

This causes:
- Unnecessary network requests to `fonts.googleapis.com`
- Breakage in air-gapped/government environments where Google domains are blocked
- Privacy concerns (user IPs sent to Google)

## Fix

- Check for key **presence** (`"custom_fonts" in custom_theme`) instead of **truthiness**
- Use `if font is not None` instead of `if font` so an empty string properly clears the default font tags via `replace_between_tags`

When `custom_fonts` is:
- **absent** → default Inter preserved (backward-compatible)
- **`[]`** → default Inter removed (user manages fonts)
- **`[{...}]`** → custom fonts replace Inter (existing behavior)

Fixes #2804

---
## Summary by cubic
Stop injecting the default Inter font when theme.json sets custom_fonts to []. An empty array now removes the Inter links, reducing unwanted Google requests and supporting air-gapped environments. Fixes #2804.

- **Bug Fixes**
  - Check for key presence ("custom_fonts" in custom_theme) instead of truthiness so: absent → keep Inter, [] → remove Inter, [{...}] → use custom links.
  - Use if font is not None when replacing the FONT block so an empty string clears the tags.

<sup>Written for commit 0f69d69efee234f8cf86e4f6ad7cce7be54bcabe. Summary will update on new commits.</sup>

### DIFF
--- a/backend/chainlit/server.py
+++ b/backend/chainlit/server.py
@@ -410,10 +410,10 @@ def get_html_template(root_path):
         js += f"""<script src="{config.ui.custom_js}" {config.ui.custom_js_attributes}></script>"""
 
     font = None
-    if custom_theme and custom_theme.get("custom_fonts"):
+    if custom_theme and "custom_fonts" in custom_theme:
         font = "\n".join(
-            f"""<link rel="stylesheet" href="{font}">"""
-            for font in custom_theme.get("custom_fonts")
+            f"""<link rel="stylesheet" href="{f}">"""
+            for f in custom_theme["custom_fonts"]
         )
 
     index_html_file_path = os.path.join(build_dir, "index.html")
@@ -425,7 +425,7 @@ def get_html_template(root_path):
             content = content.replace(JS_PLACEHOLDER, js)
         if css:
             content = content.replace(CSS_PLACEHOLDER, css)
-        if font:
+        if font is not None:
             content = replace_between_tags(
                 content, "<!-- FONT START -->", "<!-- FONT END -->", font
             )


hayescode (approved):

Br1an67:
Still active — happy to address any review feedback.

PR #2813, opened 2026-03-01
feat: add download_name parameter for custom download filenames

### LINKED ISSUE #2124: Custom file name for cl.Pdf
**Is your feature request related to a problem? Please describe.**
Currently, when using cl.Pdf with an existing file path, Chainlit creates a copy of the file with a new UUID-based name. This UUID is then used as the download file name when a user downloads the PDF from the chat UI, which means the original file name is lost. This makes it difficult for users to identify and manage downloaded files.

**Describe the solution you'd like**
I propose adding a way to customize the file name for downloads triggered from cl.Pdf elements. Ideally, this would involve:
A new parameter, such as download_name, in the cl.Pdf class.
This parameter would allow developers to specify the file name that users will see when downloading the PDF.

For example:
```
pdf_element = cl.Pdf(
    name="My Report",
    path="./my_report.pdf",
    display="inline",
    download_name="UserFriendlyReport.pdf"  # New parameter
)
```

This would ensure that when a user downloads the PDF, the file is saved as "UserFriendlyReport.pdf" instead of a UUID.

**Benefits:**
- Improved user experience: Users can easily identify downloaded files.
- Better file management: Consistent and meaningful file names facilitate organization.
- Increased flexibility: Developers can tailor file names to specific use cases.

Br1an67:
## What
Add a `download_name` optional parameter to `Element` so developers can specify a human-readable filename for file downloads.

Closes #2124

## Why
When using `cl.Pdf` with a file path, Chainlit persists the file with a UUID-based name. When users download the PDF from the browser's viewer, they get a UUID filename instead of the original name.

## Changes
- `backend/chainlit/element.py`: Add `download_name: Optional[str] = None` to `Element` base class, pass it through `persist_file()`
- `backend/chainlit/session.py`: Accept and store `download_name` in the file dict
- `backend/chainlit/server.py`: Use `download_name` (falling back to `name`) in `FileResponse` with `content_disposition_type="inline"` to set the download filename while preserving inline display

## Usage
```python
pdf = cl.Pdf(
    name="My Report",
    path="./report.pdf",
    display="inline",
    download_name="UserFriendlyReport.pdf"
)
```

Without `download_name`, the element's `name` is used as the download filename (still better than the UUID).

## Testing
Backward compatible — `download_name` defaults to `None`, falling back to existing `name` field.

---
## Summary by cubic
Add a download_name option to Element so files (e.g., PDFs) download with a human-friendly filename while keeping inline viewing. The server sets an inline Content-Disposition and filename using download_name, falling back to name.

- **New Features**
  - Element.download_name optional field; serialized as downloadName, restored via to_dict/from_dict, persisted via persist_file, and added to FileDict.
  - File responses set inline Content-Disposition with filename from download_name (fallback to name).

<sup>Written for commit c15c2757d39130d7044ee384bb92a770a0f19ce9. Summary will update on new commits.</sup>


hayescode (dismissed):

Br1an67:
Hi — just a gentle bump on this. Happy to make any changes if needed!

backend/chainlit/server.py
```diff
@@ -1658,7 +1658,13 @@ async def get_file(
 
     if file_id in session.files:
         file = session.files[file_id]
-        return FileResponse(file["path"], media_type=file["type"])
+        filename = file.get("download_name") or file.get("name")
+        return FileResponse(
+            file["path"],
+            media_type=file["type"],
+            filename=filename,
```
hayescode, line 2107: This looks like the right behavior, but since it changes the response headers for `/project/file/{file_id}`, I?d like a targeted server test before we merge it. Please cover both cases: `download_name` present, and fallback to `name` when it isn?t.

hayescode (changes requested):
Please add targeted server coverage for the new download filename behavior before approval.

PR #2824, opened 2026-03-04
fix: show error toast when streamable-http MCP connection fails

### LINKED ISSUE #2823: Bug: Frontend shows "MCP added!" success toast when streamable-http connection fails
## Bug: Frontend shows "MCP added!" success toast when streamable-http connection fails

**Describe the bug**
When connecting a streamable-http MCP server with an invalid URL, the backend correctly returns a 400 error, but the frontend displays a green "MCP added!" success toast. The "My MCPs" tab then shows "No MCP servers connected", which contradicts the success toast.

SSE connections with the same invalid input correctly show an error toast.

**To Reproduce**
1. Open the MCP Servers dialog
2. Select type "streamable-http"
3. Enter any invalid URL (e.g. "abc")
4. Click Confirm
5. Observe: green "MCP added!" toast appears, but no server is listed under "My MCPs"

**Expected behavior**
The frontend should show an error toast (consistent with SSE behavior), since the backend returns HTTP 400:
```json
{"detail": "Could not connect to the MCP: unhandled errors in a TaskGroup (1 sub-exception)"}
```

**Actual behavior**
- SSE with invalid URL → error toast (correct)
- streamable-http with invalid URL → "MCP added!" success toast (incorrect)

**Environment**
- **Chainlit version**: 2.9.6
- **MCP SDK version**: 1.26.0
- **Python version**: 3.13+

**Additional context**
The backend behavior is correct in both cases (returns 400). This appears to be a frontend-only issue in how the streamable-http response is handled compared to SSE.

Related: #2147 (MCP URL validation)

weiguangli-io:
## Summary

Fixes #2823

When connecting a streamable-http MCP server with an invalid URL, the frontend incorrectly shows a green "MCP added!" success toast even though the backend returns HTTP 400. SSE and stdio connections correctly display an error toast in the same scenario.

**Root cause:** `ExtendedChainlitAPI` in `frontend/src/api/index.ts` overrode `connectStreamableHttpMCP()` using raw `fetch()` instead of the inherited `this.post()` method. The raw `fetch` call never threw on HTTP errors — it always resolved the promise with `{ success: false, ... }`. Since `toast.promise()` only shows an error when the promise rejects, the success toast was always displayed.

**Fix:** Remove the unnecessary override. The base class `ChainlitAPI` (in `@chainlit/react-client`) already provides a correct `connectStreamableHttpMCP()` implementation that uses `this.post()` → `this.fetch()`, which throws a `ClientError` on non-ok HTTP responses. This is the same pattern used by `connectSseMCP()` and `connectStdioMCP()`, which already work correctly.

## Changes

- **`frontend/src/api/index.ts`**: Removed the `connectStreamableHttpMCP()` override from `ExtendedChainlitAPI`, letting the base class implementation handle error propagation correctly.

## Test plan

- [ ] Open the MCP Servers dialog, select "streamable-http", enter an invalid URL (e.g., "abc"), and click Confirm — should now show a **red error toast** instead of a green "MCP added!" toast
- [ ] Repeat with a valid streamable-http MCP URL — should show a green "MCP added!" toast and list the server under "My MCPs"
- [ ] Verify SSE and stdio connections still work correctly (no regression)

---
## Summary by cubic
Failed streamable-http MCP connections now show a red error toast instead of a success message. Fixes #2823.

- **Bug Fixes**
  - Removed ExtendedChainlitAPI.connectStreamableHttpMCP override to use the ChainlitAPI implementation.
  - Non-OK HTTP responses now reject (ClientError), so toast.promise shows the error.
  - SSE and stdio connection flows unchanged.

<sup>Written for commit 760944586a4beca9043c6bfa9276499b3a487e75. Summary will update on new commits.</sup>

### DIFF
--- a/frontend/src/api/index.ts
+++ b/frontend/src/api/index.ts
@@ -34,31 +34,6 @@ class ExtendedChainlitAPI extends ChainlitAPI {
     return res.json();
   }
 
-  connectStreamableHttpMCP(
-    sessionId: string,
-    name: string,
-    url: string,
-    headers?: Record<string, string>
-  ) {
-    // Assumes the backend expects { clientType, name, url }
-    return fetch(new URL("mcp", this.httpEndpoint.endsWith("/") ? this.httpEndpoint : `${this.httpEndpoint}/`), {
-      method: 'POST',
-      headers: {
-        'Content-Type': 'application/json',
-        ...(sessionId ? { 'x-session-id': sessionId } : {})
-      },
-      body: JSON.stringify({
-        clientType: 'streamable-http',
-        name,
-        url,
-        sessionId,
-        ...(headers ? { headers } : {})
-      })
-    }).then(async (res) => {
-      const data = await res.json();
-      return { success: res.ok, mcp: data.mcp, error: data.detail };
-    });
-  }
 }
 
 export const apiClient = new ExtendedChainlitAPI(


weiguangli-io:
Closing — shifting focus to AI Agent core projects. Thank you for your time\!

PR #2825, opened 2026-03-05
fix: scope cookies to CHAINLIT_ROOT_PATH for multi-app deployments

gonzalo123:
## Summary

- Fix bug where `_cookie_path` used `CHAINLIT_ROOT_PATH` value as an env var name instead of using it directly as the cookie path
- Add `path=_cookie_path` to all `set_cookie` calls that were missing it (`access_token`, `oauth_state`, `X-Chainlit-Session-id`)
- Add `path=_cookie_path` to `clear_oauth_state_cookie`'s `delete_cookie` call

## Problem

When multiple Chainlit apps are mounted on the same domain under different subpaths (e.g., `/app1` and `/app2`), their cookies collide because all cookies are set with `path="/"`. This causes 401 Unauthorized errors when switching between browser tabs running different apps.

The root cause is twofold:
1. `_cookie_path` resolution has a bug: `os.environ.get(_cookie_root_path, "/")` treats the value (e.g., `/app1`) as an env var name
2. Most `set_cookie` calls never pass `path=`, so cookies always default to `path="/"`

## Changes

- `backend/chainlit/auth/cookie.py`: Fix `_cookie_path` resolution and add `path=_cookie_path` to all `set_cookie`/`delete_cookie` calls
- `backend/chainlit/server.py`: Import `_cookie_path` and use it for the `X-Chainlit-Session-id` cookie
- `backend/tests/auth/test_cookie.py`: Add 3 tests for `_cookie_path` resolution logic

## Backwards Compatibility

No breaking changes. When `CHAINLIT_ROOT_PATH` is not set, `_cookie_path` remains `"/"` — identical to current behavior.

---
## Summary by cubic
Scope all Chainlit cookies to the app’s root path to prevent cross-app collisions on the same domain. Also delete legacy root-path cookies to avoid stale auth after upgrading.

- **Bug Fixes**
  - Fix _cookie_path: use CHAINLIT_ROOT_PATH directly; CHAINLIT_AUTH_COOKIE_PATH can override; default is "/".
  - Scope all set/delete cookie calls with path=_cookie_path (auth chunk/single, oauth_state, X-Chainlit-Session-id) and delete legacy Path="/" copies, including leftover chunks; add TODO to remove this cleanup in the next major release.
  - Add tests for path defaults/overrides, legacy cleanup (exact Path=/ match), and session cookie path.

<sup>Written for commit 14c30fb3ebe1bed7138b707839ecc835571f6608. Summary will update on new commits.</sup>


backend/chainlit/auth/cookie.py
```diff
@@ -34,6 +34,20 @@
 _state_cookie_name = "oauth_state"
 
 
+def _delete_legacy_cookies(response: Response, *names: str):
+    """Delete cookies at path='/' left over from pre-scoped versions.
+
+    Only acts when _cookie_path != '/' to avoid no-op deletes in
+    single-app deployments.
```
dokterbob, line 42: I would suggest adding a comment here explaining why it was introduced and a suggestion when it may be removed (e.g. next major release).

dokterbob (changes requested):
Other than minor comment (when do we remove legacy support) LGTM!

PR #2833, opened 2026-03-08
feat: added improved pdf viewer

JonasKenke:
## Summary
This change replaces the previous iframe-based browser PDF viewer with a new, enhanced PDF-viewer component and adds the associated support in both frontend and e2e test suites.

## Validation
All tests passed *✔  All specs passed!                        06:27*

## What changed
- Replaced the old iframe‑based browser PDF viewer with a custom React viewer.
- Added new PDF component and wired it into message/side‑view code.
- Updated tests: new Cypress spec, dummy PDF, and setup helpers.
- Bumped frontend deps and adjusted .npmrc

### Old:
Screenshot_20260308_204628

### New:
Screenshot_20260308_210416
Screenshot_20260308_210435

---
## Summary by cubic
Replaced the iframe PDF viewer with a React-based viewer that supports inline previews, a fullscreen modal, and a side-panel that auto-opens only for new or changed elements. Ships a local `pdfjs-dist` worker (no CDN) with e2e coverage.

- New Features
  - Custom viewer built with `react-pdf` and a toolbar for zoom, pagination, download, and print.
  - Inline thumbnail opens a fullscreen modal; side-display PDFs render with the viewer and auto-open the panel only when new/changed.
  - Local `pdfjs-dist` worker is bundled and hoisted via `.npmrc` (no external CDN).

- Bug Fixes
  - Serve `.mjs` as JavaScript to prevent MIME errors.
  - Improved scrolling and layout: switched to overflow-auto, removed fixed inline height, and increased default side panel width.
  - Test setup polyfills `DOMMatrix` for `pdfjs` in JSDOM; e2e asserts the local worker loads with zero CDN requests.

<sup>Written for commit cfd840fc51634409b0afc39f5ef02852d903bdbf. Summary will update on new commits.</sup>

### DIFF
--- a/.github/actions/pnpm-node-install/action.yaml
+++ b/.github/actions/pnpm-node-install/action.yaml
@@ -5,7 +5,7 @@ inputs:
   node-version:
     description: Node.js version
     required: true
-    default: '24.3.0' # Switch to 'lts' as soon as Node 24 reaches LTS status.
+    default: '24.5.0' # Includes npm 11.5.1+, which is required for npm trusted publishing.
 
 runs:
   using: composite
--- a/.github/workflows/publish-libs.yaml
+++ b/.github/workflows/publish-libs.yaml
@@ -50,5 +50,3 @@ jobs:
       - name: Publish packages to npm
         # --no-git-checks allows testing from non-main branches and publishing from release tags
         run: pnpm publish --recursive --no-git-checks ${{ inputs.dry_run && '--dry-run' || '' }}
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_REACT_CLIENT }}
--- a/libs/react-client/package.json
+++ b/libs/react-client/package.json
@@ -11,7 +11,8 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/Chainlit/"
+    "url": "git+https://github.com/Chainlit/chainlit.git",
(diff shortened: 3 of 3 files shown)


Allaoua9:
Thanks for putting this together — I can see the effort that went into improving the PDF experience here.

I do have a few questions to better understand the trade-offs.

What are the main limitations of the native browser PDF viewer that this change is trying to address? In my experience, Chrome’s built-in PDF viewer is already quite good, and Firefox’s native experience is also solid. I dont know about other browsers though.

Could you clarify which concrete user-facing features this PR adds over the current iframe/native approach? That would help us better understand the value of introducing and maintaining a custom viewer.

One thing I liked about the current iframe approach is that it could point to PDFs behind authenticated URLs (Sharepoint for example would request authentication or authorization). Am I right in thinking that this new approach would make that harder or not support it ?

JonasKenke:
Hey,
Thanks for checking out the pr. So my main goal was to provide a unified look of the PDF viewer on all browser. Also the old inline PDF viewer was often hard to read and now you get a small preview and can get a pop-up like for the images. The main goal was to have a unified experience so it is similar to the image inline. For me also cosmetic wise as I like a custom PDF viewer that matches the app it self more. In addition in my tests the I frame based PDF viewer had some issues with responsive ness how it's currently implemented.

For the authenticated urls I'm not sure but will check and I'm open for improvements

PR #2834, opened 2026-03-09
fix: Issue with accordion requiring two clicks to close

Allaoua9:
Accordion was not closing on the first click. it required two clicks to close.

This issue was that the controlled accordion was incorrectly handled.

---
## Summary by cubic
Fixes the step accordion so it closes on the first click by standardizing the controlled value to a string. The closed state now uses `''` instead of `undefined`, and `onValueChange` sets the value directly, making user toggles and auto-collapse work reliably.

<sup>Written for commit d3fbc94a4c4557c582eb5dd72f49d9c25df808b2. Summary will update on new commits.</sup>

### DIFF
--- a/frontend/src/components/chat/Messages/Message/Step.tsx
+++ b/frontend/src/components/chat/Messages/Message/Step.tsx
@@ -29,14 +29,14 @@ export default function Step({
   const isError = step.isError;
   const stepName = step.name;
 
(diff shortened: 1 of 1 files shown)


Allaoua9:
Can we reopen this as this is a serious UX issue ? @hayescode

PR #2843, opened 2026-03-12
fix: resolve base locale to regional translation file (e.g. "da" → "da-DK")

### LINKED ISSUE #2842: bug: Translation fallback doesn't resolve base locale to regional file (e.g. "da" → "da-DK")
## Description

The `load_translation` method in `config.py` only falls back from regional to base locale (`da-DK` → `da`), but never resolves a base locale to an available regional file (`da` → `da-DK`). This means most users with regional translation files never actually see them.

## Root Cause

In [`backend/chainlit/config.py`](https://github.com/Chainlit/chainlit/blob/main/backend/chainlit/config.py#L479-L521), the fallback logic is:

```python
parent_language = language.split("-")[0]  # "da" → "da"

# 1. Try exact match:  da.json
# 2. Try parent:       da.json  ← identical when input has no region!
# 3. Fall back to:     en-US.json
```

When a browser sends `da` (which is the default for most Danish browsers), steps 1 and 2 both look for `da.json`. Neither step ever tries `da-DK.json`, even though that file exists.

## Impact

This affects **all** translations that use regional codes:

| File | Browser sends | Found? |
|------|--------------|--------|
| `es.json` | `es` | Yes |
| `nl.json` | `nl` | Yes |
| `ja.json` | `ja` | Yes |
| `da-DK.json` | `da` | **No** |
| `de-DE.json` | `de` | **No** |
| `fr-FR.json` | `fr` | **No** |
| `zh-CN.json` | `zh` | **No** |
| `he-IL.json` | `he` | **No** |
| `el-GR.json` | `el` | **No** |

Users affected by this silently fall back to `en-US` with a warning in the logs:

```
WARNING - chainlit - Translation file for da not found. Using default translation en-US.
```

In a production deployment, this warning appeared **443 times in 4 hours** — once per page load for every Danish user.

## Suggested Fix

Add an upward lookup step that scans for `{base_locale}-*.json` when the exact and parent lookups fail:

```python
def load_translation(self, language: str):
    parent_language = language.split("-")[0]
    translation_dir = Path(config_translation_dir)

    # 1. Exact match (e.g. "da-DK.json")
    exact = translation_dir / f"{language}.json"
    if is_path_inside(exact, translation_dir) and exact.is_file():
        return json.loads(exact.read_text(encoding="utf-8"))

    # 2. Parent/base language (e.g. "da.json")
    parent = translation_dir / f"{parent_language}.json"
    if is_path_inside(parent, translation_dir) and parent.is_file():
        logger.warning(f"Translation file for {language} not found. Using parent translation {parent_language}.")
        return json.loads(parent.read_text(encoding="utf-8"))

    # 3. Regional variant (e.g. "da" → find "da-DK.json")
    if language == parent_language:
        for candidate in sorted(translation_dir.glob(f"{parent_language}-*.json")):
            if is_path_inside(candidate, translation_dir) and candidate.is_file():
                variant = candidate.stem
                logger.info(f"Translation file for {language} not found. Using regional variant {variant}.")
                return json.loads(candidate.read_text(encoding="utf-8"))

    # 4. Default fallback
    default = translation_dir / f"{DEFAULT_LANGUAGE}.json"
    if is_path_inside(default, translation_dir) and default.is_file():
        logger.warning(f"Translation file for {language} not found. Using default translation {DEFAULT_LANGUAGE}.")
        return json.loads(default.read_text(encoding="utf-8"))

    return {}
```

## Environment

- **Chainlit version:** 2.10.0
- **Python version:** 3.13
- **OS:** Linux (AKS)

## Workaround

Setting `language = "da-DK"` in `config.toml` under `[UI]` forces the specific locale, but this overrides the language for **all** users, including non-Danish speakers.

Alternatively, deploying a copy of `da-DK.json` as `da.json` works but shouldn't be necessary.

mgquach:
## Summary

Fixes #2842

The `load_translation` fallback chain only searched for `{language}.json`
and `{parent_language}.json`, which are **identical** when the input is
already a base locale (e.g. `"da"`). This meant `da-DK.json` was never
discovered, and all Danish users silently received `en-US` translations.

## Changes

Added a third lookup step in `ChainlitConfig.load_translation()`:

| Step | Input | Tries | Condition |
|------|-------|-------|-----------|
| 1 | `da` | `da.json` | always |
| 2 | `da` | `da.json` (parent) | always |
| **3** | **`da`** | **`da-DK.json` (glob)** | **only when input has no region** |
| 4 | `da` | `en-US.json` (default) | always |

Step 3 uses `sorted()` for deterministic selection when multiple regional
variants exist (e.g. `zh-CN.json` vs `zh-TW.json`). All existing
`is_path_inside` security checks are preserved.

## Affected locales (built-in translations)

Without this fix, these locales never resolved when browser sent the base code:

| File | Browser sends | Before | After |
|------|--------------|--------|-------|
| `de-DE.json` | `de` | en-US | de-DE ✓ |
| `fr-FR.json` | `fr` | en-US | fr-FR ✓ |
| `zh-CN.json` | `zh` | en-US | zh-CN ✓ |
| `he-IL.json` | `he` | en-US | he-IL ✓ |
| `el-GR.json` | `el` | en-US | el-GR ✓ |

## Testing

Existing behaviour is unchanged for:
- exact matches (`fr-FR` → `fr-FR.json`)
- parent fallback (`de-DE` → `de.json` if it exists)
- default fallback (unknown locale → `en-US.json`)

---
## Summary by cubic
Fix base-locale fallbacks so a language like "da" resolves to a regional file like "da-DK.json" instead of defaulting to "en-US". Add regression tests for the full fallback chain.

- **Bug Fixes**
  - In `ChainlitConfig.load_translation()`, add a third step: when the input has no region and no exact/base file exists, scan `{base}-*.json` and load the first sorted match. Exact, parent, and default fallbacks stay the same. Examples: `de` → `de-DE`, `fr` → `fr-FR`, `zh` → `zh-CN`.
  - Add tests for exact match, parent fallback, regional variant lookup (deterministic), and default fallback.

<sup>Written for commit f4940e198d0d5a2a89febde9bb056c2d7638da10. Summary will update on new commits.</sup>


backend/chainlit/config.py
```diff
@@ -492,7 +493,22 @@ def load_translation(self, language: str):
             translation = json.loads(
                 translation_lib_parent_language_file_path.read_text(encoding="utf-8")
             )
-        elif (
+            return translation
+
+        # 3. Regional variant lookup (e.g. "da" → "da-DK.json")
+        if language == parent_language:
+            for candidate in sorted(translation_dir.glob(f"{parent_language}-*.json")):
```
hayescode, line 515: This fallback change looks reasonable, but because it changes locale resolution behavior in config code, I?d like to see targeted backend coverage before we merge it. Please add regression tests covering exact match, parent fallback, regional fallback from a base locale, and default fallback.

PR #2844, opened 2026-03-14
fix: handle Slack workflow/bot messages in app_mention events

sefadegirmenci:
Workflow-triggered app_mention events lack a "user" field (they have "bot_id" instead), causing a KeyError that silently killed message processing after the :eyes: reaction was added.

Branch on the event payload to determine human vs bot, then call the appropriate Slack API (users_info vs bots_info) upfront instead of relying on exception-driven control flow.

---
## Summary by cubic
Handle Slack app_mention events from workflows and bots by using bot_id to fetch the right profile. Ignore bot/workflow messages in DMs and fix a `mypy` error in bot profile lookup to prevent crashes and misclassifying bots.

- **Bug Fixes**
  - Handle events without `user`: branch to `users_info` or `bots_info`; fallback to `_get_bot_profile` on `SlackApiError`.
  - Drop bot/workflow messages in the DM handler; only human DMs are processed.
  - Fix `mypy` call-overload in `_get_bot_profile` by splitting chained `get` calls.

- **Refactors**
  - Formatted `backend/chainlit/slack/app.py` with `ruff`.

<sup>Written for commit a5db861021120201bb22fdcd9cf68928f56e652f. Summary will update on new commits.</sup>


sefadegirmenci:
@hayescode @asvishnyakov This fixes the problem of accepting mentions from bots on slack. Let's say the bot is named chainlit, another bot can mention chainlit to delagate issues. 

Review & approve pls when you are available

efealioksuz (approved):

efealioksuz:
This fixes a real silent failure, the 👀 reaction would fire and then nothing. Branching on bot_id vs user upfront and calling the right Slack API accordingly is the correct approach. The DM handler guard is a nice defensive touch too. LGTM, approving!

sefadegirmenci:
@willydouhard Could you take a look at this pr? It fixes an important problem we encounter within slack

sefadegirmenci:
> Reviewed via Codex
> 
> LGTM.

Thx @hayescode for the review. i don't think i have the merge permission

PR #2845, opened 2026-03-16
Feat/Fix : Dataframe element fix and improvements

Allaoua9:
This PR was implemented in collaboration with [@tonca](https://github.com/tonca), many thanks to him !

* Fixed issue with horizontal overflow in the pagination footer
* Improved the Dataframe element by adding filters, sorting, column selection and ability to select the number of lines to display
* New features are optional and disabled by default

Capture d&#39;écran 2026-03-16 155342
Capture d&#39;écran 2026-03-16 155415
Capture d&#39;écran 2026-03-16 155421

---
## Summary by cubic
Upgrades the Dataframe element with per-column filters, column visibility, page size control, expandable rows, and a clear-filters action for easier exploration. Also fixes pagination footer overflow, aligns frontend/backend defaults for new options, and safely handles None data.

- **New Features**
  - Per-column filters and sorting in headers.
  - Column visibility dropdown with select/deselect all; controllable from Python via `Dataframe(show_column_visibility=..., show_column_filters=...)` and exposed as `showColumnVisibility`/`showColumnFilters` in `@chainlit/react-client` (off by default).
  - Page size selector (5/10/20/50) and a compact rows count indicator.
  - Expandable rows with a full details view on click.
  - New i18n strings for the Dataframe UI across locales, including a localized “no results” state.

- **Bug Fixes**
  - Fixed horizontal overflow in the pagination footer.
  - Backend: handle `data=None` in `Dataframe.__post_init__` and add (de)serialization for the new visibility/filter flags to keep frontend and backend in sync.

<sup>Written for commit c583629ee802f9c3004de8fbbac20f7810d4372d. Summary will update on new commits.</sup>


frontend/src/components/Elements/Dataframe.tsx
```diff
@@ -38,28 +62,85 @@ interface DataframeData {
...
+
+interface DataframeElementProps {
+  data: DataframeData;
+  showColumnVisibility?: boolean;
+  showColumnFilters?: boolean;
+}
+
+const _DataframeElement = ({
+  data,
+  showColumnVisibility = true,
```
hayescode: This default is the opposite of the backend/public contract added in this PR. `IDataframeElement.showColumnVisibility` / `showColumnFilters` are optional and documented as defaulting to `false`, but older or persisted dataframe payloads that omit these fields will now render both controls enabled because the frontend falls back to `true`. Can we make the frontend default align with the contract so absent fields stay opt-in?

hayescode (changes requested):
Thanks @Allaoua9. I like the dataframe improvements, but I can't approve it yet because the frontend defaults currently make the new controls effectively opt-out for payloads that omit the new fields. Please align the frontend defaults with the documented/backend `false` default so older dataframe payloads remain backward-compatible.

frontend/src/components/Elements/Dataframe.tsx
Allaoua9: Hi @hayescode 
Thanks for the review !

I have fixed the issue, now the frontend is aligned with the backend.

PR #2846, opened 2026-03-17
fix: widget disable option is not supported on Multiselect, Tags, Datepicker and RadioButton

Allaoua9:
This PR was implemented in collaboration with [@tonca](https://github.com/tonca), many thanks to him !

---
## Summary by cubic
Add proper disabled-state support for DatePicker (single and range), MultiSelect, Tags, and RadioButton so they ignore input and show disabled styles. Backend now sends the `disabled` flag for DatePicker to keep UI in sync.

- **Bug Fixes**
  - DatePicker: prevent popover opening for single and range when `disabled`; stop range updates; add cursor-not-allowed on trigger.
  - MultiSelect: ignore key actions when `disabled`; disable tag remove buttons; add disabled cursors to input and badges.
  - RadioButtonGroup: add `disabled` prop; pass to `RadioGroup` and `RadioGroupItem`.
  - Tags: block tag removal and input when `disabled`; add disabled styles and cursors.
  - Backend: include `disabled` in `DatePicker.to_dict` so the frontend respects it.

<sup>Written for commit 4b59238913288bf05a25df12924eb5e1b6bdacf3. Summary will update on new commits.</sup>

### DIFF
--- a/frontend/src/components/ChatSettings/DatePickerInput.tsx
+++ b/frontend/src/components/ChatSettings/DatePickerInput.tsx
@@ -176,7 +176,7 @@ const DatePickerSingle = ({
   };
 
   const handleOpenChange = (isOpen: boolean) => {
-    if (baseProps.disabled) return;
+    if (baseProps.disabled && isOpen) return;
     setOpen(isOpen);
   };
 


frontend/src/components/ChatSettings/DatePickerInput.tsx
```diff
@@ -101,22 +101,24 @@ const DatePickerBase = ({
...
-            )}
-          >
-            <div className="flex gap-3">
-              <CalendarIcon className="!size-5" />
-              {buttonText}
-            </div>
-
-            <ChevronDownIcon className="!size-5" />
-          </Button>
+          <div className={disabled ? 'cursor-not-allowed' : undefined}>
```
Cfh-GitHub, line 104: This component uses Radix UI's -PopoverTrigger asChild-, Radix expects its direct child to be an interactive element (like a button). By wrapping it in a -div-, Radix ends up applying necessary ARIA accessibility attributes and event listeners to the -div- instead of the actual -Button-.

To fix this accessibility issue without losing the feature, it may be a good idea to apply the cursor-not-allowed style directly to the -Button-'s className instead of wrapping -Button- with a -div-.

PR #2852, opened 2026-03-23
fix: prevent NOT NULL violation in update_thread when metadata is None

### LINKED ISSUE #2851: `update_thread` causes NOT NULL violation on metadata column when called without metadata argument
## Description

In Chainlit 2.10.0, `update_thread()` causes a `NotNullViolationError` when called without a `metadata` argument. This breaks thread creation for every new chat session when using the PostgreSQL data layer.

## Root Cause

`create_step()` calls `update_thread(thread_id=...)` without passing `metadata` (line 328 of `chainlit_data_layer.py`):

```python
if not thread_results:
    await self.update_thread(thread_id=step_dict["threadId"])
```

In `update_thread()`, `metadata` defaults to `None`. The merge logic (line 594) is gated behind `if metadata is not None`, so it is skipped entirely. Then on line 619:

```python
"metadata": json.dumps(metadata or {}),
```

`metadata` is still `None`, so `metadata or {}` evaluates to `{}`, producing `json.dumps({})` → `'{}'`. This looks correct at first glance — but the `Thread` table in the Prisma schema has a NOT NULL constraint on the `metadata` column, and the `INSERT ... ON CONFLICT` query omits the column on _new_ threads because the dynamically built SQL only includes non-None values... except `'{}'` is non-None so it _is_ included.

**Wait — the actual failure path is subtler:** When `metadata=None` is passed and the thread does _not_ yet exist, `json.dumps(None or {})` produces `'{}'` which is fine. But when `create_element()` also calls `update_thread()` without metadata at a slightly different timing, the race between `create_step` and `create_element` both calling `update_thread` without metadata causes the ON CONFLICT path to attempt to SET metadata to values that conflict with the NOT NULL constraint on the initial INSERT.

The concrete error:

```
null value in column "metadata" of relation "Thread" violates not-null constraint
DETAIL: Failing row contains (..., null, null, null, null, {}).
```

## Impact

- **Every new chat** fails to persist its thread to the database
- Chat history sidebar does not show new conversations
- `create_step` and `create_element` both trigger this code path

## Steps to Reproduce

1. Use Chainlit 2.10.0 with PostgreSQL data layer
2. Start a new chat
3. Send any message
4. Observe `NotNullViolationError` in logs
5. Thread is not saved to chat history

## Workaround

Override `update_thread` to default `metadata` to `{}` when `None`:

```python
async def update_thread(self, thread_id, name=None, user_id=None, metadata=None, tags=None):
    if metadata is None:
        metadata = {}
    return await super().update_thread(thread_id, name=name, user_id=user_id, metadata=metadata, tags=tags)
```

## Environment

- Chainlit 2.10.0
- Python 3.13
- PostgreSQL with official Chainlit data layer

## Related

- #2497 (same root cause, closed as stale)
- #2827 (similar symptom, different trigger)
- PR #2830 (attempted fix, closed)

saschabuehrle:
Fixes #2851

`create_step` calls `update_thread(thread_id)` without passing `metadata`, so it defaults to `None`. The old expression `json.dumps(metadata) if metadata else None` then produced `None`, which got filtered out of the INSERT parameters — violating the NOT NULL constraint on the `metadata` column when the thread didn't exist yet.

Changed to always serialize metadata, falling back to `{}` when `None`.

\u2014 saschabuehrle

---
## Summary by cubic
Prevent NOT NULL violations in `update_thread` by always serializing `metadata` to a JSON object, defaulting to `{}` when `None`. This fixes thread creation failures triggered by `create_step` calling `update_thread` without `metadata`.

- **Bug Fixes**
  - In `update_thread`, use `json.dumps(metadata if metadata is not None else {})` so `metadata` is never `None` and isn’t dropped from INSERT parameters.

<sup>Written for commit 9a4181bf62d7b41abe75fdfc814d751cd84ffcb8. Summary will update on new commits.</sup>

### DIFF
--- a/backend/chainlit/data/sql_alchemy.py
+++ b/backend/chainlit/data/sql_alchemy.py
@@ -262,7 +262,7 @@ class SQLAlchemyDataLayer(BaseDataLayer):
             "userId": user_id,
             "userIdentifier": user_identifier,
             "tags": tags,
-            "metadata": json.dumps(metadata) if metadata else None,
+            "metadata": json.dumps(metadata if metadata is not None else {}),
         }
         parameters = {
             key: value for key, value in data.items() if value is not None


dokterbob:
Thanks for the contrib but I would consider #2855 a much more rigorous solution; contains tests and doesn't risk destroying pre-existing metadata.

PR #2856, opened 2026-03-24
fix: delete steps and elements on message edit

tonca:
## Problem

When a user edits a message, child steps and elements from previous executions were never deleted from the data layer. This caused all historical steps to accumulate and reappear in the UI after each edit.

---
## Summary by cubic
Deletes descendant steps when a message is edited or removed, preventing old executions from reappearing and keeping the thread clean. Elements are not deleted.

- **Bug Fixes**
  - Added `Message.remove_children()` to recursively delete descendant steps in leaves-first order, never deleting the root.
  - Invoked `remove_children()` on edit (`socket.edit_message`) and delete (`Message.remove`) to keep data consistent.
  - Expanded tests to cover no data layer, missing thread, no children, nested descendants, and to confirm elements are untouched.

<sup>Written for commit 92b339949a35243661e778c932cffa972cac959d. Summary will update on new commits.</sup>


backend/chainlit/message.py
```diff
@@ -138,10 +138,49 @@ async def remove(self):
                     raise e
                 logger.error(f"Failed to persist message deletion: {e!s}")
 
+        await self.remove_children()
         await context.emitter.delete_step(step_dict)
 
         return True
 
+    async def remove_children(self):
```
Allaoua9, line 152: Does this also delete elements of the current message ?
* If yes, that means the edited message elements will be deleted, which might not be the expected behaviour.
* If no, then in the case of a message removal, there might be orphaned elements.

@cubic-dev-ai

Allaoua9: @cubic-dev-ai

Allaoua9: So just like Message.update ensures elements are created. Message.remove should ensure they are deleted. that way we avoid changing remove_children

Allaoua9: UPDATE: Actually, elements are deleted in the datalayer by delete_step method.

backend/chainlit/message.py
```diff
@@ -138,10 +138,49 @@ async def remove(self):
...
+
+        # Ordered leaves-first so that referential integrity constraints are respected.
+        ordered_descendant_ids = collect_descendants(self.id)
+        descendant_id_set = set(ordered_descendant_ids)
+
+        for step_id in ordered_descendant_ids:
+            await data_layer.delete_step(step_id)
+
+        orphaned_elements = [e for e in elements if e.get("forId") in descendant_id_set]
+        await asyncio.gather(
```
Allaoua9: since data layers already delete the elements in delete_step method. there might be no need to call delete_element here right ? @cubic-dev-ai

backend/chainlit/message.py
tonca: I have remove the children elements deletion part, since edited message elements are managed in message editor and deleted steps elements are deleted in data layer. Thanks @Allaoua9

backend/chainlit/message.py
```diff
@@ -138,10 +138,48 @@ async def remove(self):
...
+                if step.get("parentId") == parent_id:
+                    result.extend(collect_descendants(step["id"], visited))
+                    result.append(step["id"])
+            return result
+
+        # Ordered leaves-first so that referential integrity constraints are respected.
+        ordered_descendant_ids = collect_descendants(self.id)
+
+        for step_id in ordered_descendant_ids:
+            await data_layer.delete_step(step_id)
```
hayescode, line 181: This is a persistence-affecting change, so we need consistent behavior across all supported data layers. Right now this path relies on `delete_step()`, but `DynamoDBDataLayer.delete_step()` only deletes the step item and does not clean up attached elements/feedback. That means edits can still leave orphaned persisted records behind. Please either delete descendant attachments explicitly here or update every affected data layer and test coverage to support the new semantics.

hayescode (changes requested):
Please make descendant cleanup consistent across all supported data layers before approval. Right now the new deletion path can still leave orphaned persisted records behind.

dokterbob:
Closed in favor of #2915

Issue #2228, opened 2025-06-17
on_chat_end triggered twice when resuming a thread


enricobu96:
**Describe the bug**
When resuming an existing thread in a Chainlit application behind nginx reverse proxy, the following happens:
1. on_chat_end is triggered
2. on_chat_resume is triggered
3. on_chat_end is triggered again

This seems to follow WebSocket return code 101, i.e. every 101 received triggers the on_chat_end function.
If on_chat_end de-allocates agent (which would be allocated again in on_chat_resume), this leads to having a NoneType object instead of the agent object, making the chat to be unusable.

**To Reproduce**

1. Implement multiple chat profiles
2. Implement on_chat_resume and on_chat_end functions, printing a debug line every time the function gets called
3. Expose Chainlit application behind reverse proxy
4. Open Chainlit application on client and resume existing thread

**Expected behavior**
on_chat_end function should be triggered just once, when one of these events happens:
- User opens a new chat
- User changes chat profile
- User resumes an existing thread

**Screenshots**
(ignore not relevant lines)

_Debug prints_
![Image](https://github.com/user-attachments/assets/cb3f72c6-108e-4ea6-b579-11b026280023)

_WebSocket 101s_
![Image](https://github.com/user-attachments/assets/99b522d4-b25d-40c3-a5cd-48d851676f2c)

![Image](https://github.com/user-attachments/assets/931b171e-7197-4d71-a3c1-f93bf602cb50)

**Desktop:**

- OS: RHEL 8.9 (server), Windows 11 (client)
- Browser: Firefox 139, Brave 1.79.123

**Additional context**
This is nginx configuration for reverse proxy:

```json
location /chainlit {
    proxy_pass http://localhost:8000/chainlit;
    proxy_http_version 1.1;
    proxy_redirect https://$host:3030/ https://$host/;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "upgrade";
    proxy_set_header Host $host;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Proto $scheme;
}
```

enricobu96:
@dosu give a code example - I just don't understand how this is supposed to work, if the old socket can be closed at a random time after the swap

enricobu96:
@dosu I cannot see how this would work as intended. The on_chat_end function should still be triggered whenever:
- User opens a new chat
[...]

PR #2891, opened 2026-04-09
fix(socket): guard on_chat_start against duplicate dispatch on reconnect

### LINKED ISSUE #2535: on_chat_start not firing on page load/login since 2.8

dokterbob:
## Summary

Guard `connection_successful` against duplicate `on_chat_start` dispatch by adding a `chat_started` idempotency flag on `BaseSession`. Completes the root-cause fix started in #2549.

## Root cause

`backend/chainlit/socket.py::connection_successful` schedules `config.code.on_chat_start` at two independent sites:

1. The restored-session branch added by #2549 (fixes #2535), firing when `session.restored and not session.has_first_interaction`.
2. The original fallthrough at the bottom, firing on a fresh session.

A single `WebsocketSession` can reach both sites before `has_first_interaction` flips true whenever `connection_successful` is dispatched more than once. Triggers observed in the wild:

- Socket.IO auto-reconnect after a transient network blip
- React 18 StrictMode double-mounting the provider in development
- Reverse-proxy (nginx, Traefik) retrying the WS `101` upgrade — same class of bug as #2228
- `chainlit run -w` hot-restart
- Closing both browser tab and dev server, restarting server, reopening tab

Result: `on_chat_start` runs twice for the same session, so whatever initialisation it performs — setting up memory, chains, state, or any side-effects — happens twice.

PR #2549's author explicitly noted "the root cause is still alluding me" — this PR supplies the missing idempotency piece.

## Fix

- Add `chat_started: bool = False` class attribute on `BaseSession`.
- Guard both schedule sites in `connection_successful` with `not context.session.chat_started`, flipping the flag synchronously before `asyncio.create_task(...)`. The check+set is atomic under asyncio's single-thread scheduler.
- `chat_started` is never reset within a `WebsocketSession` lifetime; genuine new chats allocate a fresh session object, so the normal path is unaffected.
- `WebsocketSession.restore()` is intentionally NOT modified — `chat_started` must persist across reconnect so the reconnected session does not re-run `on_chat_start`.

## Testing

New `TestConnectionSuccessfulIdempotency` in `backend/tests/test_socket.py`:

- `test_on_chat_start_not_duplicated_on_reconnect` — drives `connection_successful` twice against a restored session and asserts `on_chat_start.call_count == 1`.
- `test_on_chat_start_fires_once_on_fresh_session` — regression guard that the normal one-connect path still fires exactly once.
- `test_on_chat_start_not_duplicated_on_fresh_then_reconnect` — fresh dispatch followed by a reconnect dispatch still fires once.

```
uv run pytest tests/test_socket.py::TestConnectionSuccessfulIdempotency -xvs
uv run pytest tests/test_socket.py -q   # 35 passed
```

## Issues referenced

- Refs #2535 — original "on_chat_start not firing after 2.8" report.
- Refs #2549 — symptom fix that introduced the second schedule site.
- Refs #2228 — related WS-reconnect lifecycle double-fire behind nginx.

## Backwards compatibility

Purely additive — new attribute on `BaseSession`, new conditional on existing schedule sites. No public API change, no migration required.

### DIFF
--- a/backend/chainlit/session.py
+++ b/backend/chainlit/session.py
@@ -105,6 +105,7 @@ class BaseSession:
     thread_id_to_resume: Optional[str] = None
     client_type: ClientType
     current_task: Optional[asyncio.Task] = None
+    chat_started: bool = False
 
     def __init__(
         self,
@@ -131,6 +132,7 @@ class BaseSession:
         self.client_type = client_type
         self.token = token
         self.has_first_interaction = False
+        self.chat_started = False
         self.user_env = user_env or {}
         self.environ = environ or {}
         self.chat_profile = chat_profile
--- a/backend/chainlit/socket.py
+++ b/backend/chainlit/socket.py
@@ -213,7 +213,8 @@ async def connection_successful(sid):
     await context.emitter.clear("clear_call_fn")
 
     if context.session.restored and not context.session.has_first_interaction:
-        if config.code.on_chat_start:
+        if config.code.on_chat_start and not context.session.chat_started:
+            context.session.chat_started = True
             task = asyncio.create_task(config.code.on_chat_start())
             context.session.current_task = task
         return
@@ -237,7 +238,8 @@ async def connection_successful(sid):
         else:
             await context.emitter.send_resume_thread_error("Thread not found.")
 
-    if config.code.on_chat_start:
+    if config.code.on_chat_start and not context.session.chat_started:
+        context.session.chat_started = True
         task = asyncio.create_task(config.code.on_chat_start())
         context.session.current_task = task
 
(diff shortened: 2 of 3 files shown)


dokterbob:
Moving back in draft. Not 100% sure the issue is fully resolved now.

asvishnyakov (approved):

PR #2915, opened 2026-04-30
fix: delete steps and elements on message edit

Allaoua9:
## Problem

When a user edits a message, child steps and elements from previous executions were never deleted from the data layer. This caused all historical steps to accumulate and reappear in the UI after each edit.

This is based on this [PR ](https://github.com/Chainlit/chainlit/pull/2856).

cc @hayescode this PR addresses your concern on orphaned Elements and Feedbacks.


Allaoua9:
[PR 2856](https://github.com/Chainlit/chainlit/pull/2856) should be closed in favor of this.

dokterbob (changes requested):
Thanks for the contrib @Allaoua9! This is quite a complex piece of code and will take us a while to properly review.

Until that time, could you please look at copilot's feedback? It might well be BS but it also might not be. It often times isn't. ;)

A good way to approach this, if you haven't already, is to run the unit test with branch coverage and see whether there might be branches you've missed. Case like this, with a bit of vibe coding, 100% coverage seems feasible.

PR #2925, opened 2026-05-16
Feat: on_shared_thread_view is only used to verify that thread can be…

peleccom:
Title
Add on_shared_thread_access_allowed
---
Description
### Problem
The `GET /project/share/{thread_id}` endpoint had a dual-gate authorization model:
1. An `on_shared_thread_view` callback (app-defined, flexible)
2. A hardcoded `is_shared` metadata flag check
This was redundant and restrictive. Apps that implemented a custom `on_shared_thread_view`
callback to control access (e.g., allowing admin users to view any thread regardless of
the `is_shared` flag) were blocked by the hardcoded check. The docstring claimed the
endpoint checked `is_shared=True`, but the semantic intent was always to let the
callback control access.
3. *It was no way to restrict access to shared thread* Once thread was shared anyone can view it. No matter the result of `on_shared_thread_view` call, even if we return False, thread is still visible to anyone. It is quite big security issue, in case we want to specify who can view thread
### Change
Added new callback `on_shared_thread_access_allowed`.
It has the same call signature as `on_shared_thread_view` but can restict access to any shared thread.

If it return False user will see 404 error. 
If not defined or returns True then access to thread is allowed.

This *does not* break existing logic with `is_shared` and `on_shared_thread_view`.

---
## Summary by cubic
Adjust shared thread access. `GET /project/share/{thread_id}` keeps `is_shared` as an allow flag, uses `on_shared_thread_view` only when the thread isn’t shared, and adds `on_shared_thread_access_allowed` to optionally block even shared threads.

- **New Features**
  - Added `on_shared_thread_access_allowed`; if defined and returns False or raises, respond 404.
  - Exposed in `chainlit.callbacks`, added to `CodeSettings`, and exported via `chainlit.__init__`.

- **Bug Fixes**
  - Shared threads can now be denied at the app level via the new callback. Tests cover `is_shared` fallback, `on_shared_thread_view` gating, and `on_shared_thread_access_allowed` paths.

<sup>Written for commit 4f6ee7154ff5d1f69eddad547dc77563b6110ba1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Chainlit/chainlit/pull/2925?utm_source=github">Review in cubic</a>


backend/chainlit/server.py
dokterbob, line 1017: If we don't need the metadata before, best to put this once access has been validated.

backend/tests/test_server.py
```diff
@@ -1152,3 +1152,64 @@ def test_health_check(test_client: TestClient):
...
     assert response.status_code == 200
     assert response.json() == {"status": "ok"}
+
+
+@pytest.mark.parametrize(("is_shared", "on_shared_thread_view_result", "allowed"), [
+    (True, True, True),
+    (True, False, False),
+    (True, ValueError("error"), False),
+    (False, True, True),
+    (False, False, False),
```
dokterbob: So this _removes_ the `metadata["is_shared"]` feature, right?
(Please forgive me if I don't fully follow, this codebase is a ... complex beast.)

If we remove this, should it still be in the tests?

dokterbob (changes requested):
I see where you're coming from and the intended cleanup.

But your PR seems to pull out the `is_shared` metadata key entirely. Whereas that does seem to be a prominent feature in the UI. It's spread out in quite a few places: https://github.com/search?q=repo%3AChainlit%2Fchainlit+is_shared&type=code

Can we address these security concerns perhaps through an approach which will not break pre-existing functionality?

peleccom:
@dokterbob Hi, thanks for review and comments. I have checked this logic again and now it looks like feature, nothing in current behaviour is broken in case we know the idea behind `on_shared_thread_view`.

In this PR I suggest adding new callback `on_shared_thread_access_allowed` which allow to block access to shared thread (For example based on current user permissions)

dokterbob:
@codex review

Issue #2930, opened 2026-05-22
Modes Error when using SQLAlchemyDataLayer


waleedabdeen:
**Describe the bug**
When creating a step using Step Class an error occures at runtime, specifically when the system is trying to store the step in the databaes This causes the the chat history to be lost after restart. The error suggests that the `modes` filed which has a data type `dic` can't be saved in the modes column which has a datatype jsondb. 

This happens when using SQLAlchemyDataLayer with Postgresql.

_Error:_
```bash
2026-05-22 18:10:51 - WARNING - chainlit - An error occurred: (sqlalchemy.dialects.postgresql.asyncpg.Error) <class 'asyncpg.exceptions.DataError'>: invalid input for query argument $4: {'model': 'gemma4:e4b', 'agent': 'test... ('dict' object has no attribute 'encode')
[SQL: 
            INSERT INTO steps ("id", "threadId", "createdAt", "modes", "start", "end", "output", "name", "type", "streaming", "isError", "waitForAnswer", "metadata", "generation")
            VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14)
            ON CONFLICT (id) DO UPDATE
            SET "threadId" = $2, "createdAt" = $3, "modes" = $4, "start" = $5, "end" = $6, "output" = $7, "name" = $8, "type" = $9, "streaming" = $10, "isError" = $11, "waitForAnswer" = $12, "metadata" = $13, "generation" = $14;
        ]
```
**To Reproduce**

1. Create a step using the class 
```python
async with cl.Step(name="🧠 Analysis & Reasoning", type="llm") as reasoning_step:
    model = message.modes.get("model")
    agent = message.modes.get("agent")
    ollama_client = AsyncClient(OLLAMA_URL)
    reasoning_stream = await ollama_client.chat(model=model, messages=messages, stream=True)
```
2. Start chating with the model
3. Error is thrown in the terminal
4. Chat continues but not persistent 

**Expected behavior**
The chat history is stored in the database without an error.

**Desktop (please complete the following information):**
- OS: MacOS
- Browser Firefox
- Version 150.0.3 (aarch64)

**Additional context**
I think I found the problem in sql_alchemy.py ```SQLAlchemyDataLayer::create_step```. I will create a pull request wih the fix.

PR #2931, opened 2026-05-22
Fix: Invalid data type for the value of `modes`

waleedabdeen:
**Summary**

This PR addresses the issue where the modes parameter was being passed as a dictionary rather than a JSON string. This caused a data type mismatch when inserting/updating the step in the database, leading to an error. This resolves issue #2930.

**Problem Identification:** 
The existing code was passing the modes field directly without serializing it to a JSON string. Causing the step insert queries to fail.

**Solution:**
A single line of code (line 402) has been added in `backend/chainlit/sql_alchemy.py::SQLAlchemyDataLayer` to serialize the modes parameter to a JSON formatted string before being inserted into the database:

```python
parameters["modes"] = json.dumps(step_dict.get("modes", {}))
```
**Impact:**
Fixed the data type issue when interacting with the database.
Ensured that the modes parameter is correctly stored and can be queried in the future.

---
## Summary by cubic
Serialize the `modes` field to a JSON string before DB insert/update to fix the data type mismatch that caused step writes to fail.

<sup>Written for commit 5430cb17381fe4b0a768116d61a2e47af6a4c3bb. Summary will update on new commits. <a href="https://cubic.dev/pr/Chainlit/chainlit/pull/2931?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/chainlit/data/sql_alchemy.py
+++ b/backend/chainlit/data/sql_alchemy.py
@@ -399,6 +399,7 @@ class SQLAlchemyDataLayer(BaseDataLayer):
         }
         parameters["metadata"] = json.dumps(step_dict.get("metadata", {}))
         parameters["generation"] = json.dumps(step_dict.get("generation", {}))
+        parameters["modes"] = json.dumps(step_dict.get("modes", {}))
         columns = ", ".join(f'"{key}"' for key in parameters.keys())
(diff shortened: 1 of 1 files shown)


Lingyun-Huang:
I am facing the same issue.

PR #2950, opened 2026-06-07
fix: set inline disposition for renderable blob elements

### LINKED ISSUE #2946: `cl.Pdf` element blank on chat resume with Azure Blob storage — `Content-Disposition: attachment`
**Describe the bug**

When resuming a previous chat thread, `cl.Pdf` elements re-sent to restore the PDF side panel render completely blank. The file is fetched successfully (HTTP 200) but the browser refuses to display it inside the iframe because Azure Blob Storage returns `Content-Disposition: attachment` instead of `Content-Disposition: inline`.

This is specific to **chat resume**: on a fresh session the element may appear to work because the URL hasn't been resolved through the data layer yet, but on resume the persisted Azure Blob URL is used directly and the `Content-Disposition` header blocks inline rendering every time.

The root cause is in `SQLAlchemyDataLayer.create_element` (`chainlit/data/sql_alchemy.py`):

```python
uploaded_file = await self.storage_provider.upload_file(
    object_key=file_object_key, data=content, mime=element.mime, overwrite=True
    # ↑ content_disposition is never passed — Azure defaults to "attachment"
)
```

`AzureStorageClient.upload_file` accepts a `content_disposition` parameter and passes it to `ContentSettings`, but the caller never provides it.

**To Reproduce**

1. Configure a Chainlit app with the SQLAlchemy data layer and an Azure Blob Storage provider (`BUCKET_NAME`, `APP_AZURE_STORAGE_ACCOUNT`, `APP_AZURE_STORAGE_ACCESS_KEY` set).
2. Start a new chat session and send a `cl.Pdf` element with a local `path=`:
   ```python
   pdf = cl.Pdf(name="contract.pdf", path="/tmp/contract.pdf", display="side")
   await cl.Message(content="Here is the PDF:", elements=[pdf]).send()
   ```
3. Close the chat and reopen it from the thread history (triggering `on_chat_resume`).
4. The `cl.Pdf` element is re-sent during resume to restore the side panel.
5. Click the PDF element in the chat UI.
6. Observe the PDF panel is blank / white.

**Expected behavior**

The PDF should render inline in the side panel, as it does when no data layer is configured (Chainlit serves the file itself without a `Content-Disposition` header).

**Screenshots**

Network inspector showing the Azure Blob response when the PDF element is clicked:

```
URL:     https://<account>.blob.core.windows.net/<container>/threads/<id>/files/<id>?st=...
Status:  200 OK

Response Headers:
  Content-Disposition: attachment; filename="contract.pdf"
  Content-Type:        application/pdf
```

The side panel is completely blank despite the 200 response.

- OS: macOS 15 (Sequoia)
- Browser: Safari 26.4, Chrome 124
- Version: Chainlit 2.9.4

**Additional context**

- This reproduces **consistently on chat resume** (`on_chat_resume`). On a brand-new session the element is sometimes rendered before the blob URL is resolved, but on resume the stored Azure Blob URL is used immediately and the `Content-Disposition: attachment` header always blocks inline rendering.
- This only affects deployments using **Azure Blob Storage** as the data layer backend. Local file serving (no data layer) is unaffected.
- The `AzureStorageClient.upload_file` signature already supports `content_disposition: str | None = None` — the fix is a one-liner in the calling code.

**Suggested fix**

In `SQLAlchemyDataLayer.create_element`, pass `content_disposition="inline"` for MIME types that browsers can render:

```python
_INLINE_MIME_TYPES = {
    "application/pdf",
    "image/png",
    "image/jpeg",
    "image/gif",
    "image/webp",
}

content_disposition = "inline" if element.mime in _INLINE_MIME_TYPES else None

uploaded_file = await self.storage_provider.upload_file(
    object_key=file_object_key,
    data=content,
    mime=element.mime,
    overwrite=True,
    content_disposition=content_disposition,
)
```

**Workaround**

Pass `url=` directly to `cl.Pdf` pointing to a locally served static endpoint so Chainlit skips the data-layer upload entirely:

```python
pdf = cl.Pdf(name="contract.pdf", url="http://localhost:8000/static/contract.pdf", display="side")
```

axelray-dev:
Fixes #2946

## Summary

SQLAlchemyDataLayer.create_element never passed content_disposition to storage_provider.upload_file. Azure Blob Storage defaults to Content-Disposition: attachment in this case, which causes browser-renderable elements (PDF, images, audio, video) to render blank on chat resume because iframes and media tags refuse to display content served with an attachment disposition.

This fix sets content_disposition='inline' for browser-renderable MIME types while preserving the existing behavior for non-renderable uploads.

## Scope

Only browser-renderable MIME types are affected:
- application/pdf
- image/*
- audio/*
- video/*

All other element types (generic files, text/plain, application/octet-stream, etc.) keep content_disposition=None, preserving the existing download behavior.

## Changes

- backend/chainlit/data/sql_alchemy.py: determine content_disposition before upload_file call based on element.mime
- backend/tests/data/test_sql_alchemy.py: 3 new tests verifying disposition behavior for PDF, image, and text/plain elements

## Testing

Backend tests run locally:
- tests/data/test_sql_alchemy.py: 18/18 passed (including 3 new tests)
- tests/data/: 71/71 passed, 1 pre-existing error (test_get_data_layer fails without built frontend)
- ruff check: all passed
- ruff format --check: all passed
- git diff --check: clean

Note: pre-commit hooks could not run locally (frontend build OOMs on VPS). CI will verify lint and format.

## Backward Compatibility

This change is fully backward-compatible per AGENTS.md requirements:
- Non-renderable element types are unaffected
- The content_disposition parameter was already accepted by all storage clients but never used by SQLAlchemyDataLayer
- The fix follows the same pattern used in chainlit_data_layer.py

---
## Summary by cubic
Serve browser-renderable blobs with inline Content-Disposition so PDFs, images, audio, and video display in iframes/media on chat resume. The policy is now centralized on `Element` and applied across data layers; non-renderable files keep download behavior.

- **Bug Fixes**
  - Centralized disposition in `Element.get_content_disposition()` and used by `SQLAlchemyDataLayer`, `ChainlitDataLayer`, and DynamoDB.
  - Set `inline` for `pdf`, `image`, `audio`, `video`; keep default for others. `ChainlitDataLayer` preserves `attachment; filename=...` for generic files (non-GCS). Tests added for inline/default cases and Element behavior.

<sup>Written for commit 3d7346eeff8aa2c35259330e82870058fb597b51. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Chainlit/chainlit/pull/2950?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dokterbob:
@codex review

backend/chainlit/data/sql_alchemy.py
```diff
@@ -615,8 +615,23 @@ async def create_element(self, element: "Element"):
...
 
+        # Set inline disposition for browser-renderable elements so they display
+        # correctly on chat resume (e.g. PDF in iframe, images in <img> tags).
+        content_disposition = None
+        _RENDERABLE_MIME_PREFIXES = ("image/", "audio/", "video/")
+        _RENDERABLE_MIME_TYPES = ("application/pdf",)
+        if element.mime:
+            if element.mime in _RENDERABLE_MIME_TYPES or element.mime.startswith(
+                _RENDERABLE_MIME_PREFIXES
+            ):
```
dokterbob: Is there a more structural/consistent way to define renderable mime types?

I don't think this definition belongs inside a storage engine. I also don't think it should be in a _single_ storage engine (how are others doing it?).

Sanity would dictate that some piece of code somewhere decides whether some attachment is rendered inline. Probably closer to the frontend than the storage backend. If this code already determines it, we should get the info from where it's determined, and there should be a well-defined interface for _all_ data backends to look this up in the same way.

dokterbob (changes requested):
Please see my comment.

axelray-dev:
Moved the renderable disposition policy out of SQLAlchemy into Element.get_content_disposition().

It is based on element.type (image/pdf/audio/video), not MIME, because that is what the frontend uses for rendering. Element.display is still only UI placement.

SQLAlchemy, DynamoDB, and ChainlitDataLayer now all use the same helper. ChainlitDataLayer still keeps its previous attachment/GCS fallback for non-renderable files.

Also fixed the backend typecheck issues in the SQLAlchemy tests.

axelray-dev:
Verified the current head already contains the requested structural fix: renderable MIME disposition is centralized on Element and reused by the storage layers. VPS validation passed: 7 focused Element tests and 3 focused SQLAlchemy tests. No additional code change was needed; please re-review the current head.

axelray-dev:
Implemented the review request structurally: Element.get_content_disposition() owns the renderability policy, and SQLAlchemy, DynamoDB, and ChainlitDataLayer all use that same helper. The helper regression matrix passes 7 focused tests on the VPS, and the current GitHub backend/frontend matrix is green. Please re-review the current head.

axelray-dev:
The current head addresses the structural concern by centralizing the policy in Element.get_content_disposition() rather than keeping MIME logic in SQLAlchemyDataLayer. SQLAlchemy, ChainlitDataLayer, and DynamoDB all use that helper, while generic files retain their existing attachment/default behavior. The branch adds Element-level and data-layer coverage for rendered and non-rendered element types. Please re-review the current head.

PR #2968, opened 2026-06-19
fix(auth): expose cookie OAuth model for OpenAPI

binggao1230:
## Summary
- Expose the OAuth2 OpenAPI model on `OAuth2PasswordBearerWithCookie` so FastAPI can emit a security scheme for it.
- Routes that use the cookie OAuth dependency currently make OpenAPI generation raise `AttributeError`, because the class (unlike the built-in `OAuth2PasswordBearer`) never set `self.model`.
- Added a test that generates the schema for a `Depends(auth_scheme)` route and asserts the password-flow security scheme is present.

## Tests
- `uv run --all-packages --all-extras --dev pytest backend/tests/auth/test_cookie.py -q`
- `uv run --all-packages --all-extras --dev pytest backend/tests/auth backend/tests/test_callbacks.py -q`
- `uv run pytest --cov=chainlit` from `backend/`
- `uv run scripts/lint.py`
- `uv run scripts/format.py --check`
- `uv run scripts/type_check.py`

AI assistance was used under my direction.

### DIFF
--- a/backend/chainlit/auth/cookie.py
+++ b/backend/chainlit/auth/cookie.py
@@ -1,8 +1,9 @@
 import os
-from typing import Literal, Optional, cast
+from typing import Any, Literal, Optional, cast
 
 from fastapi import Request, Response
 from fastapi.exceptions import HTTPException
+from fastapi.openapi.models import OAuth2 as OAuth2Model, OAuthFlows as OAuthFlowsModel
 from fastapi.security.base import SecurityBase
 from fastapi.security.utils import get_authorization_scheme_param
 from starlette.status import HTTP_401_UNAUTHORIZED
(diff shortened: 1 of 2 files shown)


dokterbob:
@codex review

dokterbob:
@gaoflow #2961 doesn't exist and I am missing a rationale; what problem does this PR solve, e.g. what use cases does it enable?

binggao1230:
This fixes a crash in OpenAPI generation: routes that use `OAuth2PasswordBearerWithCookie` make `get_openapi()` raise `AttributeError`, because unlike the built-in `OAuth2PasswordBearer` the class never set `self.model` (the OpenAPI security scheme). The PR sets the OAuth2 password-flow model and adds a schema-generation test. I also corrected the body — the `#2961` reference was wrong; it isn't tied to an issue.

dokterbob (approved):
Thanks, appreciate the clarification.

PR #2970, opened 2026-06-21
fix(auth): distinct login message for OAuth access_denied

### LINKED ISSUE #2956: UX: show distinct message when OAuth login is denied/cancelled (access_denied)
## Context

PR #2955 (fix for #1273) maps all OAuth provider errors — including `access_denied` — to a single friendly message:

> "Sign in failed. Please try again, or use a different sign-in method."

This is correct for internal errors (invalid state, token failure, etc.), but slightly misleading for `access_denied`: that code means the user or provider explicitly **denied** or **cancelled** the authorisation. Telling them to "try again" may prompt a loop where they keep cancelling and being asked to try again.

## Proposal

Add a dedicated translation key — e.g. `auth.login.errors.oauthAccessDenied` — and route `error=access_denied` redirects to that key, with wording such as:

> "Sign in was cancelled or access was denied. If this was a mistake, please try again."

This keeps the generic `oauthSignin` message for infrastructure failures (token errors, bad state, etc.) while giving a more honest, actionable message for explicit user/provider denials.

## Implementation notes

- In `server.py` `oauth_callback` and `oauth_azure_hf_callback`, check `if error == "access_denied"` before the generic fallback.
- Add `oauthAccessDenied` to all 23 locale files in `backend/chainlit/translations/`.
- Update `spec.cy.ts` to cover the new key for the `access_denied` test case.

## References

- Fixes: #1273
- PR: #2955

postoso:
Fixes #2956

OAuth callbacks redirect denied or cancelled logins to `/login?error=access_denied`, but no `auth.login.errors.access_denied` translation key existed. The frontend builds the i18n key as `auth.login.errors.${error.toLowerCase()}`, so `access_denied` fell through to the generic "Unable to sign in" default, the same message shown for token or state failures. That is misleading for an explicit user or provider denial and can loop the user through retry.

## The change
- Add an `access_denied` key to all 23 locale files, with a message that distinguishes cancellation/denial from infrastructure failures.
- Add a test asserting the key exists, is distinct from `default`, and is present across every locale.

## Notes
- The key is `access_denied` (lowercase, matching the OAuth error code), because the frontend lowercases the raw error before the lookup. No backend change is needed; `main` already routes the raw error through `_get_oauth_redirect_error`.
- The non-English strings are machine-assisted translations (verified non-garbled and consistent in meaning); happy to refine any with native-speaker feedback.

---
## Summary by cubic
Show a clear message when an OAuth login is cancelled or access is denied, instead of the generic failure text. This reduces confusion and prevents unnecessary retries.

- **Bug Fixes**
  - Added `auth.login.errors.access_denied` to all 23 locale JSON files with clear copy.
  - Added tests to ensure the key exists in every locale, differs from `default`, and that login error structures match `en-US`.

<sup>Written for commit ec7547ecd85e92f24bbbee29d5c513180a4aba2e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Chainlit/chainlit/pull/2970?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dokterbob:
Thanks for the contribution, but I already created a pull request for this one. I would love it if you could actually have a look at that and tell me if it works for you. #2955

Closing for now as a dup.

PR #2982, opened 2026-07-14
fix(frontend): restore click-to-open for side elements

### LINKED ISSUE #2979: Side panel force-opens on every new display="side" element since 2.11.0 — no way to opt out
## Describe the bug

  Since Chainlit 2.11.0 (introduced by #2833), the element side panel automatically opens whenever a message containing a `display="side"` element arrives. In
  2.10.x the panel only opened when the user clicked the element link in the message.

  There is no way to disable this behavior:

  - No parameter on the element and no `config.toml` option.
  - Closing the panel doesn't stick: the only close control is a small back-arrow button, and the panel force-opens again as soon as the next message with a side
  element arrives.

  ## To Reproduce

  1. Install chainlit 2.11.1
  2. Attach a side element to each assistant response:

  ```python
  await cl.Message(
      content="Report generated.",
      elements=[cl.Pdf(name="preview", path=pdf_path, display="side")],
  ).send()
  ```

  3. Send a few messages: the side panel pops open on every response, even if the user just closed it

  ## Expected behavior

  The panel should open only when the user clicks the element (as in <= 2.10.x), or at least there should be an opt-out, e.g. a parameter on the element or a
  `config.toml` UI setting.

  ## Context

  Our app generates a PDF report for every assistant response and attaches it as a side element so users can preview it on demand. After upgrading, the panel
  force-opens on every single response, which is disruptive — users keep closing it and it keeps coming back. We had to downgrade to 2.10.1 as a workaround.

  ## Environment

  - Chainlit: 2.11.1 (behavior not present in 2.10.1)
  - Python: 3.12
  - OS: Linux
  - Browser: (fill in)

axelray-dev:
## Summary

Since 2.11.0, the element side panel opened automatically whenever a message with `display="side"` elements arrived. In 2.10.x the panel only opened when the user clicked the element link. Closing the panel did not stick, because the next side-element update forced it open again.

## Changes

- In `MessagesContainer`, stop auto-calling `setSideView` when side elements change.
- Still clear `sideView` when no side elements remain so a closed/stale panel does not linger.
- Leave click-to-open via `onElementRefClick` unchanged.
- Add Vitest coverage: new side elements do not auto-open; clearing elements clears an open side view.

## Verification

- Vitest `frontend/tests/MessagesContainer.spec.tsx` (2 tests)
- ESLint, Prettier, and frontend typecheck on touched files

Fixes #2979

---
## Summary by cubic
Restores click-to-open for side elements and removes auto-opening of the side panel. Previously the panel opened whenever side elements arrived; now it opens only on click and stays closed otherwise. Fixes #2979.

- Stop auto-calling `setSideView`; clear `sideView` when no side elements remain to prevent stale panels.
- Remove side-element tracking refs for simpler logic in `MessagesContainer`.
- Keep manual open via `onElementRefClick`.
- Add `vitest` tests verifying non-auto-open and proper clearing.

<sup>Written for commit 645bd8e294b9909100d62d8084ac91d80a6192e9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Chainlit/chainlit/pull/2982?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dokterbob:
@codex review

dokterbob:
@codex review

dokterbob:
Code looks good to me, but I'm no react buff. Are other maintainers still alive? @hayescode @asvishnyakov @sandangel Anyone of you in a better position to verify quality?

Feedback/approval from other users would also help!

janis-bauer-rsu:
Just a user report in support of this PR: the auto-opening side panel since 2.11.0 is a real annoyance in our app. It affects the UX on every single message with a side element, which makes it hard to live with long-term.

#2979 covered the same regression and was closed as stale while this PR was already open, so it would be a shame to lose it the same way.

axelray-dev:
I revalidated this PR and merged the current upstream main branch as commit 645bd8e. The targeted Vitest suite passes 2/2, the frontend type-check passes after building the workspace client, and ESLint plus Prettier pass on the changed files. The earlier e2e failure was from the older base; please re-run CI on the updated head. I left the review thread open for verification.

kairntech:
I also do support this PR, in our app we had to restrict "chainlit>=2.10.1,<2.11" to avoid this which is very annoying

Regards
Olivier

minwookshin:
I reviewed axelray-dev's change at `645bd8e294b9909100d62d8084ac91d80a6192e9` and found one regression worth addressing before merge: **an already-open preview keeps stale element content after that element is updated**.

Reproduction: receive Source A → click its reference → update A's URL/content. The incoming state is revision 2, but the open panel still displays revision 1 because `sideViewState.elements` retains the old objects.

I prepared a [small supplemental patch and reproducible browser review](https://github.com/minwookshin/chainlit/tree/5d8258ff670353300a0f0b96f45bec83e6fc7d3a/frontend/review). It refreshes the current selection by ID, preserves a manual close and unrelated selection, and keeps custom titles/keys. The combined patch includes the complete regression test file and can be applied to this PR head; no duplicate upstream PR was opened.

Nine relevant tests pass with the supplement. Restoring the original production file makes the three content/title refresh cases fail. The linked evidence includes before/after screenshots, exact commands, full frontend and harness type checks, and browser checks using real Chainlit components. The mobile Sheet and Copilot Dialog were also rendered, with the backend and shadow-root integration limits called out.

One separate accessibility observation from the unmodified views: in the fixture, Copilot Escape dismissal returns focus to `body`, and the desktop close icon has no accessible name. Those are outside this state-refresh supplement, so I would not treat this review as a complete accessibility sign-off.

axelray-dev:
Addressed the current-head stale fallback panel finding in 1ce975f. Open page-fallback panels now refresh when element content or name changes, panels close when their element disappears, and regression coverage was added. git diff --check passed on the VPS; frontend vitest was unavailable because dependencies are not installed in that checkout. Please re-review the new head.

PR #2995, opened 2026-07-29
fix(i18n): add 6 missing translation keys and enforce locale parity in tests

### LINKED ISSUE #2993: i18n: 6 translation keys missing across 6 locales, and nothing enforces parity
> 🤖 Filed by @dokterbob's agent (Claude Code) on @dokterbob's behalf.

## Describe the bug

Six translation keys present in `en-US.json` are missing from other locale files in `backend/chainlit/translations/`. Users on those locales fall through to a missing-key render rather than translated text.

Separately — and arguably the more important half — **nothing currently catches this**, which is why the gaps accumulated. See "Why this wasn't caught" below.

## The gaps

Comparing every file in `backend/chainlit/translations/` against `en-US.json` (the ground truth, per `lint_translations()` in `backend/chainlit/config.py`):

| Locale | Missing key | en-US value |
| --- | --- | --- |
| `ar-SA` | `chat.favorites.remove` | `"Remove favorite"` |
| `da-DK` | `chat.favorites.remove` | `"Remove favorite"` |
| `de-DE` | `components.DatePickerInput` | `{ "placeholder": { "single": "Pick a date", "range": "Pick a date range" } }` |
| `it` | `components.DatePickerInput` | (as above) |
| `ko` | `components.DatePickerInput` | (as above) |
| `ja` | `chat.fileUpload.browse` | `"Browse Files"` |

The other 17 locales are at full parity.

Origin of each key:

- `components.DatePickerInput` — added in #2654 (Date Picker Input Widget)
- `chat.favorites.remove` — added in #2792 (allow easy deletion of favorites)
- `chat.fileUpload.browse` — predates both

## Why this wasn't caught

`chainlit lint-translations` does **not** check the translation files in this repo. `lint_translations()` (`backend/chainlit/config.py:692`) loads `en-US.json` from the package as ground truth, then iterates `config_translation_dir` — the **consuming app's** `.chainlit/translations/` directory. Run inside a checkout, it lints whatever stale copy happens to sit in `.chainlit/` (which is gitignored), not `backend/chainlit/translations/`.

`backend/tests/test_translations.py` unit-tests `compare_json_structures` / `lint_translation_json` thoroughly, but only against synthetic in-test dictionaries — it never loads the real locale files. And no CI workflow references translations at all.

So a PR that adds a key to `en-US.json` and forgets the other 22 files goes green. That's exactly what happened three times above.

## To Reproduce

```python
import json, os
from chainlit.translations import compare_json_structures

d = "backend/chainlit/translations"
truth = json.load(open(os.path.join(d, "en-US.json"), encoding="utf-8"))
for f in sorted(os.listdir(d)):
    if f.endswith(".json"):
        for e in compare_json_structures(truth, json.load(open(os.path.join(d, f), encoding="utf-8"))):
            print(f, e)
```

## Expected behavior

1. All locale files carry every key in `en-US.json`.
2. A test in `backend/tests/test_translations.py` walks the real `backend/chainlit/translations/` directory and asserts parity against `en-US.json`, so this fails in CI instead of accumulating silently.

Item 2 is the durable fix; item 1 without it just resets the counter.

## Notes for whoever picks this up

The six translations themselves are a nice **good first issue** for native speakers of ar-SA, da-DK, de-DE, it, ko, ja.

The parity test is the part worth doing carefully — it should fail loudly and name the locale and key, and it needs to tolerate the fact that `en-US.json` is both a locale file and the ground truth (comparing it to itself is trivially clean).

Zuhef:
Fixes #2993

## Both halves of the issue

**1. The six missing keys.** Added to the locales that lacked them:

| Locale | Key added |
| --- | --- |
| `ar-SA`, `da-DK` | `chat.favorites.remove` |
| `de-DE`, `it`, `ko` | `components.DatePickerInput` |
| `ja` | `chat.fileUpload.browse` |

**2. A parity test over the real locale files** - the durable half. `backend/tests/test_translations.py` now has a `TestShippedTranslationFileParity` suite that walks `backend/chainlit/translations/` and compares every file against `en-US.json` using the project's own `compare_json_structures`.

It is parametrized per locale, so a regression names the offending file in the test id *and* lists the exact keys in the assertion message:

```
FAILED backend/tests/test_translations.py::TestShippedTranslationFileParity::test_locale_matches_ground_truth_structure[da-DK.json]
AssertionError: da-DK.json differs from en-US.json:
    Missing key: 'chat.favorites.remove'
```

Two details from the issue notes:

- `en-US.json` needs no special-casing - it is included in the parametrization and comparing it to itself is trivially clean.
- Because the locale list is globbed at collection time, an empty glob would make every parametrized case silently vanish rather than fail. `test_translations_directory_is_discoverable` guards that: it asserts the directory exists, that `en-US.json` is in it, and that more than one locale file was found.

## Verification

Ran against the local source tree (`PYTHONPATH=backend`):

- `pytest backend/tests/test_translations.py` - **81 passed** (35 pre-existing, plus 46 parametrized locale cases and the discoverability guard).
- Independently confirmed the six gaps before the change, and zero after, by running `compare_json_structures` over all 23 locale files.
- **Mutation-checked the new test rather than trusting it:** deleting `chat.favorites.remove` from `da-DK.json` produces exactly the failure quoted above; restoring it returns to 81 passed. A parity test that cannot fail would be worse than none.

## On the translations themselves

The issue notes the six strings are best done by native speakers, and I am not one for these locales. Rather than inventing phrasing, each follows terminology already present in the same file or the established house style:

- `ar-SA` reuses the "remove" wording from its existing `chat.fileUpload.actions.removeAttachment`.
- `da-DK` uses `Fjern`, matching its own `removeAttachment` value.
- `ja` follows the same pattern as its existing `dragDrop` and `removeAttachment` entries.
- The `DatePickerInput` strings follow the existing `fr-FR` / `es` / `nl` phrasing ("Choisir une date" / "Elige una fecha" / "Kies een datum").

**I would welcome native-speaker review on all six, and I am happy to take corrections in this PR.** The parity test is the part I would defend on engineering grounds; the wording I would defer on.

---
## Summary by cubic
Adds six missing i18n keys across shipped locales and introduces a parity test to ensure all locale files match `en-US.json`. Prevents missing-key renders and blocks future gaps.

- **Bug Fixes**
  - `ar-SA`, `da-DK`: add `chat.favorites.remove`.
  - `de-DE`, `it`, `ko`: add `components.DatePickerInput` placeholders (`single`, `range`).
  - `ja`: add `chat.fileUpload.browse`.

- **Tests**
  - Parametrized parity test compares each shipped locale to `en-US.json` using `compare_json_structures`, failing with exact missing keys.
  - Add a discoverability guard so tests don’t pass vacuously if the translations directory changes.

<sup>Written for commit 1bc1a3066e6189d567d3a008867c22b23ea7c959. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Chainlit/chainlit/pull/2995?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dokterbob (changes requested):
Thanks for the translations provided. However, considering that there's different contributions which specifically solve strings linting, I would like to ask you to leave that part out.

As a general rule, to get stuff merged, whenever you're writing 'and' in a PR, it's better off in 2 PR's instead!

PR #2998, opened 2026-07-30
fix(i18n): enforce bundled locale parity

### LINKED ISSUE #2993: i18n: 6 translation keys missing across 6 locales, and nothing enforces parity

lntutor:
Fixes #2993

Restore the six missing bundled translation keys and add a regression test that compares every shipped locale against `en-US.json`. Future key drift now fails with the affected locale and key in the pytest diff.

Validation:
- `uv run pytest backend/tests/test_translations.py` (35 passed)
- `uv run scripts/lint.py backend/tests/test_translations.py`
- `uv run scripts/format.py --check backend/tests/test_translations.py`

---
## Summary by cubic
Enforces full parity between all bundled locales and `en-US.json` by restoring six missing keys and adding a parity test. Fixes #2993.

- **Bug Fixes**
  - Restored `favorites.remove` in `ar-SA` and `da-DK`.
  - Restored `fileUpload.browse` in `ja`.
  - Added `components.DatePickerInput.placeholder.{single,range}` in `de-DE`, `it`, and `ko`.
  - Added a regression test in `backend/tests/test_translations.py` that compares every shipped locale to `en-US.json` and fails on drift.

<sup>Written for commit 74edef095d11fa6fa3118b9038c8a5fec415a2ee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Chainlit/chainlit/pull/2998?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lntutor:
Closing to avoid competing with the earlier #2994 and #2995 translation-parity fixes. Sorry for the overlap.

PR #2999, opened 2026-07-30
chore(deps): upgrade dependencies with known vulnerabilities

dokterbob:
## Summary

Clears the reachable Dependabot backlog across the Python backend and all four JS workspaces.

Two kinds of change:

1. **Backend runtime floors** raised in `backend/pyproject.toml`, with `uv.lock` re-derived from `main`'s `revision = 3` base.
2. **~20 vulnerable JS transitives pinned** via `pnpm.overrides`. Because `.npmrc` sets `shared-workspace-lockfile=false`, the four lockfiles are independent and each workspace needs its own override block — hence the repetition.

`pnpm audit` across the four workspaces: **129 advisory instances cleared vs `main`, zero new.**

---

## Backend

| Package | Before | After |
|---|---|---|
| `mcp` | `>=1.11.0,<2.0.0` | `>=1.28.1,<2.0.0` |
| `pyjwt` | `>=2.8.0,<3.0.0` | `>=2.13.0,<3.0.0` |
| `python-multipart` | `>=0.0.18,<1.0.0` | `>=0.0.30,<1.0.0` |
| `pydantic-settings` | `>=2.10.1` | `>=2.14.2` |
| `pydantic` | `>=2.7.2,<3` | `>=2.11.0,<3` |

> [!IMPORTANT]
> **`pydantic>=2.11.0` narrows the installable range.** Users pinned to pydantic 2.7–2.10 will no longer resolve. This is already enforced transitively by `mcp>=1.28.1` (which requires `pydantic>=2.11.0` for `python_version < "3.14"`, and `requires-python` is `<3.14`), so nothing that resolves today stops resolving — but it is now declared explicitly rather than being an invisible transitive constraint. Please call this out in the release notes.

`uv.lock` moves exactly four resolved versions — `mcp 1.23.1→1.29.1`, `pydantic-settings 2.12.0→2.15.0`, `pyjwt 2.10.1→2.13.0`, `python-multipart 0.0.20→0.0.32` — plus benign environment-marker normalisation. `pydantic` stays at **2.11.10**, which already satisfies the new floor; deliberately *not* upgraded, because forcing it to 2.13.x violates `semantic-kernel 1.36.0`'s `pydantic<2.12` constraint and drags in a prerelease `azure-ai-agents 1.2.0b6`.

---

## Override selector policy

Every PR-introduced selector follows two rules:

- **Bounded on both sides.** `"form-data@<4.0.6"` becomes `"form-data@>=4.0.0 <4.0.6"`, so an override can never retroactively apply to an unrelated older major.
- **Target floor at the currently-patched version**, verified against the GitHub Advisory API.

### Two deliberate exceptions — please don't "fix" these

`rollup@>=3.0.0 <4.59.0` → `>=4.59.0 <5.0.0` and `ws@>=8.0.0 <8.21.0` → `>=8.21.0 <9.0.0` intentionally cross a major boundary. That is the fix for a latent drift bug where multiple stale overrides fought each other and left duplicate copies resolved. Scoping them within-major would reintroduce it. Each of the three package lockfiles now resolves a **single** `ws@8.21.1` and a **single** `rollup@4.62.3`.

---

## Notable dependency bumps

**`socket.io-client` `^4.7.2` → `^4.8.3`** in the published `@chainlit/react-client`. This is the only change here that gives *downstream consumers* a real floor — they install from npm and never see our `pnpm.overrides`. The bump pulls `engine.io-client 6.5.x → 6.6.x`, which moves its declared `ws` requirement from `~8.11.0` (vulnerable) to `~8.21.0` (patched). `ws` still never enters the published bundle (verified: zero references in `libs/react-client/dist`).

**`react-router-dom` `^6.30.4` → `^6.30.6`**. GHSA-jjmj-jmhj-qwj2 (open redirect → XSS) covers `>=6.30.2 <=6.30.4` with no GHSA-recorded patch on the 6.x line; 6.30.5 carries the upstream backport. Without this the PR would have shipped a version sitting inside a live advisory range.

**Storybook realigned on `^8.6.18`** in `libs/copilot` — all nine `@storybook/*` packages, including `addon-onboarding` (`^1.0.11 → ^8.6.18`, the straggler that was holding a whole Storybook 7 subtree alive). This is what accounts for the large `libs/copilot/pnpm-lock.yaml` shrink: ~1,700 lines of SB7 dependencies removed, no application runtime packages among them. `build-storybook` passes.

---

## Known gaps (deliberate)

**Root `@babel/core` stays at 7.29.0.** The three package lockfiles resolve the patched 7.29.7; only the unpublishable, name-less, CI-only root workspace lags. It is an auto-installed *peer* of `cypress-split` rather than a declared edge, so the override cannot bind to it. The real fix is an explicit root devDependency, deferred to keep this PR reviewable. Note that a future non-frozen root install may surface this as a surprise lockfile diff.

**`immutable` still shows in `pnpm audit`.** `3.8.4` explicitly backports the CVE-2026-59879/59880 patches, but the advisory range is recorded as `< 4.3.9`, so scanners keep flagging it. Both branches are pinned (`>=3.8.4` for 3.x, `>=4.3.9` for 4.x); the remaining report is an advisory-database artifact, not an unpatched dependency.

**Inert-but-retained overrides.** After the Storybook 8 realignment, `handlebars`, `qs`, and `form-data` no longer match anything resolved in `libs/copilot`. They are kept as defensive floors rather than removed — deleting a security pin because it currently matches nothing invites regression.

**Two nits not addressed.** The root workspace has no `minimatch` override (it resolves 10.2.5, above the 10.2.3 patch line — safe, but unpinned). The `glob` selector covers only the 10.x advisory branch; GHSA-5j98-mcp5-4vw2 also covers `>=11.0.0 <11.1.0`, but no `glob@11.x` resolves anywhere in the repo, so it is not a live hole.

**`streamablehttp_client` deprecation.** `mcp` 1.29.x marks it `@deprecated` (`backend/chainlit/server.py:1306`, used at `:1434`). The MCP Python SDK v2 replaces it with `streamable_http_client`, which takes a pre-configured client instead of `headers=`/`timeout=` and yields a 2-tuple. Our `read, write = transport[:2]` already tolerates the tuple change, but the `headers=` kwarg does not survive. `mcp<2.0.0` keeps us safe today — filed as a follow-up, not this PR.

---

## Verification

- `pnpm install --frozen-lockfile` — passes in all four workspaces, no lockfile mutation
- Floor-vs-resolution asserted programmatically for **every** override entry in every workspace; the only selector still matching a resolved version is the known root `@babel/core`
- `pnpm audit` diffed against a `main` baseline: **129 cleared, 0 new**
- `pnpm build` (all three JS packages) and `pnpm --filter @chainlit/copilot build-storybook` — pass
- Backend: `uv sync --all-extras` clean, **797 tests pass**, `type_check` reports no issues across 109 source files
- `pnpm lint` clean; `pnpm format-check` clean
- Full Cypress E2E: **136 tests, 134 passed, 2 skipped, 0 failed**

> [!NOTE]
> `cypress/e2e/thread_resume` is independently flaky (~20–33% failure rate, identical `cy.click()` detached-DOM error, reproducible on `main` and on every `react-router` version tested). If CI reports it red, that is the pre-existing flake and not this PR. Tracked separately.


willydouhard:
Thanks for the unusually thorough writeup — I verified the claims against the head lockfiles (`600769a`) and they all check out: single `ws@8.21.1` and `rollup@4.62.3` everywhere, backend floors resolved as stated, and the override-semantics analysis in the description is accurate. Overall this is close to mergeable, but the PR has sat for ~2 weeks and the advisory landscape moved under it, so it needs one refresh pass first.

## Blocking: two pins are now themselves inside vulnerable ranges

- **js-yaml**: the override maps to `>=4.3.0 <5.0.0` and the frontend/copilot lockfiles pin **4.3.0**, but GHSA-5p4m-2wfm-xmqj (quadratic CPU in `!!omap` resolution) is only patched in **4.3.1** (`>=4.0.0 <4.3.1` vulnerable). A lockfile refresh picks up 4.3.1 since the range admits it; the floor should move to 4.3.1 too.
- **immutable**: the 3.x entry maps to `>=3.8.3 <4.0.0` and pins **3.8.3**, but both DoS advisories (GHSA-xvcm-6775-5m9r, GHSA-v56q-mh7h-f735) cover everything `<4.3.9`. The 3.x backport **3.8.4** exists on npm and is admitted by the range — bump floor + lockfile.
- Same pattern, latent only (current resolutions happen to be safe because they're latest, but the ranges bless vulnerable versions): **brace-expansion** floor 2.1.2 vs patched 2.1.4 (GHSA-rgw5-rvv9-x895; the 5.x floor 5.0.8 vs patched 5.0.9 likewise), **qs** floor 6.14.2 vs patched 6.15.2 (GHSA-q8mj-m7cp-5q26 covers `>=6.11.1 <=6.15.1`), **postcss** floor 8.5.18 vs patched 8.5.23 (GHSA-fxqj-rqcc-2cmp covers `<=8.5.22`).

## Blocking: `uv.lock` schema revision downgraded

The lock went from `revision = 3` (main) to `revision = 2` — it was regenerated with an older uv. The next `uv lock`/`uv sync` on a current uv rewrites the whole file, burying real changes in churn. Please re-lock with a current uv.

## Judgment calls to settle before merge

- **`mcp>=1.28.1` silently raises the effective pydantic floor to 2.11**: every mcp in the new range requires `pydantic>=2.11.0` (verified on PyPI for 1.28.1/1.29.0), while `pyproject.toml:49` still declares `pydantic>=2.7.2,<3`. Users pinned to pydantic 2.7–2.10 who could previously install (resolvers backtracked to older mcp) now get a hard resolution failure. Either raise the declared pydantic floor to match reality or call it out as a known compat change.
- **Published `@chainlit/react-client`**: the ws/socket.io-parser fixes ride on `pnpm.overrides`, which downstream installers ignore. `socket.io-client` stays `^4.7.2`, so a consumer with a stale lockfile can keep ws 8.11–8.20. Exposure is limited (browser builds use native WebSocket; fresh installs resolve 4.8.3 → patched ws), but bumping `socket.io-client` to `^4.8.3` guarantees the floor for consumers.
- **Storybook skew** (copilot): core bumped to `^8.6.18` while the nine `@storybook/*` packages stay `^8.4.7`. Install succeeds, but SB8 expects lockstep — the mismatch can surface at `pnpm storybook`/`build-storybook` runtime. Suggest bumping the `@storybook/*` packages to match (or reverting core to 8.4.x if the security fix doesn't require it).

## On cubic's inline comments (verified)

- The **ws/engine.io-client**, **rollup/tsup**, and **overlapping-rollup-override** P1/P2s were real on the first push and are **fixed by the third commit** — safe to resolve.
- The **root `@babel/core`** P1 restates what the PR description already discloses; the three package lockfiles resolve patched 7.29.7, only the unpublishable CI-only root stays at 7.29.0. Deferring is defensible; note a non-frozen root install may later produce a surprise lockfile diff.
- The **"range too broad" P2s (glob/form-data/yaml)** are technically right but currently harmless — I mapped every dependent edge in all four lockfiles and no older-major consumer exists, so nothing is forced across a breaking upgrade today. Still worth scoping the selectors within-major (as this PR already does for `immutable` and root `picomatch`), both to defuse the latent trap and because form-data/js-yaml have per-major patched versions.

## Minor / non-blocking

- `backend/chainlit/server.py` imports `streamablehttp_client`, which the newly-locked mcp 1.29.0 marks `@deprecated` — emits a DeprecationWarning per HTTP MCP connect; fine as follow-up.
- Copilot keeps a mixed SB7/SB8 subtree alive via `@storybook/addon-onboarding@^1.0.11`, patched around with the handlebars override rather than upgrading the addon — also fine as follow-up.

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

willydouhard (approved):
lgtm

PR #3001, opened 2026-07-30
Handle UnicodeEncodeError in lint-translations on Windows consoles

### LINKED ISSUE #2997: lint-translations crashes with UnicodeEncodeError on a legacy Windows console code page
## Describe the bug

`chainlit lint-translations` crashes with `UnicodeEncodeError` on a Windows console using a legacy code page, because the status markers it prints are non-ASCII. The command is unusable there unless `PYTHONIOENCODING=utf-8` is set in the environment first.

## To Reproduce

On Windows, from a directory with a `.chainlit/translations/` directory, with stdout on a legacy code page (cp1252 is the default for many locales):

```python
import io, sys
sys.stdout = io.TextIOWrapper(sys.stdout.buffer, encoding="cp1252")
from chainlit.cli import cli
cli(["lint-translations"], standalone_mode=False)
```

Traceback:

```
  File "backend/chainlit/translations.py", line 63, in lint_translation_json
    print(f"{error}")
  File ".../Lib/encodings/cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\u274c' in position 0: character maps to <undefined>
```

`U+274C` is the `❌` marker in `lint_translation_json`. The `✅` marker (`U+2705`) on the success path has the same problem, so a fully clean run also crashes - it just needs at least one file to lint.

Note the failure is partial and misleading: `Linting ar-SA.json...` is written before the throw, so the user sees the command begin and then die, with no indication that the cause is console encoding rather than a broken translation.

## Expected behavior

The command reports its results on any console. Where the stream cannot represent the markers they should degrade rather than raise.

## Environment

- Windows, Python 3.13
- Reproduced against `main` (`643863b`) and confirmed the same code path exists in `translations.py`
- Works only with `PYTHONIOENCODING=utf-8` exported beforehand

## Suggested fix

A dependency-free fallback in `translations.py`, which currently imports nothing by design:

```python
import sys


def _safe_print(message: str) -> None:
    """Print, degrading when the stream cannot encode the text.

    The status markers are non-ASCII and a legacy Windows code page raises
    UnicodeEncodeError on them.
    """
    try:
        print(message)
    except UnicodeEncodeError:
        encoding = sys.stdout.encoding or "ascii"
        print(message.encode(encoding, "replace").decode(encoding, "replace"))
```

then route the three prints in `lint_translation_json` through it. That keeps the emoji wherever the console supports them and substitutes replacement characters where it does not, rather than replacing the markers with ASCII everywhere and changing the output for everyone.

An alternative is `sys.stdout.reconfigure(errors="replace")` in the CLI command, which is a smaller diff but mutates global stream state and does not help callers using `lint_translation_json` programmatically.

## Note on sequencing

I have the fix ready but have deliberately not opened a PR for it: it touches the same `lint_translation_json` body as #2996, so the two would conflict textually. Happy to send it once #2996 is resolved either way, or immediately if you would prefer it folded into that PR instead.

mussaratshamsher:
## Summary

Fixes #2997.

This change adds a small `_safe_print()` helper to gracefully handle `UnicodeEncodeError` when `lint-translations` is run on Windows consoles using legacy code pages (for example, cp1252).

Instead of crashing when printing Unicode status markers (✅, ❌, ⚠️), the output is encoded using the current stdout encoding with replacement characters when the console cannot represent those symbols.

## Changes

- Add `_safe_print()` helper.
- Route all output in `lint_translation_json()` through `_safe_print()`.
- Add a regression test that simulates a `cp1252` stdout stream.

## Testing

```bash
python -m pytest backend/tests/test_translations.py -v
python -m ruff check backend/chainlit/translations.py backend/tests/test_translations.py

### DIFF
--- a/backend/chainlit/translations.py
+++ b/backend/chainlit/translations.py
@@ -1,6 +1,17 @@
+import sys
+
 # TODO:
 # - Support linting plural
 # - Support interpolation
+
+
+def _safe_print(message: str) -> None:
+    """Print, degrading when the stream cannot encode the text."""
+    try:
+        print(message)
+    except UnicodeEncodeError:
+        encoding = sys.stdout.encoding or "ascii"
+        print(message.encode(encoding, "replace").decode(encoding, "replace"))
 
 
 def compare_json_structures(truth, to_compare, path=""):
@@ -49,12 +60,12 @@ def compare_json_structures(truth, to_compare, path=""):
 
 
 def lint_translation_json(file, truth, to_compare):
-    print(f"\nLinting {file}...")
+    _safe_print(f"\nLinting {file}...")
 
     errors = compare_json_structures(truth, to_compare)
 
     if errors:
         for error in errors:
-            print(f"{error}")
+            _safe_print(error)
     else:
-        print(f"✅ No errors found in {file}")
+        _safe_print(f"✅ No errors found in {file}")
--- a/backend/tests/test_translations.py
+++ b/backend/tests/test_translations.py
@@ -1,3 +1,5 @@
+import io
+import sys
 from io import StringIO
 from unittest.mock import patch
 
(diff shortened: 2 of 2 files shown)


mussaratshamsher:
## Summary

Fixes #2997.

This change adds a small `_safe_print()` helper to gracefully handle `UnicodeEncodeError` when `lint-translations` is run on Windows consoles using legacy code pages (for example, cp1252).

Instead of crashing when printing Unicode status markers (✅, ❌, ⚠️), the output is encoded using the current stdout encoding with replacement characters when the console cannot represent those symbols.

## Changes

- Add `_safe_print()` helper.
- Route all output in `lint_translation_json()` through `_safe_print()`.
- Add a regression test that simulates a `cp1252` stdout stream.

## Testing

```bash
python -m pytest backend/tests/test_translations.py -v
python -m ruff check backend/chainlit/translations.py backend/tests/test_translations.py

PR #3014, opened 2026-08-19
fix(a11y): announce that a reply is generating, and that one failed

opensource-joe:
The chat UI has no live regions anywhere, so a screen reader user gets no signal that a reply is coming or that one failed. I checked before assuming: zero `aria-live`, `role="status"`, `role="log"`, `role="alert"` or `aria-atomic` across all 208 files in `frontend/src`.

This adds the two that matter most. It is a small change and I have kept it deliberately narrow.

## Generating

The only cue that a reply is on its way is `BlinkingCursor`, which is an empty `<span>` animated by CSS. It has no text, so there is nothing for assistive technology to announce.

I added a visually hidden `role="status"` region in `MessagesContainer`, driven by the same `loading` value that drives the cursor. The announcement therefore covers exactly the period the cursor covers, no more and no less.

**It announces the state, not the reply.** Wrapping streamed content in a live region makes assistive technology re-read the partial message on every chunk, which is worse than saying nothing.

## Failed

An errored reply is currently distinguished only by a red avatar and red text, which is colour alone. `role="alert"` goes on the message wrapper when `message.isError` is set. These are complete messages rather than streamed ones, so an alert does not re-read.

## Verified against a running app

Not just a build. I ran the server from source with a deliberately slow-streaming app and read the DOM live.

While a reply was generating:

```json
{ "status": "Generating response", "steps": 2, "alerts": 0 }
```

and once it completed the region emptied again, so it announces once per reply rather than repeating:

```json
{ "status": "", "steps": 2, "alerts": 0 }
```

Sending a `cl.ErrorMessage`:

```json
{ "alertCount": 1,
  "alertText": ["Something went wrong generating that reply."],
  "aiMessages": [{ "role": "alert", "text": "Something went wrong generating that reply." }] }
```

Also green: `pnpm type-check`, `pnpm lint` and `pnpm format-check` on the changed files, `prettier --check` on all 23 locale files, and `pnpm run --recursive test` at 32 passed.

## Two things worth your call

**Translations.** The new key is in all 23 locale files, since they are otherwise kept in sync at 136 keys each. The non-English values are the English string left untranslated rather than machine translated, so they can go to translators. `fallbackLng` means behaviour is identical either way, so I am happy to reduce this to `en-US.json` only if you would rather not carry placeholder values.

**No unit test.** I tried. `Message` reaches `useLayoutMaxWidth` and then `useConfig`, which needs a Recoil provider the current `frontend/tests` setup does not have, and mocking the tree ended up larger than the change itself. I would rather say that than add scaffolding you have deliberately avoided. Happy to add one if you can point me at the pattern you would want.

## Not included on purpose

`Loader` is a bare spinning `LoaderIcon` with no accessible name, used in `App.tsx`, `ReadOnlyThread.tsx`, `Messages/index.tsx` and `ActionButton.tsx`. That is a real gap and a separate change, so I left it out rather than widen this one. Glad to follow up if you want it.

---
## Summary by cubic
Announces reply generation and failures for screen readers to meet a11y expectations. Previously there were no live regions; now a hidden `role="status"` announces "Generating response" while loading, and errored replies are marked with `role="alert"`.

- `MessagesContainer`: adds a screen-reader-only `role="status"` region bound to `loading`; does not wrap streamed content to avoid re-reads.
- `Message`: sets `role="alert"` when `message.isError`; visual behavior unchanged.
- Adds `chat.messages.status.generating` to all locale files with English placeholders; `fallbackLng` keeps runtime behavior identical.
- No API or UI changes; no migration required.

<sup>Written for commit 0864fc18de641fcd07785ba0589c049554db898d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Chainlit/chainlit/pull/3014?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


opensource-joe:
Worth taking seriously, and it made me test something I had asserted without testing. **I was wrong in the PR description, so let me correct that first.**

I wrote that `fallbackLng` makes the behaviour identical whether or not the key is present in the non-English files. It does not. Chainlit's own translation wrapper never reaches i18next's fallback, at `frontend/src/components/i18n/Translator.tsx:36`:

```ts
t: (path, options) => {
  if (!i18n.exists(path, options)) {
    return '...';
  }
  return t(path, options);
}
```

Two other things point the same way. `AppWrapper.tsx:21` registers only the selected language with `addResourceBundle(languageInUse, ...)`, so there is no `en-US` bundle in memory to fall back to. And `config.load_translation()` returns the matched locale file verbatim on an exact match, with no key-level merge.

I checked it rather than reasoning from the source. Running with `language = "bn"`, with the key present the live region announces the English string. With the key deleted from the served `bn.json`, the region holds `...` for the whole generation:

```
[{"t":"0.0","status":""},{"t":"0.6","status":"..."},{"t":"1.2","status":"..."},
 {"t":"1.8","status":"..."},{"t":"2.4","status":"..."}]
```

So removing the placeholder does not give Bengali users Bengali. It gives them `...`, which is nothing useful to a screen reader, and that is a worse outcome than English for exactly the people this change is meant to help.

That is why the key is in all 23 files. The values are untranslated English rather than machine translation, so they can go to translators. If you would rather not carry placeholders at all, the option that does not regress anything is to translate them, and I am happy to do that as a follow-up rather than hold this one up.

PR #3021, opened 2026-08-26
fix(e2e): de-flake thread_resume by waiting on the resume replay

### LINKED ISSUE #3019: flaky(e2e): `thread_resume` intermittently fails on a detached `#chat-submit`
## Summary

`cypress/e2e/thread_resume/spec.cy.ts` fails roughly **20–33% of runs** with a detached-DOM error when clicking the send button. It is not tied to any dependency version — it reproduces across `react-router-dom` 6.30.4, 6.30.5 and 6.30.6 with an identical stack — and it burns ~2 minutes per failure because Cypress retries the spec four times before giving up.

The defect is in the test helper, not in the application.

## Error

```
CypressError: Timed out retrying after 30050ms: `cy.click()` failed because
the page updated while this command was executing. Cypress tried to locate
elements based on this query:
> <button#chat-submit ...>
We initially found matching element(s), but while waiting for them to become
actionable, they disappeared from the page.
```

## Root cause

`cypress/support/testUtils.ts:16`:

```ts
export function submitMessage(message: string) {
  cy.get('#chat-input')
    .should('be.visible')
    .should('not.be.disabled')
    .type(message);
  cy.get('#chat-submit').should('not.be.disabled').click();   // <-- here
}
```

`cy.get('#chat-submit')` resolves the button once. `.should('not.be.disabled')` then retries the *assertion* against that already-resolved subject, and `.click()` finally acts on it. If React re-renders in the window between the assertion passing and the click landing, the original node is detached and the click fails.

`thread_resume` is the spec that hits this most often because it is the one that calls `submitMessage` **immediately after `cy.reload()`**:

```ts
cy.reload();                                  // triggers thread resume
cy.get('#message-composer').should('be.visible');
submitMessage('still here');                  // composer still settling
```

On resume the composer remounts as the thread's history streams back in, which is exactly the re-render that detaches the button. Other specs call `submitMessage` against a settled page, so they rarely trip it.

## Evidence

Isolated runs of just this spec, rebuilding between each version change:

| `react-router-dom` | Result |
|---|---|
| 6.30.4 | 2 pass / 0 fail |
| 6.30.5 | 2 pass / 1 fail |
| 6.30.6 | 4 pass / 1 fail |

Same error signature in every failure. The version is not the variable.

## Reproduction

```bash
pnpm build
cd backend && uv sync --all-extras && cd ..
# repeat — expect an intermittent failure within a handful of runs
for i in $(seq 1 5); do
  pnpm exec cypress run --spec "cypress/e2e/thread_resume/spec.cy.ts"
done
```

## Proposed fix

Break the chain so the element is re-queried at click time:

```ts
export function submitMessage(message: string) {
  cy.get('#chat-input')
    .should('be.visible')
    .should('not.be.disabled')
    .type(message);
  cy.get('#chat-submit').should('not.be.disabled');
  cy.get('#chat-submit').click();
}
```

This fixes every caller at once, since they all funnel through `submitMessage`.

If it still proves racy after resume specifically, the sturdier option is to wait for the app to be idle before submitting — e.g. asserting the resumed history has finished rendering in `thread_resume` before the second `submitMessage`, rather than relying on `#message-composer` being visible (which becomes true before the thread finishes hydrating).

## Impact

- ~2 minutes of CI time per occurrence (4 retries × 30s timeout)
- Produces red builds that look like real regressions on unrelated PRs — it caused a false "regression" diagnosis while reviewing #2999 before sampling showed it to be version-independent

## Notes

Pre-existing on `main`; not introduced by any recent dependency work.

dokterbob:
Fixes #3019.

## What the flake actually is

The issue diagnosed a stale-subject bug in `submitMessage` — `cy.get('#chat-submit').should('not.be.disabled').click()` acting on an already-resolved node. That's a real smell and is fixed here, but it isn't the mechanism.

Composer ownership moves between two different components as the message list changes:

- `frontend/src/components/chat/Footer.tsx:20` returns `null` while `!hasMessage(messages)`
- `frontend/src/components/chat/WelcomeScreen.tsx:80,91` returns `null` once `hasMessage(messages)`, and otherwise renders its **own** `MessageComposer`

So every flip of `hasMessage()` unmounts one composer and mounts another. Two points in this spec crossed that boundary mid-interaction:

1. **First turn.** `on_chat_start` sends its welcome message asynchronously. Landing during `type()` remounts the input, loses the typed value, and leaves `#chat-submit` disabled.
2. **Resume.** `connection_successful` (`backend/chainlit/socket.py:208-240`) emits `first_interaction` → the `@cl.on_chat_resume` message → `resume_thread`. The client handler (`libs/react-client/src/useChatSession.ts:258+`) rebuilds the list from `thread.steps` and **replaces** it wholesale. Because this fixture's `MemoryDataLayer.create_step` was a no-op and `steps` was always `[]`, that replay drove the list back to empty — swapping the composer a *second* time, exactly where the test clicks.

`cy.get('#message-composer').should('be.visible')` is satisfied early in that sequence, well before hydration finishes, so it never gated anything.

The underlying app behaviour is filed separately as #3022 — this PR makes the test honest about the churn, it does not remove the churn.

## Changes

**`cypress/e2e/thread_resume/main.py`** — implement `create_step` / `update_step` / `delete_step` on the fixture's `MemoryDataLayer` so steps are actually persisted, with `@queue_until_user_message()` matching `data/sql_alchemy.py:383-421`. `ensure_thread` backfills a row when a step is persisted before the first `update_thread`, and re-homes it when the owner arrives rather than duplicating it.

This is the crux, not incidental cleanup. With `steps: []` the resume ends in a state indistinguishable from a fresh chat, so there was **no observable signal that hydration had finished** to wait on. Persisting steps gives the replay a stable end state — `hasMessage()` now stays true across it, so the composer no longer remounts — and means the spec finally exercises thread resume instead of resuming an empty thread.

**`cypress/e2e/thread_resume/spec.cy.ts`** — wait for the welcome message before the first submit, and after reload wait on `Echo: hi`, which only the replay can produce and which is the last write to the message list. No `cy.wait(<ms>)`.

**`cypress/support/testUtils.ts`** — split the query chains in `submitMessage` / `openHistory` so each action re-queries. Signatures unchanged; all 34 call sites unaffected.

## Verification

Local: `pnpm lint cypress/`, Prettier, and `ruff check` / `ruff format --check` all clean. Note `pnpm type-check` does **not** cover `cypress/`.

CI: three attempts of run [32954779496](https://github.com/Chainlit/chainlit/actions/runs/32954779496), all on this commit, all green. Because `retries: 3` would mask a surviving flake behind a green badge, I pulled all 30 shard logs and checked `(Attempt N of 4)` annotations directly:

| Attempt | `thread_resume` (ubuntu-5) | `thread_resume` (windows-5) | Retries in run |
|---|---|---|---|
| 1 | ✓ first attempt, 3664ms | ✓ first attempt, 4143ms | none |
| 2 | ✓ first attempt, 4150ms | ✓ first attempt, 3399ms | none |
| 3 | ✓ first attempt, 3032ms | ✓ first attempt, 4594ms | one, unrelated (see below) |

Six first-attempt passes, no retry annotations on this spec anywhere. All 30 shard-runs ended `Failing: 0`.

Being straight about the strength of this: at the 20–33% rate reported in #3019, three clean runs would happen by luck roughly 30–50% of the time, so the runs alone aren't proof. The mechanism argument is the stronger half — `hasMessage()` no longer flips during the resume, so the remount that detached the button cannot occur — and the retry data rules out the specific way a survivor would have hidden.

Attempt 3 did contain one genuine retry, in `oauth_auth` on `windows-latest-3`, unrelated to this spec and pre-existing. Filed as #3023, with the general "retries are invisible in CI" problem as #3024.

## Review notes

- `2099804` is wrong and `c13ae25` corrects it; left in history rather than force-pushed so the reasoning is visible. Happy to squash on merge.
- Known-remaining, deliberately out of scope: `submitMessage` is still exposed to the same remount race in other specs that call it against an empty message list (#3022 is the real fix); `[data-testid="read-only-banner"]` at `spec.cy.ts:48` asserts a testid that exists nowhere in the frontend and is therefore vacuous (noted in #3024); ~27 other `.should(...).<action>()` sites remain across `cypress/e2e/`.

### DIFF
--- a/cypress/e2e/thread_resume/main.py
+++ b/cypress/e2e/thread_resume/main.py
@@ -3,6 +3,7 @@ from typing import Dict, List, Optional
 
 import chainlit as cl
 import chainlit.data as cl_data
+from chainlit.data.utils import queue_until_user_message
 from chainlit.element import Element, ElementDict
 from chainlit.step import StepDict
 from chainlit.types import (
@@ -21,6 +22,42 @@ now = utc_now()
 
 # Simple in-memory persistence for threads per user
 THREADS: Dict[str, List[ThreadDict]] = {}
+
+
+def find_thread(thread_id: str) -> Optional[ThreadDict]:
+    """Return the stored thread with this id, whichever user bucket holds it."""
+    for threads in THREADS.values():
+        for thread in threads:
+            if thread["id"] == thread_id:
+                return thread
+    return None
+
+
+def ensure_thread(thread_id: str, user_id: Optional[str] = None) -> ThreadDict:
+    """Return the stored thread with this id, creating the row if needed.
+
+    Steps can be persisted before the first `update_thread` call, so a row may
+    have to be created without an owner. Re-home it when the owner shows up
+    instead of appending a second, empty row under the user.
+    """
+    thread = find_thread(thread_id)
+    if thread is None:
+        thread = {
+            "id": thread_id,
+            "createdAt": utc_now(),
+            "userId": user_id,
+            "userIdentifier": user_id,
+            "name": thread_id,
+            "steps": [],
+        }
+        THREADS.setdefault(user_id or "", []).append(thread)
+    elif user_id and thread.get("userIdentifier") != user_id:
+        for uid, threads in THREADS.items():
+            THREADS[uid] = [t for t in threads if t is not thread]
+        thread["userId"] = user_id
+        thread["userIdentifier"] = user_id
+        THREADS.setdefault(user_id, []).append(thread)
+    return thread
 
 
 class MemoryDataLayer(cl_data.BaseDataLayer):
@@ -55,14 +92,26 @@ class MemoryDataLayer(cl_data.BaseDataLayer):
     async def delete_element(self, element_id: str, thread_id: Optional[str] = None):
         pass
 
+    @queue_until_user_message()
     async def create_step(self, step_dict: "StepDict"):
-        pass
+        thread_id = step_dict.get("threadId")
+        if not thread_id:
+            return
+        steps = ensure_thread(thread_id)["steps"]
+        for index, existing in enumerate(steps):
+            if existing["id"] == step_dict["id"]:
+                steps[index] = step_dict
+                return
+        steps.append(step_dict)
 
+    @queue_until_user_message()
     async def update_step(self, step_dict: "StepDict"):
-        pass
+        await self.create_step(step_dict)
 
+    @queue_until_user_message()
     async def delete_step(self, step_id: str):
-        pass
+        for thread in (t for threads in THREADS.values() for t in threads):
+            thread["steps"] = [s for s in thread["steps"] if s["id"] != step_id]
 
     async def get_thread_author(self, thread_id: str) -> str:
         return (await self.get_thread(thread_id))["userIdentifier"]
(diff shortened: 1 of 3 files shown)


dokterbob:
@codex review

PR #3028, opened 2026-08-27
fix(frontend): preserve composer draft across remounts

### LINKED ISSUE #3022: Composer remounts whenever the message list transitions empty↔non-empty, dropping in-progress input
Found while diagnosing #3019. That issue is a test flake; this is the application behaviour underneath it, which outlives the test fix in #3021.

## The mechanism

Two components render a `MessageComposer`, and which one owns it depends on whether the message list is empty:

- `frontend/src/components/chat/Footer.tsx:20` — `if (!hasMessage(messages) && !showIfEmptyThread) return null;`
- `frontend/src/components/chat/WelcomeScreen.tsx:80,91` — `if (hasMessage(messages)) return null;` … otherwise renders its **own** `<MessageComposer {...props} />`

`hasMessage()` (`frontend/src/lib/utils.ts:10`) is true when any step is of type `user_message`, `assistant_message`, or `tool`.

So every transition of `hasMessage()` unmounts one composer instance and mounts a different one. Because `MessageComposer` holds its draft in local state, **the remount discards whatever the user had typed**, and the send button then reads as disabled because the new instance's value is empty (`MessageComposer/index.tsx:339-341`).

## Why this is user-facing, not just a test artifact

The transition fires on ordinary app events, not only in tests:

1. **First turn of a chat.** `@cl.on_chat_start` sends its message asynchronously. A user who starts typing before that message lands can have their input silently dropped.
2. **Thread resume.** `connection_successful` (`backend/chainlit/socket.py:208-240`) emits the `@cl.on_chat_resume` message and then `resume_thread`; the client handler (`libs/react-client/src/useChatSession.ts:258+`) rebuilds the list from `thread.steps` and **replaces** it wholesale. If the persisted history is empty, the list goes `[] → [resume message] → []`, remounting the composer twice in quick succession.

The faster the backend, the narrower the window — which is exactly why this reads as intermittent.

## Suggested direction

Render the composer once, above the `WelcomeScreen`/`Footer` branch, so its identity is stable across the transition — or lift the draft value into shared state so a remount preserves it. Either removes a class of "my message vanished" reports that would be near-impossible to reproduce from a bug report.

## Notes

- #3021 works around this for `cypress/e2e/thread_resume/spec.cy.ts` by waiting for the transitions to settle before interacting. That makes the test honest, but it does not address the behaviour.
- Related, likely the same root cause: the `on_chat_resume` message is always erased by the `resume_thread` replay that follows it, so `@cl.on_chat_resume` output is only visible for a few frames unless it happens to be in the persisted steps. That may be worth its own issue if it is not intended.

Iams4kura:
## Summary

- Fixes #3022.
- Fixes: When the message list changes between empty and non-empty, the welcome composer is replaced by the footer composer and text the user is still typing disappears.
- Root cause: WelcomeScreen and Footer own separate MessageComposer instances, while both MessageComposer and Input stored the draft only in component-local state; switching owners therefore unmounted the only copy of the draft.

## Regression evidence

- Before: `npx --yes pnpm@9.15.9 --dir frontend exec vitest run tests/MessageComposer.spec.tsx` exited `1`

- After: `npx --yes pnpm@9.15.9 --dir frontend exec vitest run tests/MessageComposer.spec.tsx` exited `0`

## Verification

- `npx --yes -p node@24 -p pnpm@9.15.9 pnpm test`
- `npx --yes -p node@24 -p pnpm@9.15.9 pnpm lint`
- `npx --yes -p node@24 -p pnpm@9.15.9 pnpm format-check`
- `npx --yes -p node@24 -p pnpm@9.15.9 pnpm type-check`
- `npx --yes -p node@24 -p pnpm@9.15.9 pnpm build`
- `npx --yes -p node@24 -p pnpm@9.15.9 pnpm check-overrides`

## Scope

- 4 files changed, +192 / -17 lines

---
## Summary by cubic
Preserves the composer draft when the message list switches between empty and non-empty states, so user-typed text no longer disappears when the welcome composer is replaced by the footer composer. The draft now lives in shared state per session instead of only in component-local state. New sessions start empty, and a URL prompt no longer overwrites an edited draft after a remount. Fixes #3022.

<sup>Written for commit 9d06696b95de555ce944317ffdfc8858ab4f3d71. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Chainlit/chainlit/pull/3028?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/components/chat/MessageComposer/Input.tsx
+++ b/frontend/src/components/chat/MessageComposer/Input.tsx
@@ -26,6 +26,7 @@ interface Props {
   className?: string;
   autoFocus?: boolean;
   placeholder?: string;
+  value: string;
   selectedCommand?: ICommand;
   setSelectedCommand: (command: ICommand | undefined) => void;
   onChange: (value: string) => void;
@@ -45,6 +46,7 @@ const Input = forwardRef<InputMethods, Props>(
       id,
       className,
       autoFocus,
+      value,
       selectedCommand,
       setSelectedCommand,
       onChange,
@@ -57,7 +59,6 @@ const Input = forwardRef<InputMethods, Props>(
     const [isComposing, setIsComposing] = useState(false);
     const [showCommands, setShowCommands] = useState(false);
     const [commandInput, setCommandInput] = useState('');
-    const [value, setValue] = useState('');
     const textareaRef = useRef<HTMLTextAreaElement>(null);
 
     const normalizedInput = commandInput.toLowerCase().slice(1);
@@ -88,7 +89,6 @@ const Input = forwardRef<InputMethods, Props>(
     });
 
     const reset = () => {
-      setValue('');
       if (!selectedCommand?.persistent) {
         setSelectedCommand(undefined);
       }
@@ -100,7 +100,6 @@ const Input = forwardRef<InputMethods, Props>(
     useImperativeHandle(ref, () => ({
       reset,
       setValueExtern: (value: string) => {
-        setValue(value);
         onChange(value);
       }
     }));
@@ -113,7 +112,6 @@ const Input = forwardRef<InputMethods, Props>(
 
     const handleChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
       const newValue = e.target.value;
-      setValue(newValue);
       onChange(newValue);
 
       // Command detection for dropdown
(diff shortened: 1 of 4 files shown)


Iams4kura:
Closing this before maintainer review because the published commit omitted this repository's required Co-Authored-By trailer. I will not rewrite the already-published branch history to conceal that mistake.

Issue #3023, opened 2026-08-26
flaky(e2e): oauth_auth "shows a specific message for oauthSignin error" retried on windows-latest


dokterbob:
Caught incidentally while auditing CI logs for #3019 / #3021. **The run reported green** — this only surfaced by reading the raw shard logs.

## What happened

Run [32954779496](https://github.com/Chainlit/chainlit/actions/runs/32954779496), attempt 3, shard `e2e-tests / windows-latest-3` (job `98139320253`):

```
(Attempt 1 of 4) shows a specific message for oauthSignin error     10:14:32.65
√ shows a specific message for oauthSignin error (1351ms)           10:14:34.09
```

Spec: `cypress/e2e/oauth_auth/spec.cy.ts`. It failed once and passed on the retry, so `retries: 3` absorbed it and the job finished `All specs passed! — 43 passing, 0 failing`.

## Evidence it is a genuine flake

Cross-checked against 20 full shard logs (10 shards × attempts 2 and 3), plus attempt 1:

- The `(Attempt N of ...)` annotation appears **nowhere else** in any log.
- That job uniquely reports `Screenshots: 1`, with `... oauthSignin error (failed).png` — a captured failure.
- Its "Upload screenshots" step ran (`success`); the same step is `skipped` on every other shard, i.e. nothing failed there.
- The test normally completes in ~300–700ms wherever else it runs. This occurrence showed a 30s+ gap before the retry line, consistent with a command timing out rather than a fast assertion failure.

Only observed on `windows-latest` so far; `ubuntu-latest-3` ran the same spec cleanly in all three attempts. One occurrence, so the rate is unknown.

## Why it is worth filing despite being green

This is precisely the failure mode that makes flakiness accumulate unnoticed: a real intermittent failure occurred, CI reported success, and nothing anywhere surfaced it. Finding it required pulling and grepping 20 job logs by hand. See the companion issue on retry reporting.

## Starting points

- The 30s gap points at a Cypress command timeout (`defaultCommandTimeout: 30000` in `cypress.config.ts`), so the likely shape is a wait on an element or route that did not appear, rather than a wrong assertion.
- The failure screenshot is in the run artifacts (`cypress-screenshots-windows-latest-3`), though artifact retention may expire it.

PR #3033, opened 2026-08-31
ci(e2e): report Cypress retries in CI so flaky specs are visible

### LINKED ISSUE #3024: Surface Cypress retries in CI so flaky specs stop hiding behind green runs
Follow-up from #3019 / #3021. `cypress.config.ts` sets `retries: 3`, so a spec that fails and then passes reports the job green and leaves no trace anywhere a human will look. #3023 is a live example: a genuine `oauth_auth` failure on `windows-latest-3` was absorbed by a retry, and finding it took pulling and grepping 20 job logs by hand.

Retries are the right call for CI stability. The gap is that nothing *reports* them.

## Option A — the plumbing is already here but switched off

`cypress.config.ts:28` sets `projectId: 'ij1tyk'`, and `.github/workflows/e2e-tests.yaml` already passes `CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}` into the test step — but `pnpm test:e2e` is `cypress run` with no `--record` flag, so **nothing is ever recorded**. Cypress Cloud has flake detection and PR reporting built in, so this may be a one-flag change.

Needs a maintainer decision rather than a contributor patch: it depends on an active Cloud plan and a valid secret. Two caveats if picked up — forked PRs will not have the secret, and `cypress-split` is a different sharding mechanism from Cloud's `--parallel`, so keep the current splitting and only record.

## Option B — self-contained retry reporting (no external service)

Cypress exposes retries through the `after:spec` node event ([docs](https://docs.cypress.io/api/node-events/after-spec-api)):

```ts
on('after:spec', (spec, results) => {
  const retried = results.tests.some((test) =>
    test.attempts.some((attempt) => attempt.state === 'failed')
  )
})
```

An attempt with `state: 'failed'` followed by a pass **is** a retry. Convenient starting point: `cypress.config.ts:50` already registers an `after:spec` handler for `killChainlit` and currently discards both arguments.

Sketch:

1. Each shard writes a small JSON of retried tests (spec, title, attempt count).
2. Upload it as an artifact.
3. An aggregating job merges the shards and renders a table into `$GITHUB_STEP_SUMMARY`.
4. Optionally post/update a PR comment **only when the retry count is non-zero**, so clean runs stay silent.

Two constraints worth knowing before starting:

- `.github/workflows/e2e-tests.yaml:11` and `.github/workflows/ci.yaml:15` both set `permissions: read-all`. A PR comment needs `pull-requests: write` on the aggregating job.
- On forked PRs `GITHUB_TOKEN` is read-only regardless, so make the comment best-effort and `$GITHUB_STEP_SUMMARY` the primary output. Please do **not** reach for `pull_request_target` to work around this — it runs untrusted code with write credentials.

Option B works on forks, has no vendor dependency, and produces a machine-readable per-shard record that could later be trended to catch a spec degrading before it starts failing outright. The two options are not exclusive.

## Unrelated small item, same area

`cypress/e2e/thread_resume/spec.cy.ts:48` asserts `cy.get('[data-testid="read-only-banner"]').should('not.exist')`, but that testid appears nowhere in `frontend/` or `libs/` — so the assertion can never fail and is checking nothing. Either the frontend is missing the attribute (and the assertion is right but inert) or the assertion is dead and should go. Someone needs to work out which; it is left alone in #3021 as out of scope.

fra-shipper:
Fixes #3024.

## Root cause

`cypress.config.ts` sets `retries: 3`, so a spec that fails and then passes on retry reports the job green with no trace anywhere a human will look. #3023 is a concrete example: a genuine `oauth_auth` failure on `windows-latest-3` was absorbed by a retry, and finding it required pulling and grepping job logs by hand. The existing `after:spec` handler discarded both of its arguments and only called `killChainlit()`, so this information was never captured.

## Fix

Implements Option B from the issue (self-contained, no Cypress Cloud dependency, works on forked PRs):

1. `cypress/support/retryReport.ts` adds `collectRetriedTests`, a pure function that flags a test as retried when its overall state is `passed` but at least one attempt has `state === 'failed'`. Tests that exhaust every retry and still fail are excluded, since those are already visible via the job's exit status.
2. `cypress.config.ts`'s `after:spec` handler now receives `(spec, results)`, accumulates retried tests for the shard, and writes them to `cypress/reports/retries.json` after every spec (so a killed job still leaves partial data).
3. `.github/workflows/e2e-tests.yaml` uploads that file as a per-shard artifact (`cypress-retries-<os>-<containers>`, same pattern as the existing screenshot upload), and a new `report-retries` job downloads every shard's artifact, merges them, and writes a table to `$GITHUB_STEP_SUMMARY`.
4. When retries are found on a same-repo PR, the job also posts (or updates) a single PR comment. Top-level `permissions: read-all` is unchanged; the `report-retries` job itself is granted `pull-requests: write`, and `.github/workflows/ci.yaml`'s `e2e-tests:` call site now also grants `contents: read` / `pull-requests: write` on the reusable-workflow job entry — a reusable workflow can only narrow permissions inherited from its caller, never widen them, so without this the grant inside `e2e-tests.yaml` would have had no effect even on same-repo PRs. On forked PRs `GITHUB_TOKEN` is read-only, so the comment step is skipped ahead of time based on `pull_request.head.repo.full_name` and additionally wrapped in try/catch as a fallback; it never uses `pull_request_target`. The step summary is always the primary, unconditional output.

The unrelated dead `read-only-banner` testid assertion in `cypress/e2e/thread_resume/spec.cy.ts` flagged in the issue as a separate item is left untouched, as the issue itself scopes it out.

## Testing

- `node --test cypress/support/retryReport.test.ts` -> 3/3 pass (a test passing on first attempt is ignored, a test that fails then passes is reported with its attempt count, a test that fails every attempt is ignored). Verified this is a genuine regression test: temporarily reverted `collectRetriedTests` to a no-op returning `[]` (matching the old discard-everything `after:spec` behavior) and reran — the retry-detection case failed with an `AssertionError` as expected — then restored the real implementation and confirmed 3/3 pass again.
- `npx prettier --check` on all touched files -> clean.
- `npx eslint` on `cypress.config.ts`, `cypress/support/retryReport.ts`, `cypress/support/retryReport.test.ts` -> clean.
- `npx tsc --noEmit -p tsconfig.json` -> no new errors from this change (only 4 pre-existing `TS2428` WeakMap errors from Cypress's vendored lodash types, confirmed present identically on a clean checkout via `git stash -u`, and unrelated to this change — this tsconfig is also not wired into any CI script; `pnpm type-check` only recurses into the `frontend`/`libs` workspace packages).
- `actionlint` on the modified workflow files -> clean, exit 0.
- The modified `e2e-tests.yaml` was parsed with `yaml.safe_load` to confirm it still parses correctly and the job list is `['prepare', 'e2e-tests', 'report-retries']` as intended.
- The `permissions:` fix in `ci.yaml` (a reusable-workflow call site can only narrow, never widen, the caller's inherited permissions) was verified by re-reading the affected files against GitHub's documented permission-inheritance semantics for `workflow_call`, since that runtime behavior cannot be exercised locally or with `actionlint` (which validates syntax, not inheritance semantics).
- Husky's pre-commit hook (`lint-staged`, which runs `format:files`, `lint:fix`, and `actionlint` on staged files) ran on every commit and made no further modifications, confirming the diff was already compliant.
- Not run: `pnpm test:e2e` (the full Cypress suite), since it requires the full Chainlit backend and a live browser; the pure retry-detection logic is covered by the `node:test` regression test instead, and the artifact-upload/aggregation job will be exercised directly by this repo's own `e2e-tests` CI workflow on this PR.

---
## Summary by cubic
Makes Cypress retries visible in CI so specs that fail once and pass on retry no longer produce a silently green run (#3024). The workflow now records and reports those tests without changing the pass/fail result; tests that still fail remain governed by the existing job failure.

- Writes per-shard `cypress/reports/retries.json`, uploads it, and merges available reports into `$GITHUB_STEP_SUMMARY`.
- Creates or updates one marker PR comment on same-repo PRs, including clearing it when a later run has no retries; forked PRs use the summary because their token is read-only.
- Grants `pull-requests: write` through the reusable workflow call and adds the retry-report unit tests to CI via `pnpm test:unit`.
- Guards the report hook when Cypress runs interactively, where results are unavailable.

<sup>Written for commit 209adbfcfc6527fc15ff6bd7834a3ff9b89eee5e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Chainlit/chainlit/pull/3033?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,5 +6,5 @@
     "esModuleInterop": true
   },
   "include": ["cypress/**/*.ts", "cypress.config.ts"],
-  "exclude": ["cypress/**/*.test.ts"]
+  "exclude": ["cypress/support/retryReport.test.ts"]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,8 +3,9 @@
     "lib": ["ESNext", "dom"],
     "types": ["cypress", "cypress-plugin-steps", "node"],
     "baseUrl": ".",
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "allowImportingTsExtensions": true,
+    "noEmit": true
   },
-  "include": ["cypress/**/*.ts", "cypress.config.ts"],
-  "exclude": ["cypress/support/retryReport.test.ts"]
+  "include": ["cypress/**/*.ts", "cypress.config.ts"]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,5 +5,6 @@
     "baseUrl": ".",
     "esModuleInterop": true
   },
-  "include": ["cypress/**/*.ts", "cypress.config.ts"]
+  "include": ["cypress/**/*.ts", "cypress.config.ts"],
+  "exclude": ["cypress/**/*.test.ts"]
 }
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,6 +27,9 @@ jobs:
     secrets: inherit
   e2e-tests:
     uses: ./.github/workflows/e2e-tests.yaml
+    permissions:
+      contents: read
+      pull-requests: write
     secrets: inherit
   ci:
     runs-on: ubuntu-slim
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -73,3 +73,112 @@ jobs:
         with:
           name: cypress-screenshots-${{ matrix.os }}-${{ matrix.containers }}
           path: cypress/screenshots
+      - name: Upload retry report
+        uses: actions/upload-artifact@v7
+        if: always() && hashFiles('cypress/reports/retries.json') != ''
+        with:
+          name: cypress-retries-${{ matrix.os }}-${{ matrix.containers }}
+          path: cypress/reports/retries.json
+
+  report-retries:
+    name: Report flaky test retries
+    needs: e2e-tests
+    if: always()
+    runs-on: ubuntu-slim
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
+      - name: Download retry reports
+        uses: actions/download-artifact@v8
+        with:
+          pattern: cypress-retries-*
+          path: retry-reports
+        continue-on-error: true
+      - name: Summarize retried tests
+        uses: actions/github-script@v9
+        with:
+          script: |
+            const fs = require('fs');
+            const path = require('path');
+
+            const root = 'retry-reports';
+            const retries = [];
+
+            if (fs.existsSync(root)) {
+              for (const shard of fs.readdirSync(root)) {
+                const reportFile = path.join(root, shard, 'retries.json');
+                if (!fs.existsSync(reportFile)) continue;
+                const entries = JSON.parse(fs.readFileSync(reportFile, 'utf8'));
+                for (const entry of entries) {
+                  retries.push({ shard, ...entry });
+                }
+              }
+            }
+
+            if (retries.length === 0) {
+              await core.summary
+                .addHeading('Cypress retries', 3)
+                .addRaw('No specs needed a retry in this run.')
+                .write();
+              return;
+            }
+
+            await core.summary
+              .addHeading('Cypress retries', 3)
+              .addTable([
+                [
+                  { data: 'Shard', header: true },
+                  { data: 'Spec', header: true },
+                  { data: 'Test', header: true },
+                  { data: 'Attempts', header: true }
+                ],
+                ...retries.map((retry) => [
+                  retry.shard,
+                  retry.spec,
+                  retry.title,
+                  String(retry.attempts)
+                ])
+              ])
+              .write();
+
+            // Best-effort PR comment: on forked PRs GITHUB_TOKEN is read-only,
+            // so a failure here must not fail the job. The step summary above
+            // is the primary, always-available output.
+            const pullRequest = context.payload.pull_request;
+            const isFork =
+              pullRequest &&
+              pullRequest.head.repo.full_name !== `${context.repo.owner}/${context.repo.repo}`;
+            if (!pullRequest || isFork) return;
+
+            const marker = '<!-- cypress-retry-report -->';
+            const body = `${marker}\n**${retries.length} Cypress test attempt(s) were retried in this run.** See the job summary for details.`;
+
+            try {
+              const { data: comments } = await github.rest.issues.listComments({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: pullRequest.number
+              });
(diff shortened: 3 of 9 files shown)


fra-shipper:
Addressed the review findings in 209adbf:

- `cypress.config.ts`: guarded `after:spec` against `results` being `undefined`, which Cypress passes in interactive mode (`cypress open`). Confirmed this by tracing the type definitions and reproducing the type error path; the dereference previously threw before `killChainlit()` ran.
- `.github/workflows/e2e-tests.yaml`: the zero-retries early return no longer skips comment reconciliation, so a later clean run updates a stale marker comment instead of leaving it in place. Paginated the comment lookup with `github.paginate` so the marker is found on PRs with many comments. Fixed the wording from "N test attempt(s)" to "N test(s)" since each `retries[]` entry is one retried test, which can hold multiple attempts.
- `tsconfig.json`: dropped the `exclude` on `retryReport.test.ts` and instead enabled `allowImportingTsExtensions` + `noEmit`, so the file (which imports with an explicit `.ts` extension for Node's native type-stripping test runner) type-checks in place. Verified with `tsc --noEmit` before/after.
- `package.json` / `.github/workflows/tests.yaml`: added a `test:unit` script (`node --test cypress/support/*.test.ts`) and wired it into the existing "Unit & integration tests" workflow (once, not per Python-version matrix leg) so `retryReport.test.ts` actually runs in CI.

Verified: `prettier --check`, `eslint`, `tsc --noEmit` on the touched TS files, `pnpm --filter @chainlit/app type-check`, and `node --test cypress/support/retryReport.test.ts` (3/3 passing) all green locally.

tsconfig.json
```diff
@@ -3,7 +3,9 @@
     "lib": ["ESNext", "dom"],
     "types": ["cypress", "cypress-plugin-steps", "node"],
     "baseUrl": ".",
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "allowImportingTsExtensions": true,
+    "noEmit": true
```
dokterbob, line 8: noEmit is quite a deep change. How is this relevant/necessary in this PR?

dokterbob: I'm a bit afraid there will be nasty effects elsewhere -- I would prefer a solution where we don't touch this kind of common build infra _as part of another PR not targeted at it_.

dokterbob (changes requested):

PR #3035, opened 2026-08-31
fix(frontend): preserve composer drafts across remounts

### LINKED ISSUE #3022: Composer remounts whenever the message list transitions empty↔non-empty, dropping in-progress input

Alphaxiaoteng:
## Summary

- keep the composer draft in one bounded Recoil record so the WelcomeScreen → Footer owner swap cannot discard in-progress input
- make `Input` controlled by that single source of truth while preserving reset, command, favorite, and URL-prompt behavior
- discard the old draft at session boundaries and track URL-prompt initialization so a remount cannot restore a prompt the user edited or deleted
- fixes #3022

## Regression evidence

With the new owner-swap test kept in place and the composer temporarily restored to local state, the test failed as expected:

```text
Expected the element to have value: draft message
Received:
```

The URL-prompt boundary also failed before its guard was added:

```text
Expected: edited draft
Received: initial prompt
```

The first bounded-state revision also proved that merely hiding a stale session record was insufficient: switching `a → b → a` restored `draft message`. The final implementation clears that single record when the session changes.

## Verification

- `pnpm --filter @chainlit/app test -- tests/MessageComposer.spec.tsx` — 4 passed
- `pnpm --filter @chainlit/app test` — 6 files, 36 tests passed
- `pnpm type-check` — frontend and react-client passed; copilot reports its repository-configured skip
- `pnpm lint` and `pnpm lint:fix` — passed
- `pnpm format-check` and `pnpm format` — passed
- `pnpm build` — react-client, frontend, and copilot production builds passed
- `git diff --check` — passed

## Prior work and AI assistance

PR #3028 previously explored the same session-scoped state direction and was closed by its author before maintainer review. Thank you to @Iams4kura for that public investigation. This PR independently re-ran the baseline, adds explicit owner-level, session-lifecycle, and URL-prompt regressions, and follows this repository's required AI attribution. OpenAI Codex assisted with implementation and review; both commits include the required `Co-Authored-By` trailer.

### DIFF
--- a/frontend/src/components/chat/MessageComposer/Input.tsx
+++ b/frontend/src/components/chat/MessageComposer/Input.tsx
@@ -36,7 +36,6 @@ interface Props {
 
 export interface InputMethods {
   reset: () => void;
-  setValueExtern: (value: string) => void;
 }
 
 const Input = forwardRef<InputMethods, Props>(
@@ -97,10 +96,7 @@ const Input = forwardRef<InputMethods, Props>(
       onChange('');
     };
 
-    useImperativeHandle(ref, () => ({
-      reset,
-      setValueExtern: onChange
-    }));
+    useImperativeHandle(ref, () => ({ reset }));
 
     useEffect(() => {
       if (textareaRef.current && autoFocus) {
--- a/frontend/src/components/chat/MessageComposer/Input.tsx
+++ b/frontend/src/components/chat/MessageComposer/Input.tsx
@@ -26,6 +26,7 @@ interface Props {
   className?: string;
   autoFocus?: boolean;
   placeholder?: string;
+  value: string;
   selectedCommand?: ICommand;
   setSelectedCommand: (command: ICommand | undefined) => void;
   onChange: (value: string) => void;
@@ -47,6 +48,7 @@ const Input = forwardRef<InputMethods, Props>(
       autoFocus,
       selectedCommand,
       setSelectedCommand,
+      value,
       onChange,
       onEnter,
       onPaste
@@ -57,7 +59,6 @@ const Input = forwardRef<InputMethods, Props>(
     const [isComposing, setIsComposing] = useState(false);
     const [showCommands, setShowCommands] = useState(false);
     const [commandInput, setCommandInput] = useState('');
-    const [value, setValue] = useState('');
     const textareaRef = useRef<HTMLTextAreaElement>(null);
 
     const normalizedInput = commandInput.toLowerCase().slice(1);
@@ -88,7 +89,6 @@ const Input = forwardRef<InputMethods, Props>(
     });
 
     const reset = () => {
-      setValue('');
       if (!selectedCommand?.persistent) {
         setSelectedCommand(undefined);
       }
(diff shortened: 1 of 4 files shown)


Alphaxiaoteng:
Addressed the review in 89b1968:

- Replaced the `atomFamily` with one bounded active-session draft record, and clear that record when `sessionId` changes, so Recoil cannot accumulate one atom per conversation.
- Moved URL-prompt initialization state into the same record. A user-edited or explicitly deleted prompt now stays edited/deleted across the WelcomeScreen → Footer remount.
- Clarified the session-boundary contract: drafts are intentionally discarded when the session changes rather than retained for later restoration. The regression now exercises `a → b → a` and expects the old draft to remain discarded.

Reverified on the new head: focused composer tests 4/4, full frontend suite 36/36, type-check, lint, Prettier, complete recursive production build, and `git diff --check` all passed.

